PR #1, opened 2021-06-16
installed antd package

Snuzzn:


JWen00 (approved):
Next time you add a package, try to only add package.json as yarn.lock is automatically generated

PR #2, opened 2021-06-17
Sano/ant design

Snuzzn:
Global styles has been set up in _craco.config.js_. You can add more styles into the `modifyVars` object.

Light/dark theme-specific styles have been added in _App.less_. Toggle component has been created to switch between them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 /node_modules
 /.pnp
 .pnp.js
+yarn.lock
 
 # testing
 /coverage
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -4,6 +4,7 @@
 /node_modules
 /.pnp
 .pnp.js
+yarn.lock
```
JWen00, line 7: Yay! Thanks

JWen00 (approved):
Good stuff.

PR #5, opened 2021-06-24
Sano/pc/degree component button

Snuzzn:
Progression dashboard is complete:
- liquid chart and degree component buttons
- when component button is clicked, smooth scrolls to appropriate list view (currently blank except for heading)
- tooltip displaying UOC/total appears when hovering over progress bars and chart


src/components/ProgressionChecker/Dashboard.js
```diff
@@ -0,0 +1,71 @@
...
+        <LiquidProgressChart completedUOC={108} totalUOC={216} />
+        <a
+          href="https://www.handbook.unsw.edu.au/undergraduate/programs/2021/3767?year=2021"
+          target="_blank"
+          rel="noopener noreferrer"
+          className="textLink"
+        >
+          <Title
+            className="text textLink"
+            style={{ marginBottom: "0.5em", marginTop: "0.8em" }}
```
JWen00, line 35: Don't put style here

src/components/ProgressionChecker/LiquidProgressChart.js
```diff
@@ -0,0 +1,93 @@
...
+    radius: 1,
+    width: 320,
+    height: 320,
+    autoFit: false,
+    statistic: {
+      title: {
+        formatter: function formatter() {
+          return "Progress";
+        },
+        style: function style() {
```
JWen00, line 54: This is a good case where ` style: () => {  fill: textColor }` would be much cleaner.

src/components/ProgressionChecker/LiquidProgressChart.js
```diff
@@ -0,0 +1,93 @@
...
+          };
+        },
+        formatter: function formatter() {
+          return (percent * 100).toFixed(0) + "%";
+        },
+      },
+    },
+    liquidStyle: function liquidStyle() {
+      return {
+        fill: percent > 0.45 ? "#9254de" : "#FAAD14",
```
JWen00, line 75: If you're using these colours over and over again, I would suggested having them in a separate file with names - for example GlobalColours

src/components/ProgressionChecker/RadialBarChart.js
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+import { RadialBar } from "@ant-design/charts";
+
+function RadialBarChart() {
+  var data = [
```
JWen00, line 5: Data is good. Keep it in a separate file or make an async call which simulates a real API call and return this data there.

JWen00 (changes requested):
Overall nice work! Some things to change: 
* Keep styling separate
* Dynamically generate your data. e.g. Title and Degreee Component Card
* Cases where `() => {}` would be better

src/components/ProgressionChecker/Dashboard.js
Snuzzn: With my latest update, I was able to remove all the in-line styling in the code except for the skeleton components which for some reason cant be changed via css classname and requires the `style: {{ }}`.

src/components/ProgressionChecker/Dashboard.js
```diff
@@ -0,0 +1,48 @@
...
+        <SkeletonDashboard />
+      ) : (
+        <animated.div className="centered" style={props}>
+          <LiquidProgressChart
+            completedUOC={degree["completed_UOC"]}
+            totalUOC={degree["UOC"]}
+          />
+          <a
+            href={`https://www.handbook.unsw.edu.au/undergraduate/programs/${currYear}/${degree["code"]}?year=${currYear}`}
+            target="_blank"
```
JWen00, line 31: Nice. Let's hope unsw never changes this format 💯

JWen00 (approved):
Hey, not sure if it's a problem on my end. I ended up having to `npm install tslib` before it worked. Before I did that, I saw the error msg: `Attempted import error: '__spreadArray' is not exported from 'tslib'.` 

 Please test that problem by re-cloning the repo and checking into your branch again. Lemme know how you go.

Beyond that, LOOKS VERY GOOD TO ME 👍

PR #11, opened 2021-07-07
Sano/added layout components and dark mode navbar

Snuzzn:
- navbar text colour is now white for dark mode
- navbar sits flush at the top of the page (body padding removed)
- dark mode toggle is no longer a menu item (so hover effect is no longer applied to it)


JWen00 (changes requested):
Looks awesome with the dark theme! I'm a huge sucker for dark theme so this makes me so happy :DD

I popped a msg in discord as well about the Term Planner not being in the navbar. Any reason it's not? :O 

Also wondering if we can temporarily plug the Logo you've made in as well <3  (If we decide to change it later, we can update it)

JWen00 (approved):
I love it! Awesome stuff 🥇

JWen00 (approved):

PR #13, opened 2021-07-08
Jenn/merge progression checker 

JWen00:
* Moved progression checker files out of '/components' and into '/pages'
* Kepts radial chart and liquid chart inside '/components' in their own folder
* renamed 'prgressionChart.js' and 'progressionChart.less' into 'main.js' and 'main.less'
* I am not particularly familiar with usage '.less' files so to get the styles to work I've imported it.. but I noticed you didn't do that before so there might be another way to do it


Snuzzn:
Looks good! I think the new file structure the new folder structure is cleaner. Importing .less files also looks good.

Snuzzn (approved):
Looks good! I think the the new folder structure is cleaner. Importing .less files also looks good.

PR #15, opened 2021-07-08
Gab/cs/degree overview

gabriellahartanto:
this one will have tons of redux change later on


src/pages/CourseSelector/CourseMenu.js
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <Menu
+      onClick={handleClick}
+      style={{ width: '70%' }}
+      defaultSelectedKeys={['1']}
+      defaultOpenKeys={['sub1']}
+      mode="inline"
+    >
+      <SubMenu key="sub1" /* icon={<MailOutlined />} */ title="Recently Viewed">
```
JWen00, line 52: What are the coments for?

src/pages/CourseSelector/CourseList.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
```
JWen00, line 1: Rather than creating a course list, you should create a `course tag` component which can be reused. This way, it's more generic and can be customised to be used in other pages as well.

For better logic, place the `course tag` inside components instead of `courseSelector`

JWen00 (commented):
* Next time don't comit `package-lock.json` and `yarn.lock` - These are auto generated and often create merge conflicts
* I brought up the tab taking up half the page when a course is selected problem with you already
* Try to not commit too much commented code/console.log()'s

PR #18, opened 2021-07-08
Sano/tp/courses drawer

Snuzzn:
- term planner year view is set up
- drag and drop courses between coures (only in offered terms)
- drag and drop courses from sidebar
- relies on fetching data from json (does not yet use redux store)


JWen00 (approved):
It looks pretty good. I am approving this now but I'd like to come back and look further into this code after our showcase.

PR #19, opened 2021-07-08
Gab/cs/add to planner

JWen00:


JWen00:
TODO:
* Search bar - select a person :C
* Add response after clicked 'add to planner'
* Abstract courseList -> courseTag 
* courseTag -> add onClick -> history.push('/../courseCode');
* For dropdown - add "No courses here" if no courses for that category

PR #23, opened 2021-07-12
Setup backend

JamesyJi:
Here are what the different directories do:
**algorithms**
- will hold internal algorithms like isCourseUnlocked or calculateCourseCompletion

**circlesApiWrappers**
- will contain simple function wrappers around api calls which the frontend can then use

**data**
- contains scrapers, processors, raw, formatted ,processed data, etc.

**databaseApi**
- express api to interact with mongodb

**internalApi**
- future flask api to interact with internal endpoints (e.g. javascript calling internal algorithms)

Files:
**init.py**
Will be the future entry point to the server. Running this should boot the server in the future. Currently it just loads and overwrites data to the database

**requirements.txt**
pip requirements

FUTURE GOALS
- symlink so frontend can see backend
- Set up internal api (flask)
- Get some basic getter functions working
- I moved a lot of files around so will have a meeting with backend team to make sure everything is ok


JWen00 (approved):
Routes look good.

PR #29, opened 2021-07-21
[ab/spn-code] adds 'code' field to specialisations

abrieger:
Adds 'code' field to specialisations formatting and processing.

Incidental update to dataHelpers


JamesyJi (approved):
Will proceed with specialisations routes now, thanks!

PR #37, opened 2021-07-26
Sano/tp/store: Refactored code to pull data from redux store

Snuzzn:
CIRCLES-175

- term planner no longer grabs data from json file (instead from redux store)
- large drag functions have been moved to a separate file to keep main.js cleaner
- new redux actions (set_years and set_unplanned) keep redux store updated when courses are moved around in term planner


JWen00 (approved):
Nice work with taking the dragdrop logic out 🥇

PR #40, opened 2021-08-19
Process exclusions in course enrolment rules

abrieger:
Plaintext exclusions appearing into enrolment rules are parsed and put into coursesProcessed.json, and then removed from the enrolment rules in preprocessing. 

**Note**: I have commented out the function _convert_program_codes_ for now. It is not serving any visible benefit at this stage and is super inefficient.


JamesyJi:
Looks good. We'll also have to expand to corequisites too

JamesyJi (approved):

PR #46, opened 2021-08-26
Sano/tp/prereq warnings

Snuzzn:
Ticket Numbers: 107, 108, 110
Completed:
- prereqs parsed from string e.g. COMP1511 & COMP1521
- prereq warning displayed when a course is moved to a term before its prereqs are completed (or if its prereqs aren't planned)
- sidebar shows options to change years and start year but these are currently non-functional
- course can be deleted after moving it

What's next:
- display prereq warning for a course when user moves or deletes its prereq
- implement  updating degree length and start year, so that it is reflected in the term planner (and in redux)


JWen00 (changes requested):
Nice work with handling the per-requisites. I look forward to the warnings when the user moves or deletes the pre-req course. 

I noticed that we currently can't delete the course/move it to unplanned? (I don't know if I'm missing something) 
^Is the plan so that they can only move from planner to unplanned and then only delete a course from unplanned? 

Some quick UI changes to consider
1. For dark theme, the warning colour is too bright and there is not enough contrast with the text
2. The green colour is too bright. Even though it's muted, the constant drag and drop activates and deactivates, making it uncomfortable for the eyes. Try a light grey(?) Up to you if you can find a colour which works!

PR #65, opened 2021-09-10
[CF-38] Create option to select number of years 

Snuzzn:
- settings menu button on top left 
- used tippy.js to create popover menu with custom html
- responsive with different sizes and with summer view
- **"change start year" and "degree length" now  functional**


JWen00 (approved):
Awesome stuff!

JWen00 (approved):

PR #70, opened 2021-09-12
FINS manual-fixes for 2022

JamesyJi:


backend/data/processors/manualFixes/FINSFixes.py
```diff
@@ -0,0 +1,156 @@
...
+
+def FINS_2622():
+    """
+    "original": "Prerequisite or co-requisite:FINS1612 and FINS2624<br/><br/>",
+
+    "processed": "|| [FINS1612 && FINS2624]"
+    """
+
+    # BOTH prerequisite and corequisite
+    return "(FINS1612 && FINS2624) || [FINS1612 && FINS2624]"
```
abrieger, line 71: If co-req, should this just be `[FINS1612 && FINS2624]`; i.e. do we still need the `(FINS1612 && FINS2624)`? I think we should stick with just having `[FINS1612 && FINS2624]` because the regex successfully does this in some instances. Thoughts?

JamesyJi: I thought the same initially but turns out, co-rerequisite means have to take in the same term (and only in the same term). So if we just had [FINS1612 && FINS2624], then they have to take FINS1612 and FINS2624 in the same term. So if it's just corerequisite, then even if they took FINS1612 in a prior term, it wouldn't count

abrieger: Are you sure about that? The glossary says "A course that must be enrolled in concurrently with another course (**unless it has already been completed**), usually due to a relationship between the courses in the program curriculum." This means that a student can complete the course prior to the target course. I think the [ ] should include all previous terms and the term being enrolled in.

JamesyJi: gg you're right

JamesyJi: I've made the changes, squash & merge if it looks good

abrieger (approved):

PR #76, opened 2021-09-16
Tabl manual fixes for 2022

JamesyJi:
Made some specific improvements to regex as well


abrieger:
Changes look good, but got a conflict that can't be resolved via web. We should probably remove conditionsProcessed and coursesProcessed from the commit - might need to do a soft reset to take those off. Let me know if you're happy for me to jump into the branch and do that.

JamesyJi:
@abrieger Yep all good. Not a big deal since we'll just run the processors to recreate those files anyways

abrieger:
> @abrieger Yep all good. Not a big deal since we'll just run the processors to recreate those files anyways

Ok cool. I've fixed the conflict and will merge now

PR #80, opened 2021-09-17
Updated programs to 2022 handbook

Jevelry:


JamesyJi (approved):
I think this looks good. Will see if any problems arise when updating the API

PR #81, opened 2021-09-18
Programs contain specialisation and course names

Jevelry:
Removed some redundant files and added names next to courses in specialisations


JamesyJi (approved):
Programs are missing stuff like "any Climate Science course" but I think we should combine the code later with specialisations so we're not doubling up on code

PR #83, opened 2021-09-23
[ab/update_run_processors] Shell driver for manual fix files

abrieger:
Adds driver for running manual fixes in `shell`. I originally used a python script, but it wasn't going to be scalable since you'd have to manually insert the imports for each manual fix file. The shell script will find all `*Fixes.py` files and execute them.

Also note change to dataHelpers so that it does not exit on choosing not to proceed with an overwrite.

Usage: `python3 runprocessors.py --type=condition --stage=manual`

Considering adding a force write option in the future so manual fix files do not keep asking whether you want to overwrite.


JamesyJi (approved):
gg the shell

PR #84, opened 2021-09-23
[ab/syntax_processing] Checks for mismatched brackets and words unknown to algorithm

abrieger:


JamesyJi (approved):
Looks fine but I haven't played around with it yet so WILL JUST TRUST THAT IT WORKS.

PR #86, opened 2021-09-24
Sano/tp/complete term

Snuzzn:
- marking term as complete via checkbox prevents courses from being moved in and out
- courses colours (incl. warning) are dulled when the term is marked as complete


JWen00 (changes requested):
Looks awesome! 

Would be nice:
* Disable the right click to open menu if the term has been checked
* Update the cursor for the course to `not-allowed` cursor when the term has been checked
```
.not-allowed {cursor: not-allowed;}
```

* Using the grab and grabbing cursor on drag (idk how difficult this is with your current implementation. Definately a bonus feature so don't worry if it gets painful to do) 
```
.grab {cursor: -webkit-grab; cursor: grab;}
.grabbing {cursor: -webkit-grabbing; cursor: grabbing;} 
```

PR #88, opened 2021-09-24
Comm manual fixes

VincentXiao-jpg:
Should be the comm manual fixes done


JamesyJi (commented):
Can you delete the .gitignore and recommit? But make sure you're in the backend/ directory, since if you commit from the root directory, it'll pick up things from the frontend and idk if .DS_Store is something that matters to them

JamesyJi (approved):

PR #89, opened 2021-09-26
Econ manual fixes

gabriellahartanto:
- 2022 version
- not sure about Assumed Knowledge in ECON4201
- not sure about the 'except' part of ECON1101 and ECON1203


backend/data/processors/manualFixes/ECONFixes.py
```diff
@@ -0,0 +1,192 @@
...
+# TODO: implement your functions here
+# NOT SURE HOW THIS WORKS (FOR THE EXCEPT PART)
+def ECON_1101(code):
+    """
+    "original": "Excluded: Students in BCom single or double-degree programs, except for 3155 Actl/Comm and 3521 Comm/Econ, are not allowed to enrol in this course. <br/><br/>"
+    
+    "processed": ""
+    """
+
+    COURSES[code]["exclusions"]["COMM#"] = 1
```
JamesyJi, line 76: This condition is so doomed, I think one possibility is to not add the "COMM#" to the exclusion and instead just return that as a warning: "Students in BCom single or double-degree programs, except for 3155 and 3521 cannot enrol"

This way we don't restrict eligible students but still warn ineligible students

backend/data/processors/manualFixes/ECONFixes.py
```diff
@@ -0,0 +1,192 @@
...
+    "processed": "Must not be Program 3715 || 3764"
+    """
+    
+    COURSES[code]["exclusions"]["MATH2841"] = 1
+    COURSES[code]["exclusions"]["MATH2801"] = 1
+    COURSES[code]["exclusions"]["MATH2901"] = 1
+    COURSES[code]["exclusions"]["MATH2099"] = 1
+    COURSES[code]["exclusions"]["ACTL2002"] = 1
+    COURSES[code]["exclusions"]["ACTL2131"] = 1
+    COURSES[code]["exclusions"]["COMM#"] = 1
```
JamesyJi, line 94: Likewise with this one exclusion

backend/data/processors/manualFixes/ECONFixes.py
```diff
@@ -0,0 +1,192 @@
...
+
+    return "ECON1203 || COMM1190 || ECON2403 || MATH1041 || MATH1231 || MATH1241 || MATH1251"
+
+def ECON_2403():
+    """
+    "original": "Prerequisites: PPEC1001 AND ECON1202 AND in program 3478 or 4797<br/><br/>"
+    "processed": "PPEC1001 && ECON1202 && in program 3478 || 4797"
+    """
+
+    return "PPEC1001 && ECON1202 && 3478 || 4797"
```
JamesyJi, line 115: Remember to surround in brackets, (3478 || 4797)! Or else the logic is ambiguous

JamesyJi (changes requested):

PR #94, opened 2021-09-28
[CF-51] Allow users to search in the search tab with debouncing

gabriellahartanto:


JWen00 (changes requested):
Heya! I moved your `SearchCourse` to the search tab.

Some things to do (which you may already have been planning to do)

* Could you remove the unnecessary commented out code? 🐤 
* Also would be helpful to provide an insight on what this is doing in your description 🙇‍♀️ 
* Add some styling 
* OnClick, direct users to another tab with their searched course and close the current search tab

Can't wait!

PR #95, opened 2021-09-28
[CF-59] Map categories of specialisation with dummy progression data

sallyysun:
mainly progression checker stuff, don't mind the degree selector


JWen00:
@sallyysun  Have a look at the ticket information. :)

PR #107, opened 2021-10-04
[CF-46] Allow users to hide a particular year in the term planner

Snuzzn:
- clicking on year will bring up an eye button to hide that year
- once 1 or more years are hidden, an unhide button will pop up next to settings button. Click on this to unhide all years.


JWen00 (approved):
Approved!

Changes made: 
* Added tooltip for 'Show all hidden years'
* Added tooltip for 'Hide year'
* Added notif for when user tries to hide every year

PR #111, opened 2021-10-05
Optimises API calls for getStructure and fixes API bug in courses caused by null values

abrieger:
Two changes:
1. Optimises getStructure API calls by removing calls to the courses database table. Originally the call was made to get the course title, but the updated specialisations and programs data now includes the course title (e.g. "COMP1511": "Programming Fundamentals", compared to before when it was "COMP1511": 1) so we can just use that.
2. null values in courses for "UOC", "description" and "levels" killed API calls, so James and I updated coursesProcessing and coursesFormatting to remove null values and replace with more suitable values.


JamesyJi (approved):
What were the odds we found this bug lol

PR #118, opened 2021-10-15
Sally/degreeselector

sallyysun:
created basic table view for progression checker (dark mode a bit screwed, fixing)


Peedee2002:
worked on this PR to make it to a *usable* state. I integrated it with the backend "getStructure" API and fixed a CSS issue, but there seems to be a whole lot more work to be done here on both a usability and a style standpoint. Here is how i made it look now:
image
image
image
@Snuzzn you can review this now to see if i borked anything.

Snuzzn (approved):
I think the progression checker is a feature that needs to be reworked (plan on doing this with onboarding). For now, this doesn't break anything so thats good. Also, I made a couple of updates to the api calls that were using the old endpoints.

PR #144, opened 2021-11-21
Recommendations sac

IsaacYaner:
update get_unlocked_courses(user)

### DIFF
--- a/backend/algorithms/recommendations.py
+++ b/backend/algorithms/recommendations.py
@@ -11,8 +11,29 @@ NOTE: ENSURE YOU HAVE BRANCHED OFF!!!
 
 NOTE: For sample user data, look at exampleUsers.json to get an idea of the structure
 """
-from .conditions import User, create_condition, CACHED_CONDITIONS_TOKENS
+from algorithms.conditions import User, create_condition, CACHED_CONDITIONS_TOKENS
 import json
+
+cached_conditions = None
+cached_course_list = None
+
+def load_condition_courselist():
+    global cached_conditions
+    global cached_course_list
+    if cached_conditions is not None:
+        return cached_conditions, cached_course_list
+    COURSE_PATH = "./data/finalData/conditionsTokens.json"
+    course_list = {}
+    with open(COURSE_PATH, encoding='UTF-8') as f:
+        cached_course_list = json.load(f)
+        f.close()
+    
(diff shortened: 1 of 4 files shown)


backend/algorithms/recommendations.py
```diff
@@ -11,14 +11,38 @@
...
+    COURSE_PATH = "./data/finalData/conditionsTokens.json"
+    PICKLE_FILE = "./algorithms/conditions.pkl"
+    course_list = {}
+    result = []
+    with open(COURSE_PATH, encoding='UTF-8') as f:
+        course_list = json.load(f)
+        f.close()
+    
+    # Load all the necessary conditions
+    cached_conditions = {} # Mapping course to condition object
```
JamesyJi, line 33: Best to have cached_conditions initialised outside the function so it's only created once since the Condition objects themselves don't change

backend/algorithms/recommendations.py
```diff
@@ -11,14 +11,38 @@
...
+    for course in course_list:
+        cached_conditions[course] = create_condition(CACHED_CONDITIONS_TOKENS[course], course)
+
+    for course in course_list:
+        if course in cached_conditions and cached_conditions[course] != None:
+            # The course is not locked and there exists a condition
+            cond = cached_conditions[course]
+            if (cond.is_unlocked(user))["result"] is False:
```
JamesyJi, line 41: Did you mean is True?

backend/algorithms/recommendations.py
```diff
@@ -11,14 +11,38 @@
 
 NOTE: For sample user data, look at exampleUsers.json to get an idea of the structure
 """
-
+from .conditions import User, create_condition, CACHED_CONDITIONS_TOKENS
```
JamesyJi, line 14: can you import from the backend directory? So change .conditions to algorithms.conditions.

You can run from the backend directory with python3 -m algorithms.recommendations

If everything is run from the backend directory, it makes importing to other areas like the API much easier

JamesyJi (changes requested):
Some minor changes requested, otherwise looks good!

backend/algorithms/recommendations.py
IsaacYaner: I added a function for loading the cached conditions and changing the paths. They are running properly by calling python -m algorithms.recommendations now.

PR #146, opened 2021-11-22
Select legacy courses api

VincentXiao-jpg:
Api done for select legacy


JamesyJi (approved):
We might need to return some extra stuff like UoC later on since some legacy courses might not exist in current handbook anymore

PR #153, opened 2021-12-15
Recommendations sac

IsaacYaner:
Completed original design of get_courses_path_to.
There are several questions about this method, and they are commented above the function.


JamesyJi (changes requested):
@IsaacYaner I think your suggestion is very good (not counting the courses the user has already unlocked). Could you implement that? I think you would need to go through all the courses and call is_unlocked() on them to check. Maybe can have a helper function like "get all unlocked courses".

PR #173, opened 2022-01-23
CIRCLES-250: fix made for backend, made frontend use the right endpoint

Peedee2002:
fixes the bug and makes the right call from the frontend side. Also deleted redundant API files.


JamesyJi (approved):

Snuzzn:
Awesome, route works now! :) But some fields which were in the old version of this endpoint are no longer present (see below)
image

Snuzzn (approved):

PR #174, opened 2022-01-24
Peter/circles 249 add type field to get all unlocked

Peedee2002:
added the type field as a list of strings. This can be because a course may match multiple descriptions.


Snuzzn (approved):

Snuzzn (approved):
Looks good. I think there a couple of courses that you may have to manually categorise in the db. For example, MATH1131 (& MATH1231) and MATH1231 (& MATH1241) are core courses in computer science, but they aren't given a type in the response for this endpoint. Similarly, COMP3121 should be a core course not a computing elective. 
Anyway, besides that minor issue, I should be able to use this endpoint to categorise the courses now. Thanks!

Peedee2002:
@Snuzzn oh my bad... those are bugs. I'll fix those asap.

PR #175, opened 2022-01-28
Peter/circles 258 clean up code i dont like

Peedee2002:
i was annoyed by some of the duplicated/overly complex code i found so i refactored it. I tested this in frontend and i dont think i broke anything, but we will see what happens in CI. This causes a change in the root stuff for the API again, so i fixed it on the frontend. I am sorry it keeps changing, hopefully that will stop now.


JamesyJi (commented):
Looking at the categories code, it is much cleaner. However I'm not sure I wrote extensive tests for categories mainly because they haven't been used much yet. So will keep an eye on this + push more tests later if I have time.

Will leave the final approval up to @Snuzzn

Snuzzn (approved):

Snuzzn (approved):

PR #180, opened 2022-02-05
CIRCLES-256: added the API tests to CI and added more robust tests to getAllUnlocked

Peedee2002:
Notice that these are integration tests - i added docker to the CI to run the backend and the database in docker in the background of these tests. This makes it *very* easy to write these tests, but they will depend on the database - if it changes, these tests may also need to change. Usually the good thing to do here is to avoid making assumptions/use black box tests, but that is often hard.


.github/workflows/ci.yml
```diff
@@ -25,8 +25,20 @@ jobs:
...
+    - name: set up docker compose
+      run: |
+        mkdir env
+        echo "MONGODB_USERNAME=github-ci" >> env/backend.env
+        echo "MONGODB_PASSWORD=github-ci" >> env/backend.env
+        echo "MONGODB_DATABASE=mongodb" >> env/backend.env
+        echo "MONGO_INITDB_ROOT_USERNAME=github-ci" >> env/mongodb.env
+        echo "MONGO_INITDB_ROOT_PASSWORD=github-ci" >> env/mongodb.env
+        docker compose build backend
+        docker compose up backend -d
```
abrieger: Nice update to the workflow! GitHub Actions is pog. You could combine lines 36 and 37 with a --build on 37 to force build on the compose but makes no difference otherwise 🙂

Peedee2002:
on a whim, i also changed wam and UOC to be queries instead of persistant, and made them be able to accept a category.

.github/workflows/ci.yml
Peedee2002: ooh i didnt know about this param - changed 😃

JamesyJi (approved):
Looks good.

My slight concern is that when changing user.wam to user.wam(), the wam has to be recalculated each time. This maxes out at ~32 iterations which is fine for us since we don't currently have any batch updates (i.e. calculating the WAMs of 1000 users to determine an average).

^ IMO python was not the best choice but it is what it is

PR #182, opened 2022-02-11
Sano/dw/integration and redesign

Snuzzn:
Sorry, this PR is quite big. Originally, this was meant to be just about integration and I thought I might add a few small design changes but that quickly got out hand. 
Here are the changes:
- Major/minor steps are hooked up to API
- Degree wizard settings are stored in localStorage and retrieved each visit
- Loading page when website is visited, to determine whether the user has existing data stored in the browser (go to course selector) or if it's their first time (go to degree wizard)

image

-  Design change: originally one question per page (with scrolling) -> now questions sit flush with each other and animate in when the previous step is completed (inspired by [youth allowance form](https://www.servicesaustralia.gov.au/how-to-claim-youth-allowance-for-students-and-apprentices?context=43916#dc-53194-s-208431-208539-208545-208550) ). 
- Navbar removed from degree wizard (so user can't move around to other pages if they haven't filled out the questions)
image
- I also removed dark theme for deployment as there are too many styles that aren't working for it.


frontend/src/pages/DegreeWizard/steps/DegreeStep.jsx
```diff
@@ -1,26 +1,25 @@
...
   const fetchAllDegrees = async () => {
-    // const res = await axios.get("http://localhost:8000/api/getPrograms");
+    // const res = await axios.get("http://localhost:8000/programs/getPrograms");
     // setOptions(res.data["programs"]);
     setOptions({
-      3778: "Bachelor Computer Science",
-      3779: "Bachelor of Science",
-      3777: "Bachelor of Arts",
-      3775: "Bachelor of Commerce/Law",
+      3778: "Computer Science",
```
Peedee2002: Hmm i think I should do this on the backend - this restriction is something that the backend needs to deal with, not the frontend. Ill change the endpoint to just return CS for now.

Peedee2002 (approved):
well done! lgtm

frontend/src/pages/DegreeWizard/steps/DegreeStep.jsx
Snuzzn: right, yea that would be better. I've updated it so it uses the endpoint. 
So for now, many results pop up but when you end up changing the endpoint, it'll just be comp sci (and whatever else is ready).

PR #184, opened 2022-02-15
CF-108: Fixed bug where general requirements (from getStructure) was breaking the sidebar..

Snuzzn:
Since categories under 'general' (in `programs/getStructure`) don't usually have a courses key, it was breaking the sidebar.  For now, I've now ignored objects that don't have the `courses` object, so unlocked courses won't ever go into these categories e.g. General Education and Flexible Education. 

image

This may also be ideal in a design sense, because if they did, those categories would become very full right from the start since pretty much any course can be considered to be unlocked as General Education or Flexible Education.  

I think we'll have to discuss how to approach this later. I'm thinking that a user has to manually search for a course and add it to planner, for it to be appear under those 'general' categories. That way, those categories don't fill up with tons of unlocked courses. We'd probably need a separate endpoint to determine if a course can be considered as general education or flexible education. Anyway, this issue will need further discussion but for now, this PR will resolve the breaking sidebar.


Peedee2002 (approved):
Lgtm

PR #186, opened 2022-02-22
CIRCLES-275: refactor conditions.py to simplify it

Peedee2002:
completely refactored now, got rid of firstCompositeCondition and got rid of camelcase.
The reason why this PR seems massive is just because of the renaming.


backend/algorithms/objects/conditions.py
```diff
@@ -211,78 +215,17 @@ def set_logic(self, logic):
...
-                    continue
-
-        unlocked = self.validate(user, warnings)
+            return True, []
 
-        return {
-            "result": unlocked,
-            "warnings": warnings
-        }
+        validations = [cond.validate(user) for cond in self.conditions]
```
JamesyJi, line 222: It's late at night but I'm just gonna trust you handled the warnings in a better way. The original purpose of FirstCompositeConditions was also to be the top level where warnings are collected. However seems like the team is coming up with something better!

backend/runprocessors.py
```diff
@@ -10,13 +10,13 @@
 from algorithms.load_conditions import cache_conditions_pkl_file
 from algorithms.log_broken import log_broken_conditions 
 
-from data.scrapers.programsScraper import scrape_programs as scrape_prg_data
-from data.scrapers.specialisationsScraper import scrape_spn_data
-from data.scrapers.coursesScraper import scrape_courses as scrape_course_data
```
JamesyJi, line 15: Maybe can standardise the function names as well inside all these files instead of us having to rename the function in the import

JamesyJi (approved):
Looked through the files. Left 2 comments. 1 is commentary on original purpose of FirstCompositeCondition just to ensure your new solution will account for it. The other is a suggestion on standardising function names inside the scrapers and formatters. 

It looks cleaner!

backend/algorithms/objects/conditions.py
```diff
@@ -1,75 +1,69 @@
...
 CACHED_PRGORAM_MAPPINGS_FILE = "./algorithms/cache/programMappings.json"
 with open(CACHED_PRGORAM_MAPPINGS_FILE) as f:
     CACHED_PRGORAM_MAPPINGS = json.load(f)
-    f.close()
 
-class CourseCondition():
+class Condition():
+    @abc.abstractmethod
+    def validate(self, user) -> tuple[bool, list[str]]:
+        pass
```
hdphuong, line 32: maybe add a line of comment here so that other people know what the returned values mean. The tuple is not self-explanatory

hdphuong (approved):
looks neat

backend/algorithms/objects/conditions.py
Peedee2002: Basically what I did is that I forced every validate function to return a "warnings" list. This list is then flattened for composite conditions. I think this is better because it lets us deal with all "Condition" objects in the same way, allowing us to verify the return type and the fact that it has a "validate" method statically.

In a sense I basically used hindsight to unify all the conditions objects under one interface.

Snuzzn (approved):

PR #188, opened 2022-02-26
CF-107: Unschedule all courses in term planner button

leonardo-fan:
- Added Unschedule all button into optionsHeader
- Added UNSCHEDULE_ALL planner action
- Added UNSCHEDULE_ALL planner reducer
- Turned previous UNSCHEDULE reducer for one course into a function that is now accessed by both the UNSCHEDULE case and UNSCHEDULE_ALL case


frontend/src/reducers/plannerReducer.js
```diff
@@ -38,6 +38,35 @@ if (planner) initialState = extractFromLocalStorage(planner);
...
+  const unscheduleCourse = (code) => {
+    let updatedUnplanned = state.unplanned;
+    updatedUnplanned.push(code);
+    const termTag = state.courses.get(code).plannedFor;
+
+    const yearI = parseInt(termTag.slice(0, 4)) - state.startYear;
+    const termI = termTag.slice(4);
+
+    const nTerm = new Object(
+      state.years[yearI][termI].filter((course) => course !== code)
```
Peedee2002: I'm not sure what's with the use of the `new` keyword so often. I see stuff like `new Map` too. I know you didn't write this code, but couldn't this be written more expressively with using {} notation? I know this is a nitpick, so ill approve you anyways if you find this easier to read.

Peedee2002 (approved):

frontend/src/reducers/plannerReducer.js
leonardo-fan: Yeah sure I see, this would just involve removing the ```new Object``` and ```new Map``` keyword right? And if something is being assigned to only ```new Map()``` this would instead be replaced by ```{}```?

Peedee2002: yea that is the hope. I dont think its quite that simple (I tried doing this before and gave up because it was taking too long) but that is the idea yes.

leonardo-fan: ahh righto I see, will have a look

Snuzzn (changes requested):
This looks great! I like how you put the 'unschedule' stuff into its own function and reused it. 
I did notice an edge case: when the user has no courses in the term planner and they click on the button, it breaks.
image
I think there are 2 options to go about this scenario: not show the button if the planner is empty OR show a [notification](https://ant.design/components/notification/) if they click on the button saying that they don't have any courses planned. 
The former option is probably better but it might be harder to implement. Aside from that, great work! 

Also, while you're doing that, can you also add a hover cursor to all the top left buttons and perhaps a box-shadow (https://getcssscan.com/css-box-shadow-examples). I remember in the last meeting, someone pointed out that they didn't look like buttons. This should be a pretty small change so I think it doesn't need its own ticket. Thanks!

Snuzzn (approved):
Awesome! lgtm :)

PR #189, opened 2022-02-26
josh/circles 257 test unselectCourse

joooooooooooooooooooooooooooooooooooosh:
This adds some basic tests for the unselectCourse API. This is by no means an exhaustive suite of tests.


backend/server/tests/test_unselectCourse.py
```diff
@@ -0,0 +1,33 @@
+import json
+import requests
+
+PATH = "algorithms/exampleUnselectCourse.json"
+
+with open(PATH) as f:
+    USERS = json.load(f)
```
Peedee2002: technically this is a single user, not USERS

backend/server/tests/test_unselectCourse.py
```diff
@@ -0,0 +1,33 @@
...
+
```
Peedee2002, line 13: most of this functionality is already tested by the function this API test is calling, which is tested here:
https://github.com/csesoc/Circles/blob/dev/backend/algorithms/tests/test_user_unselect.py
I think that some of the testing you do here is more rigourous, so we can move some of it there. Here, i would have preferred something that tests the other stuff this function does:
eg 
- it fixes UserData to make sure it populates the UOC if it is not supplied by the frontend
- the shape is slightly changed. check that this happens. 

also while you are editing this function, you can add some fastAPI niceties such as adding a response model and an example response.

joooooooooooooooooooooooooooooooooooosh: fastAPI docs and UOC checks have been added.

Peedee2002 (approved):
Nice changes! all that is left is fixing the merge conflict and we are done :)

PR #190, opened 2022-02-26
CF-110: display 'none' text for requirements with no prereqs

gtangelo:
- Change Requirements text to display 'None' if there are no prereqs returned from a course

### DIFF
--- a/frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
+++ b/frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
@@ -150,7 +150,9 @@ export default function CourseDescription() {
               </Title>
               <Space direction="vertical" style={{ marginBottom: "1rem" }}>
                 <Text>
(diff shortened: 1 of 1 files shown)


frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -150,11 +150,7 @@ export default function CourseDescription() {
               </Title>
               <Space direction="vertical" style={{ marginBottom: "1rem" }}>
                 <Text>
-                  <div
-                    dangerouslySetInnerHTML={{
-                      __html: course.raw_requirements,
```
Peedee2002, line 155: the reason we dangerously set html here is because the handbook API sends us raw HTML for this and we want to display the same way. We probably want to keep it how it was before.

Peedee2002 (approved):
Noice

Snuzzn (approved):
lgtm!

PR #191, opened 2022-02-26
CIRCLES-267: rework path_to to return what is unlocked when taking the course

banaboi:


backend/data/processors/courses_processing.py
```diff
@@ -123,7 +124,8 @@ def process_enrolment_path(processed: dict, formatted: dict, data: dict) -> None
             if "path_to" not in data[prereq]:
                 # Set up dict if not yet added
                 data[prereq]["path_to"] = dict()
-            data[prereq]["path_to"][processed["code"]] = 1
+            # data[prereq]["path_to"][processed["code"]] = 1
+            data[prereq]["path_to"] = courses.coursesUnlockedWhenTaken(prereq)
```
banaboi: This is missing userData parsed in, but where would this come from?

Peedee2002: yea - the way you have it now, it cant come from anywhere - so we should completely seperate this functionality from getCourse as you have mostly already done.

backend/server/routers/courses.py
```diff
@@ -178,3 +178,35 @@ def unselectCourse(userData: UserData, lockedCourses: list, unselectedCourse: st
...
+    """ Returns all courses which are unlocked when given course is taken """
+    
+    ## define the user object with user data
+    user = User(fixUserData(userData.dict()))
+    
+    ## initial state
+    courses_initially_unlocked = getAllUnlocked(userData)
+    
+    ## add course to the user
+    courseToAdd = {courseToBeTaken: (getCourse(courseToBeTaken), None)}
```
Peedee2002: this expects something in this form:

"COMP1511" : [6, 70] where 6 is the UOC, and 70 is the course mark.
I dont think this will behave as you expect.

backend/server/routers/courses.py
```diff
@@ -178,3 +178,35 @@ def unselectCourse(userData: UserData, lockedCourses: list, unselectedCourse: st
...
+    user.add_courses(courseToAdd)
+    
+    ## final state
+    courses_now_unlocked = getAllUnlocked(user.get_user_data())
+    
+    ## compare 
+    before = []
+    after = []
+    course_keys = list(courses_initially_unlocked.keys())
+    for course in course_keys:
```
Peedee2002: this and the next should be list comprensions.
or better yet, make them explicitly sets https://www.w3schools.com/python/python_sets.asp. Using this will probably make this code incredibly elegant.

backend/server/routers/courses.py
banaboi: Are you sure? Looks like the key is a string and the value is a tuple of ints? Am i missing something? 
image

Peedee2002: I think any iterable is acceptable here - the key must be the course name and the iterable must have the first thing be the UOC and the second be the grade

banaboi: Ah okay, will have a look

Peedee2002: Basically I think the only thing you need to do is:
GetCourse(course) -> getCourse(course)['UOC']

Peedee2002:
Nice, looks great! Are there any tests that come with this? Just a few API tests for this would be awesome.

Peedee2002:
Also, since you changed the front-end (kinda) a screenshot of your changes would be cool too.
Sorry if I'm being a bit harsh here lol - I'm just traumatised by the amount of test cards I had to do and write

Snuzzn:
Ok, so right now this isn't working in the frontend because you did a get request when the endpoint is a post. 
As for the payload (user), it requires a bit of preparation/reorganisation of frontend data which I initially did in the CourseMenu.jsx (when integrating`getAllUnlocked`). I've now refactored it so that its in a helper function (in its own file). This means you just need to call that helper function to prepare the payload when making your api call. You can refer to `CourseMenu.jsx` to see how it's used. 
Sorry about that. I should have had it in a helper function from the start, but hopefully it's now clear what you should do. :)

Snuzzn (changes requested):
Ok, so right now this isn't working in the frontend because you did a get request when the endpoint is a post. 
As for the payload (user), it requires a bit of preparation/reorganisation of frontend data which I initially did in the CourseMenu.jsx (when integrating the `getAllUnlocked` endpoint). I've now refactored this preparation stage so that its in a helper function (in `helper.js`). This means you just need to call that helper function to prepare the payload when making your api call. You can refer to `CourseMenu.jsx` to see how it's used. 
Sorry about that. I should have had it in a helper function from the start, but hopefully it's now clear what you should do. 🙂

Snuzzn (commented):
Sorry for the late review. I wasn't notified of the request for some reason. Anyway, the endpoint doesn't seem to be working. I tried testing it on Postman and it gives me error 500 (internal server error).
image
I used the following input as the request body:
```json
{
  "program": "3778",
  "specialisations": {
    "COMPA1": 1,
    "MARKA2": 1
  },
  "courses": {
    "COMP1521": 70,
    "COMP1511": 70
  },
  "year": 2
}
```

banaboi:
image

image


Seems to be working okay

backend/server/routers/courses.py
```diff
@@ -122,7 +122,7 @@ def getAllUnlocked(userData: UserData):
     that they have already completed"""
 
     coursesState = {}
-    user = User(fixUserData(userData.dict()))
+    user = User(fixUserData(userData.dict())) if type(userData) != User else userData
```
banaboi, line 125: Added this so i could just pass in the user object

frontend/src/pages/CourseSelector/helper.js
```diff
@@ -0,0 +1,20 @@
...
+  const { startYear, courses } = planner;
+  const { programCode, specialisation, minor } = degree;
+
+  const specialisations = {};
+  specialisations[specialisation] = 1;
+  if (minor !== "") specialisations[minor] = 1;
+
+  let selectedCourses = {};
+  for (const course of courses.keys()) {
+    selectedCourses[course] = 70;
```
Peedee2002, line 11: is this some dummy value? I dont see the segnificance of the number `70` here.

backend/server/routers/courses.py
Peedee2002: oh btw i think you forgot to add the test file

frontend/src/pages/CourseSelector/helper.js
banaboi: I think this was sano's changes he made to my branch to make things easier on the frontend side

backend/server/routers/courses.py
banaboi: Yeah doing that now

banaboi: Added

backend/server/routers/courses.py
```diff
@@ -178,3 +178,25 @@ def unselectCourse(userData: UserData, lockedCourses: list, unselectedCourse: st
     affectedCourses = User(fixUserData(userData.dict())).unselect_course(unselectedCourse, lockedCourses)
 
     return {'affected_courses': affectedCourses}
+
+@router.post("/coursesUnlockedWhenTaken/{courseToBeTaken}")
```
Peedee2002: what happened to the fastAPI stuff? stuff like the responseModel (a list of strings), and example output are very useful for the frontend team to understand how to use this API.

Peedee2002 (approved):
nice work! one small change and we can merge in :)

backend/server/routers/courses.py
banaboi: Yeah with that the route was returning just {"courses_state": {}} but I'll try and add it back in

banaboi: Added

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -30,14 +31,34 @@ export default function CourseDescription() {
...
+    const res = await axios.post(
+      `http://localhost:8000/courses/coursesUnlockedWhenTaken/${id}`,
+      JSON.stringify(prepareUserPayload(degree, planner)),
+        {
+          headers: {
+            "Content-Type": "application/json",
+          },
+        }
+    );
+    dispatch(setCoursesPathTo(res.data.courses_unlocked_when_taken));
```
Snuzzn: I don't think this needs to be a `dispatch()` since `coursesPathTo` is local state. 
That's why there's an error in the console:
image

So I think if you should just remove dispatch (so it's just `setCoursesPathTo(res.data.courses_unlocked_when_taken)`), it should fix that. Besides that, everything looks good to me! :)

Snuzzn (approved):

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
banaboi: Fixed

PR #195, opened 2022-02-27
CF111: Option to manually add a course as gen-ed or free elective

gtangelo:
Added dropdown for course selector's 'Add to planner' button that allows the user to add a course either as a gen ed or free elective. Once added, the course is then displayed it has been added on the sidebar under 'General Education' or 'Flexible Education' respectively.

image


Peedee2002 (approved):
lgtm!
The logic you added is slightly complex inside that for loop - it would be nice if that is simplified, but im not too fussed there.

Snuzzn (approved):
Lgtm!

PR #198, opened 2022-03-11
CIRCLES-270 refactor caching school and faculty data

Rory-Mac:
Remove obsolete mappings folder and update cache.py to avoid hard-coding school/faculty data


backend/algorithms/cache/cache.py
```diff
@@ -72,27 +72,73 @@ def cache_handbook_note():
...
+        else: 
+            match_object = re.search("^([\w]+)", School)
+        match = match_object.group()
+        school_token += match
+        return school_token
+
+    # add faculties to mappings.json
+    for course in courses:
+        faculty = courses[course]['faculty']
+        if faculty in mappings:
```
Peedee2002: this can probably be simplified with:
```
if faculty not in mappings:
      do_stuff()
```

Peedee2002 (changes requested):
well done! just a few things:
- make sure youve run the runprocessors.py fille - this will actually run this code to re-write the mappings and coursemappings. 
- some places in ManualFixes have hardcoded the expected names of faculties and schools, have you looked over them to check they arent a problem? you can search the codebase for the strings 'F ' and 'S ' to locate them.

backend/algorithms/cache/cache.py
```diff
@@ -72,27 +72,71 @@ def cache_handbook_note():
...
+        faculty = courses[course]['faculty']
+        if faculty not in mappings:
+            faculty_token = tokeniseFaculty(faculty)
+            mappings[faculty] = faculty_token
+            courseMappings[faculty_token] = {} 
+    # add schools to mappings.json
+    for course in courses.values():
+        if 'school' in course:
+            school = course['school']
+            if school not in mappings:
```
Peedee2002, line 114: this is just my own preference, but i think it would be good if you got rid of a level of indentation by doing
```
if 'school' in course and course['school'] not in mappings:
```
... rest of your code.
its a small thing, so ill approve you anyways. If you agree, feel free to fix that then merge it in!

Peedee2002 (approved):
nice. You love to see that amazing number of removed loc.

Peedee2002 (commented):
sorry, i just noticed what you were talking about with your issues. So, there seems to be a few small issues:

1. with your dependancy issue, i think it might be because youre attempting to run from the incorrect directory. We can hop on a VC and chat about the fix to that. 
2. also, you can see here:
https://github.com/csesoc/Circles/runs/5597966523?check_suite_focus=true
that your pipeline is failing. This is because you seem to have broken a test. I suspect that this can be because the `CACHED_MAPPINGS` no longer exist, since it was extraneous information. You will need to investigate how to get that code running again (probably all you need to do is change the name of the file that is getting pointed at)

Peedee2002 (changes requested):
so ill leave the `request changes` label for now until we get the pipeline passing.

Peedee2002 (approved):
lgtm!

PR #200, opened 2022-03-17
Mae/circles 255 test get course search course get legacy courses

hdphuong:


backend/server/tests/test_getCourse.py
```diff
@@ -0,0 +1,28 @@
...
+def test_error():
+    x = requests.get('http://127.0.0.1:8000/courses/getCourse/COMP1234')
+    assert x.status_code == 400
+
+def test_get_a_course():
+    x = requests.get('http://127.0.0.1:8000/courses/getCourse/COMP1521')
+
+    assert x.status_code == 200
+    assert x.json()['code'] == "COMP1521"
+    assert x.json()['path_from'].get("COMP1511") != None
```
Peedee2002: this should be `is not None`

Peedee2002: there are multiple times where this same thing happens

backend/server/tests/test_searchCourse.py
```diff
@@ -0,0 +1,15 @@
...
+
+def test_search_course():
+    x = requests.get('http://127.0.0.1:8000/courses/searchCourse/COMP1')
+    assert x.status_code == 200
+    assert x.json().get("COMP1521") != None
+    assert x.json().get("COMP2521") == None
+    
+
+def test_invalid():
+    x = requests.get('http://127.0.0.1:8000/courses/searchCourse/ABCD')
```
Peedee2002, line 13: should this not be 400?

backend/server/tests/test_getCourse.py
hdphuong: :oo okok I just google the difference of `== None` and `is None`. Will change this.

backend/server/tests/test_searchCourse.py
hdphuong: hmm from an UX pov, when you search something that doesn't exist, I think it feels nicer to get an empty result rather than an error?

Peedee2002:
Hmm I guess so... it was a "valid" request, so fairOn 17 Mar. 2022 4:24 pm, Mae Vuong ***@***.***> wrote:
@hdphuong commented on this pull request.



In backend/server/tests/test_searchCourse.py:
> @@ -0,0 +1,15 @@
+import json 
+import requests 
+
+
+def test_search_course():
+    x = requests.get('http://127.0.0.1:8000/courses/searchCourse/COMP1')
+    assert x.json().get("COMP1521") != None
+    assert x.json().get("COMP2521") == None
+    
+
+def test_invalid():
+    x = requests.get('http://127.0.0.1:8000/courses/searchCourse/ABCD')

hmm from an UX pov, when you search something that doesn't exist, I think it feels nicer to get an empty result rather than an error?

Peedee2002 (approved):
Yup, nice! LGTM.

PR #220, opened 2022-03-19
made dockerfile pull latest lts, and made a docker service just for prod

Peedee2002:
looks like the issues were possibly ironed out after dependancies were updated.


Snuzzn (approved):
lgtm

docker-compose.yaml
```diff
@@ -40,6 +40,20 @@ services:
       - backend
     networks:
       - circles-net
+  # there is probably a better way to do this, i dont know it.
```
lhjt: In terms of a better way, any idea what you may think? I mean if this is just for the docker-compose, this is fine as is I think.
```suggestion
```

lhjt (changes requested):
image
In terms of commits, seems generally fine (other than one comment) but this also happens to pop up on load? I am assuming that this is just a service bug, and we can commit this anyway considering that this is meant to fix the project not building rather than there being issues like this in the UI.

docker-compose.yaml
Peedee2002: i kinda just didnt want to copy-paste, but if its ok with you, then awesome. Ill remove the comment.

lhjt (approved):

Peedee2002:
> image In terms of commits, seems generally fine (other than one comment) but this also happens to pop up on load? I am assuming that this is just a service bug, and we can commit this anyway considering that this is meant to fix the project not building rather than there being issues like this in the UI.

Hmm this didnt happen for me when i ran the app. We will have to investigate this a bit more then - but otherwise, we can go ahead with this PR.

PR #223, opened 2022-03-20
CF-115 axios default settings

gtangelo:
Included the use of axios default settings for the base url so we can easily change it later if needed (i.e. If deployment version uses a different basurl/port instead of localhost:8000)


frontend/src/axios.js
```diff
@@ -0,0 +1,4 @@
+import axios from 'axios';
+
+axios.defaults.baseURL = 'http://localhost:8000/';
```
lhjt: It'd probably be better to have this as a configurable environment variable instead that is set during build time. This means that we can set the appropriate URL automatically into the production container build pipelines.

Snuzzn (approved):
Looks good! I updated the docker-compose file to take in a new environment variable for the base url (instructions in readme). If this works for you too, merge it in and then I'll let everyone know in general chat to create that env file.

README.md
```diff
@@ -21,6 +21,9 @@ In `mongodb.env`, add:
 - `MONGO_INITDB_ROOT_USERNAME=...`
 - `MONGO_INITDB_ROOT_PASSWORD=...`
 
+In `frontend.env`, add:
+- `REACT_APP_BASE_URL=http://localhost:8000/`
```
lhjt: I'm glad that you have included this as a required environment variable; however, could you add a small sentence/description as to what this should represent? Based on this, it seems to be the URL that the react app is being hosted on, rather than necessarily the API endpoint (for example if the API is running on a different domain).

lhjt: So include something like 
```suggestion
- `REACT_APP_BASE_URL=http://localhost:8000/` (this should be ...)
```

lhjt (changes requested):

gtangelo:
Renamed the `REACT_APP_BASE_URL` to `BACKEND_API_BASE_URL` to make it more clear and added a brief description as what Jared suggested. Also, in the axios configuration, added `http://localhost:8000/` as a fallback base url for the api call in case the environment variable is not specified. :)

Peedee2002 (approved):
Nice, i like this! Just make a quick ping in general to tell everyone to add that env file when you merge this in.

lhjt (approved):
Looks good 👍

PR #234, opened 2022-03-21
Leo/tp/course missing prereq tooltip

leonardo-fan:
- added population of course prereqs from course raw requirements when adding courses via the course description
- edited the formatting and display conditions of the previous prereq tooltip


Snuzzn (approved):
Lgtm!

PR #245, opened 2022-03-23
Peter/hotfix/react 6 breaking changes

Peedee2002:
fix breaking changes from react 6, as well as some spring cleaning


Snuzzn (approved):
Lgtm! Sorry, I should have been have been more careful when allowing the update to react router.

PR #246, opened 2022-03-23
renamed frontend env variable to be recognised by react

gtangelo:
Renamed `BACKEND_API_BASE_URL` env to `REACT_APP_BACKEND_API_BASE_URL` as react only recognises env vars starting with `REACT_APP_...`


Snuzzn (approved):
Ahh yea, that's why I originally had the REACT_APP in front. But it's good that the wording is more specific now.

Peedee2002 (approved):

PR #253, opened 2022-03-23
Peter/circles 252 test planner py

Peedee2002:
done a segnificant number of things including:
- added tests for planner.py

- added a validation test to make sure all our "visible" courses are accurate. This means that the courses are able to be parsed by our backend. there are a list of exceptions collated which involve honours and some nieche courses.
- removed some transitive dependencies from our release.json and requirements.txt (to get dependabot to chill)
- changed name of environment variable in backend.env to make more sense
- fixed some frontend issues with still using localhost urls in some places.


joooooooooooooooooooooooooooooooooooosh (approved):

backend/server/tests/test_validation.py
```diff
@@ -0,0 +1,41 @@
+# we want to assure that courses that may be accessed from a container are always accurately computed.
+from contextlib import suppress
+import requests
+
+from server.tests.test_getAllUnlocked import USERS
+
+# TODO: Do we care if these courses are broken before deployment? they are the honours courses + some wierd courses
```
Peedee2002, line 7: this is an important note

README.md
```diff
@@ -15,7 +15,7 @@ MongoDB and the backend require a few environment variables to get started. In t
 In `backend.env`, add the environment variables:
 - `MONGODB_USERNAME=...`
 - `MONGODB_PASSWORD=...`
-- `MONGODB_DATABASE=mongodb`
+- `MONGODB_SERVICE_HOSTNAME=mongodb`
```
lhjt, line 18: This is a good name 👍

backend/server/tests/test_validation.py
lhjt: If you're going to include them in production, would it be possible to somehow alert users that the courses they have selected are problematic?

lhjt (approved):

backend/server/tests/test_validation.py
Peedee2002: yup @lhjt we have an indication in the frontend that we cant parse them

banaboi (approved):

gtangelo (approved):

Snuzzn (approved):

PR #255, opened 2022-03-24
fixed routing for degree wizard and small styling changes throughout app

Snuzzn:
Made a bunch of small changes to fix some routing issues on the website and some styling:
- Clicking on the circles logo takes user to the degree wizard page.
- The reset planner dialog only appears for existing users (when they click the circles logo). I've also reworded the message and made the button red to make it clear that it's an irreversible action.
- If a page is bookmarked and visited, it goes to the appropriate route after loading (previously, it defaulted to course selector)
- Removed progression checker from navbar
- Removed beta testing notification
- Changed how scrolling works on the course selector. Previously, if the course content is long, a scrollbar will appear for the whole page which looked a janky when scrolling. Now, the scrollbar only applies to the course description. Some of the CSS for the course selector could probably do with some reworking so it relies less on calc() functions with pixels, but this works for now.   
- If the user adds/removes a course, and the quickly moves to a different course, the add/remove alert is now removed rather than display the alert for the incorrect course
- School and faculty are only rendered if the data is present


leonardo-fan:
lgtm! :)

leonardo-fan:
also could be a different branch but I feel like the size of the course selector information is a bit too small and feels kinda constrained to read - what do you reckon?

leonardo-fan (approved):

Snuzzn:
@leonardo-fan You're right. I didnt really notice it before but now that I've increased the font size of the course description, it does look better to me. 
image

Peedee2002 (approved):
lgtm!

PR #257, opened 2022-03-24
Cf-135 - Sano/animated course menu

Snuzzn:
Course menu items are now animated! When courses are unlocked, they will no longer just pop into existence but slide in. 

Kapture 2022-03-24 at 23 06 19


Peedee2002:
Damn that's smooth

Peedee2002 (approved):

PR #258, opened 2022-03-24
Hussain/circles 273 add a linter

imagine-hussain:
Added a linter, configured to use `pylint` with the settings at `.pylintrc`.
Is *somewhat* generous when it comes to certain aspects of the code. See all disabled settings at `disable=` setting in the `pylintrc` accompanied with the reason for disabling.

There isnt really issues left, its mostly a result of missing docstrings for functions, methods and classes. I've done as many as I can do but there are a few that are left over that I don't quite understand. If you're working on the code base and see on the linter log that there is a missing docstring abt smth that you're comfortable describing, please add a docstring

A minor gripe is that not everything is in snake_case. I fixed most of these but, was conservative with changes because some of these were   exports to other  files and these changes can break dependencies. I may actually fix fully on the final PR but, if you see anything in camelCase that you can rename without breaking, pls feel free to.

One of the bigger things is that the Payload that is used by scrapers was moved to its own file and there's a function you can call to create a payload. At the moment, all it requires is just content_type and it fills out the rest but, if you do add extra fields or, arguments to this function, just be wary of making sure you set default values so that it doesnt break the other scrapers / tests.

**Note:** there needs to be a second part of this PR to enable a setting for the pylint pipeline that allows for it to run on pull requests.


.github/workflows/pylinter.yml
```diff
@@ -0,0 +1,52 @@
...
+      # Add specific files / filetypes here.
+      # Use `**` to wildcard through multiply paths
+      - "**.py"
+    # [un]-comment the following line to toggle for push to master
+    # branches-ignore: [master, main]
+
+  # This will only work when this workflow is on default branch
+  # TODO: Uncomment when merging
+  # pull_request:
+  #   branches: [dev, master, main]
```
lhjt, line 26: Should these lines be uncommented now (and have updated branches)?
```suggestion
  pull_request:
    branches: [dev]
```

.github/workflows/ci.yml
```diff
@@ -35,9 +35,8 @@ jobs:
         echo "MONGO_INITDB_ROOT_PASSWORD=github-ci" >> env/mongodb.env
         echo "REACT_APP_BACKEND_API_BASE_URL=http://localhost:8000/" >> env/frontend.env
         docker compose run init-mongo
-
     - name: Test algorithm with pytest
       run: |
         cd backend
         python3 -m pytest algorithms
-        python3 -m pytest server
+        python3 -m pytest server
```
lhjt: Could you add the newline back please

.github/workflows/docker.yml
```diff
@@ -48,4 +48,4 @@ jobs:
           tags: |
             ghcr.io/csesoc/circles-${{ matrix.component }}:${{ github.sha }}
             ghcr.io/csesoc/circles-${{ matrix.component }}:latest
-          labels: ${{ steps.meta.outputs.labels }}
+          labels: ${{ steps.meta.outputs.labels }}
```
lhjt: Could you add the newline back please

lhjt (changes requested):
- Should the pylint configuration files be in the workflows folder or in the root dir?

(You may also want to update this branch)

.github/workflows/pylinter.yml
imagine-hussain: iirc, this needs to be done in two stages. The first (this one) gets the pipeline into dev and then a second quick one just to uncomment this

lhjt: Hmm alright I'll trust that and we can do that afterwards.

.github/workflows/ci.yml
imagine-hussain: added - should show up now

.github/workflows/docker.yml
imagine-hussain: should show up now

lhjt (changes requested):
If you could update this branch to be in line with dev, that'd be great - otherwise lgtm (deploy wise)

lhjt (approved):

PR #269, opened 2022-03-26
fixed some wierdness with generation of conditions and made getAllUnl…

Peedee2002:
…ocked return less stuff. Also, added equivilants to exclusions and gave priority in getStructure

yes, i cant spell equivalents. 

I also made fixes with making sure runprocessors runs without errors and without further modifying the repo.


backend/server/routers/courses.py
```diff
@@ -157,12 +157,13 @@ def getAllUnlocked(userData: UserData):
     for course, condition in CONDITIONS.items():
         # Condition object exists for this course
```
banaboi: Do we want to change/move this comment based on the if logic below?

banaboi (approved):

hdphuong (approved):
nice regex, lgtm

PR #270, opened 2022-03-26
CF139 - Sano/cs/search course with debounce

Snuzzn:
Integrated search endpoint into search bars in course selector and degree wizard (when selecting previous courses).
While doing that, I reworked the debouncing to make the overall logic simpler (Ant design had a weird way of doing it).

Also, now that we're using the search endpoint (rather than getAllUnlocked), we can actually search by course name (instead of course code only), which I think makes a big difference for the user!

Kapture 2022-03-26 at 22 28 33


Peedee2002:
yea i think at some point i implemented searching by title (or at least refactored it to make it nicer) and completely forgot lol
also, we love when we implement new features while *removing* lines of code :pog:

frontend/src/pages/DegreeWizard/steps/PreviousCoursesStep.jsx
```diff
@@ -4,7 +4,7 @@ import { useSelector, useDispatch } from "react-redux";
 import { plannerActions } from "../../../actions/plannerActions";
 import { useNavigate } from "react-router-dom";
 import { EditOutlined } from "@ant-design/icons";
-import DebouncingSelect from "./DebouncingSelect";
+import DebouncingSelect from "./MultiSelectCourse";
```
Peedee2002: im too small-brain to figure out how this import actually worked, but since it works i wont question it lol

Snuzzn: oh yea, that's so weird. I have no idea why that works 🤨. But anyway, I've updated it now. Good catch!

frontend/src/hooks/useDebounce.js
```diff
@@ -0,0 +1,26 @@
+// source: https://usehooks.com/useDebounce/
+import { useEffect, useState } from "react";
+
+export const useDebounce = (value, delay) => {
```
Peedee2002, line 4: instead of something like this i usually like something like https://www.npmjs.com/package/debounce-promise
but if you already have lodash as a dependancy, its not worth it and sticking to this is a good idea.

Peedee2002 (approved):

gtangelo (approved):
lgtm!

PR #271, opened 2022-03-27
Mae/circles 290 legacy courses

hdphuong:


backend/server/routers/courses.py
```diff
@@ -117,6 +130,14 @@ def search(string):
...
+    years = sorted(ARCHIVED_YEARS, reverse=True)
+    if len(list(code_query.clone())) == 0 and len(list(title_query.clone())) == 0:
+        for year in years:
+            code_query = archivesDB[str(year)].find({"code": {"$regex": pat}})
+            title_query = archivesDB[str(year)].find({"title": {"$regex": pat}})
+            if len(list(code_query.clone())) != 0 or len(list(title_query.clone())) != 0:
+                break
+
     return {
         course["code"]: course["title"] for course in chain(code_query, title_query)
```
Peedee2002, line 144: this isnt part of your code, but i forgot i implemented this lol

backend/server/routers/courses.py
```diff
@@ -117,6 +130,14 @@ def search(string):
     code_query = coursesCOL.find({"code": {"$regex": pat}})
     title_query = coursesCOL.find({"title": {"$regex": pat}})
 
+    years = sorted(ARCHIVED_YEARS, reverse=True)
+    if len(list(code_query.clone())) == 0 and len(list(title_query.clone())) == 0:
+        for year in years:
+            code_query = archivesDB[str(year)].find({"code": {"$regex": pat}})
+            title_query = archivesDB[str(year)].find({"title": {"$regex": pat}})
+            if len(list(code_query.clone())) != 0 or len(list(title_query.clone())) != 0:
```
Peedee2002: is there a reason we have to clone here? does len(code_query) not work? Im worried that this may be an expensive computation as we expand our db.

backend/server/routers/courses.py
```diff
@@ -91,8 +92,20 @@ def fixUserData(userData: dict):
...
 def getCourse(courseCode: str):
-    """get info about a course given courseCode"""
+    """get info about a course given courseCode
+    - start with the current database
+    - if not found, check the archives
+    """
     result = coursesCOL.find_one({"code": courseCode})
+
+    if not result:
+        years = sorted(ARCHIVED_YEARS, reverse=True)
```
Peedee2002: its probably not worth making this variable, you should probably make `sorted` part of the for loop directly. If youre worried about performance, `sorted` is a generator anyways, so the performance here is identical.

Peedee2002 (approved):
nice! small changes, but well done overall, and good job on adding tests!

backend/server/routers/courses.py
Peedee2002: (another option can be a simple `if code_query or title_query:`

backend/server/routers/courses.py
```diff
@@ -117,6 +130,14 @@ def search(string):
     code_query = coursesCOL.find({"code": {"$regex": pat}})
     title_query = coursesCOL.find({"title": {"$regex": pat}})
 
+    years = sorted(ARCHIVED_YEARS, reverse=True)
+    if len(list(code_query.clone())) == 0 and len(list(title_query.clone())) == 0:
```
Peedee2002: wait i missed this - we want to include this query in all cases, and indicate that it is legacy. this means that we should be returning an 'is_legacy' for each course.

Peedee2002 (changes requested):
sorry - i clarified a detail that needed to be included in this PR. I must have been unclear in the card, sorry about that.

backend/server/routers/courses.py
hdphuong: oh ok i'll try that. i thought the default `false` behaviour only works for `None` value not on an empty iterator but will try to see if it works. Also to answer your question above, len(code_query) would consume the iterator and result in an empty list when we chain it so we need to clone() it.

Peedee2002 (approved):
yup, nice!

Peedee2002:
@Snuzzn we will soon update the getCourse endpoint to have is_legacy.

Snuzzn (approved):

PR #272, opened 2022-03-27
279 Add CompositeCategory to conditions

joooooooooooooooooooooooooooooooooooosh:
This allows for a simpler implementation of LevelCourseCategory, as well as being able to parse more complex UOC/WAM/Grade conditions consisting of lists of courses (e.g. '12UOC in (COMP1511 || COMP1521 || COMP1531)').

This also contains the updated versions of courses for the 2022 handbook, some of which have introduced new parsing issues (such as COMM1190).


lhjt (approved):
Thanks for adding more docs and type annotations :)

I think things look okay? (I'm assuming you're going to address the failed test)

joooooooooooooooooooooooooooooooooooosh:
> Thanks for adding more docs and type annotations :)
> 
> I think things look okay? (I'm assuming you're going to address the failed test)

Yep I'm just trying to figure out the correct manual fixes

Peedee2002 (changes requested):
the pure raw files seem to have been deleted. that change should probably be reverted. I think this is actually why you were able to figure out the issues that were hidden before (thanks for that <3)

Peedee2002:
actually now that i think about it (again) - we may want to keep your changes - can you tell me how you got these different raw files?

joooooooooooooooooooooooooooooooooooosh:
> actually now that i think about it (again) - we may want to keep your changes - can you tell me how you got these different raw files?

Honestly I'm not entirely sure. I think it was from running `python3 -m runprocessors --type program --stage all` and such, which reran the scrapers

Peedee2002:
> > actually now that i think about it (again) - we may want to keep your changes - can you tell me how you got these different raw files?
> 
> Honestly I'm not entirely sure. I think it was from running `python3 -m runprocessors --type program --stage all` and such, which reran the scrapers

Ahhh fair enough. data-fix usually ignores the scrapers because we would spam the unsw servers. Looks like there were changes that happened between the last time we ran it though, so good catch on that.

backend/algorithms/create.py
```diff
@@ -46,8 +49,9 @@
     CACHED_EXCLUSIONS = json.load(f)
 
 
-def create_category(tokens):
-    """Given a list of tokens starting from after the connector keyword, create
+def create_category(tokens) -> Category:
```
Peedee2002: this typehint is incorrect - we return a category and an int of how many tokens were used

backend/algorithms/create.py
```diff
@@ -81,37 +124,41 @@ def create_category(tokens):
...
     Given the parsed logical tokens list (assuming starting and ending bracket),
     and optionally a course for which this condition applies to,
-    Returns the condition
+    returns the condition
     """
     return make_condition(tokens, True, course)[0]
 
 
-def make_condition(tokens, first=False, course=None):
+def make_condition(tokens, first=False, course=None) -> CompositeCondition:
```
Peedee2002: this is also incorrect - we return a compositeCondition and how many tokens we consumed.

joooooooooooooooooooooooooooooooooooosh:
> > > actually now that i think about it (again) - we may want to keep your changes - can you tell me how you got these different raw files?
> > 
> > 
> > Honestly I'm not entirely sure. I think it was from running `python3 -m runprocessors --type program --stage all` and such, which reran the scrapers
> 
> Ahhh fair enough. data-fix usually ignores the scrapers because we would spam the unsw servers. Looks like there were changes that happened between the last time we ran it though, so good catch on that.

Yeah that's a good point. I think the large decrease in raw file sizes is due to courses that don't exist anymore in the 2022 handbook, do you want to keep these changes?

backend/algorithms/create.py
```diff
@@ -168,67 +215,42 @@ def make_condition(tokens, first=False, course=None):
...
-            # Condition for UOC requirement
-            uoc = get_uoc(token)
-            uoc_cond = UOCCondition(uoc)
+        elif is_program(token):
+            result.add_condition(ProgramCondition(token))
+        elif is_specialisation(token):
+            result.add_condition(SpecialisationCondition(token))
+        elif is_program_type(token):
+            result.add_condition(ProgramTypeCondition(token))
+        else:
```
Peedee2002, line 224: ??? wierd style

Peedee2002:
> > > > actually now that i think about it (again) - we may want to keep your changes - can you tell me how you got these different raw files?
> > > 
> > > 
> > > Honestly I'm not entirely sure. I think it was from running `python3 -m runprocessors --type program --stage all` and such, which reran the scrapers
> > 
> > 
> > Ahhh fair enough. data-fix usually ignores the scrapers because we would spam the unsw servers. Looks like there were changes that happened between the last time we ran it though, so good catch on that.
> 
> Yeah that's a good point. I think the large decrease in raw file sizes is due to courses that don't exist anymore in the 2022 handbook, do you want to keep these changes?

yup sounds good to me

backend/algorithms/objects/conditions.py
```diff
@@ -156,41 +166,74 @@ def get_warning(self, applicable_wam):
...
-        # Course code
-        self.course = course
+    def set_category(self, category_classobj: Category):
+        """ Set own category to the one given """
+        self.category = category_classobj
+
+    def validate(self, user: User) -> tuple[bool, list[str]]:
+        def _validate_course(course: Category):
+            # Grade condition can only be used with ClassCategory
+            if type(course) is CompositeCategory:
```
Peedee2002, line 187: is this not guaranteed?

Peedee2002 (changes requested):
a few more wierd quirks in your code ive spotted, but otherwise OK

backend/algorithms/create.py
joooooooooooooooooooooooooooooooooooosh: whoops!

backend/algorithms/objects/conditions.py
joooooooooooooooooooooooooooooooooooosh: no because you can have a CompositeCategory of CompositeCategories

backend/algorithms/create.py
joooooooooooooooooooooooooooooooooooosh: are you talking about the else statement?

backend/algorithms/errors.json
```diff
@@ -7211,127 +7578,121 @@
         ],
         "broke at": "Index 6, /CODE"
     },
-    "COMP3901": {
+    "COMM1120": {
```
Peedee2002: im not sure how this passed our test, but i think this should be a manual fix. Sorry that i have to put this on you to do lol i just only spotted it in this pr

backend/algorithms/errors.json
```diff
@@ -7211,127 +7578,121 @@
...
+            "standing",
+            "&&",
+            "COMM1999",
             ")"
         ],
-        "broke at": "Index 1, ("
+        "broke at": "Index 3, be"
     },
-    "COMP9491": {
+    "COMP4951": {
```
Peedee2002, line 7581: oh no the UOC bug strikes again... Ill make sure to look at this after you are done with this PR.

backend/algorithms/errors.json
```diff
@@ -7929,6 +8323,19 @@
         ],
         "broke at": "Index 1, Honours"
     },
+    "DESN1900": {
+        "condition": {
+            "original": "No prerequisites required<br/><br/>",
```
Peedee2002: i hate this person

backend/algorithms/errors.json
joooooooooooooooooooooooooooooooooooosh: hmm i feel like this should be fixed as part of my PR, e.g. AERO4110 broke as well

backend/algorithms/errors.json
joooooooooooooooooooooooooooooooooooosh: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA

backend/algorithms/create.py
Peedee2002: yea... there is an else statement followed by if statements that look similar

backend/algorithms/errors.json
Peedee2002: sure :)

backend/algorithms/errors.json
joooooooooooooooooooooooooooooooooooosh: will do, also looks like the processing stage for COMM3091 is buggy. I think we're normally able to process conditions like that

backend/algorithms/create.py
joooooooooooooooooooooooooooooooooooosh: It's because those if statements in the else all shared a large chunk of code (the remainder of the else statement).
A little weird to read but avoids a lot of duplication

Peedee2002: oh my bad - thats ok then

backend/algorithms/objects/conditions.py
Peedee2002: yup makes sense

Peedee2002:
Cool that is most of my comments - there is a few things left, and it also seems that you need to push your branch forward. but other than that, good work :) Ping me again when you have finished the stuff with AERO and COMP.

backend/algorithms/errors.json
joooooooooooooooooooooooooooooooooooosh: nvm this isn't a UOC issue, it's an honours program issue.

Peedee2002: Ah cool

Peedee2002 (approved):
yup nice work :)
thanks for picking up a lot of bugs on accident <3

PR #274, opened 2022-03-28
Cf-143: Styling changes mainly for smaller (laptop) screens

Snuzzn:
- course selector bar is now much smaller 
- term planner container is now scrollable, so if it exceeds the height of the screen, the navbar remains fixed 
- added an infographic to course selector
- fixed warning symbol sizing issue (it was inconsistent between different courses)

image

image


Peedee2002 (approved):

Snuzzn:
@leonardo-fan  I also tried going for a more flat approach to the tp buttons because I thought it would be more in line with the rest of our design. Wdyt?
image

leonardo-fan:
> @leonardo-fan I also tried going for a more flat approach to the tp buttons because I thought it would be more in line with the rest of our design. Wdyt? image

Yeah definitely looks better, much more consistent!

leonardo-fan (approved):

gtangelo (approved):
lgtm!!

PR #280, opened 2022-03-28
Course selector loading skeleton runs until data fetched

Snuzzn:
Previously, the loading skeleton was set to 2 seconds. It's now been changed, so that it turns off when the getCourse info is loaded. 
I also created a helper function (inside axios.js) to handle our requests to the backend in a simpler way. So we no longer need to run a lengthy try catch block each time we do a request.


frontend/src/components/header/DrawerContent.jsx
```diff
@@ -26,16 +26,17 @@ export const DrawerContent = ({ onCloseDrawer }) => {
...
       </Menu.Item>
       <Menu.Item key="3" onClick={() => handlePush("/term-planner")}>
         Term Planner
       </Menu.Item>
-      /* <Menu.Item key="2" onClick={() => handlePush("/progression-checker")}>
+      <Menu.Item key="2" onClick={() => handlePush("/progression-checker")}>
         Progression Checker
-      </Menu.Item> */
+      </Menu.Item>{" "}
+      */
```
gtangelo: May want to remove `*/` and `{" "}` since its not needed?

frontend/src/axios.js
```diff
@@ -1,6 +1,25 @@
 import axios from "axios";
 
 // fallback to localhost port 8000 if env variable is not specified
-axios.defaults.baseURL = process.env.REACT_APP_BACKEND_API_BASE_URL || 'http://localhost:8000/';
+axios.defaults.baseURL =
+  process.env.REACT_APP_BACKEND_API_BASE_URL || "http://localhost:8000/";
 
 axios.defaults.headers.post["Content-Type"] = "application/json";
+
+export const axiosRequest = async (method, url, body) => {
```
gtangelo, line 9: Nice that you abstracted the axios request to make it into its own function!!

gtangelo (changes requested):

gtangelo (approved):

PR #283, opened 2022-03-28
rehauled getStructure and courseSelector side panel

Peedee2002:
made getStructure endpoint behave better, showed locked courses and sorted courses on weather or not they were locked or selected, and added padlocks to locked courses.

https://user-images.githubusercontent.com/64952797/160516344-949c757d-205c-408b-904b-896c77a9ba00.mp4

this should help @SpanishPear with the confusion of why he couldnt see certain courses, as this toggle obviously tells the user why they may not see a course.


spanishpear:
Will setup circles locally tn and have a play - have some comments on UI but will wait till home this arv

frontend/src/pages/CourseSelector/courseMenu/CourseMenu.jsx
```diff
@@ -166,14 +163,19 @@ export default function CourseMenu({ structure }) {
       )}
     </div>
   );
+  function sortMenu(item1, item2) {
```
Snuzzn, line 166: In terms of UI, it might also be a good idea to sort by courseCode within unlocked/locked sections. For example:

Before
before

After
after

The 'after' image looks more clear to me. Sorting by courseCode basically separates it into faculty (e.g. COMP, MATH), then by level within. So everything looks nice and clean.

```js
function sortMenu(item1, item2) {
    return item1.unlocked === item2.unlocked
      ? item1.courseCode > item2.courseCode // sort within locked/unlocked by courseCode
      : item1.unlocked < item2.unlocked; // separate locked/unlocked
}

File: frontend/src/pages/CourseSelector/courseMenu/CourseMenu.less [hunk 2]
```diff
@@ -32,6 +32,10 @@
   font-weight: 700;
 }
 
+.locked {
```
Snuzzn, line 35: Personally, I think font-weight 100 looks so thin that it affects legibility. But if we do increase it, I think some experimentation may be needed here to maintain a visual difference between unlocked, available and selected courses. Fe team can tinker around with this later.

Snuzzn (approved):
Besides that sorting suggestion, it looks good to me. Also, I'm guessing you'll work on the getStructure minor endpoints later e.g. '/getStructure/3778/COMPA1/MARKA2' doesn't work atm.
Toggle idea also works really well!

frontend/src/pages/CourseSelector/courseMenu/CourseMenu.jsx
Peedee2002: done

Peedee2002:
thanks for catching the marketing bug btw @Snuzzn - i definitely missed that

PR #302, opened 2022-04-02
Circles 292 refactor program type creation to use a config file

imagine-hussain:
Moved filepaths out of `cachy.py` and into `cache_config.py`.

Refactored `cache.py` so that program -> code mapping rules are stored in `cache_config.json`
This json has two fields. `"codes"` and `"keywords"`
- "codes" is a list of valid codes
- "keyword_mapping" is a dict where the key is a keyword and the value is the codes that keyword maps to


backend/algorithms/cache/cache_config.json
```diff
@@ -0,0 +1,50 @@
...
+        "SOSS#",
+        "DDES#",
+        "ARTS#",
+        "DART#",
+        "MUSC#",
+        "SOCW#",
+        "PPEC#",
+        "EDST#" 
+    ],
+    "keyword_mapping": {
```
joooooooooooooooooooooooooooooooooooosh, line 27: Since these keys are being `.lowered()` anyway, I think we may as well capitalize the keys to reflect how they appear in the handbook

imagine-hussain: Do you mean just move them all to `Title Case`?

joooooooooooooooooooooooooooooooooooosh: Yeah pretty much

imagine-hussain: changes made

joooooooooooooooooooooooooooooooooooosh (approved):

PR #305, opened 2022-04-03
Mae/cf 128 legacy and rested courses

hdphuong:
- isLegacy warning is displayed
- users are allowed to drop courses wherever they want
- summer courses debug

(not yet implemented: proper green highlighting for legacy/rested courses)

https://user-images.githubusercontent.com/63543383/161429864-51c4a824-e7af-4ad1-9747-9fb2be3abebc.mov


Peedee2002 (approved):
Very nice. For the green highlighting, feel free to make a new card for it. This card was probably too big anyways.


A few questions tho:
What happens if a course is not offered in that term, but is not a legacy course?
What happens if a legacy course is placed in a term it was not offered in?

hdphuong:
> 

Hmm yeah, we need to tell the users somehow either by blocking or warning. I think it feels nicer to block since we already have tons of warnings in place. But if we block then dealing with future terms is gonna be real pain :( 
What do you think?

Peedee2002:
> > 
> 
> Hmm yeah, we need to tell the users somehow either by blocking or warning. I think it feels nicer to block since we already have tons of warnings in place. But if we block then dealing with future terms is gonna be real pain :( What do you think?

Yea, I don't think we should block it, just in case we are wrong. Warnings are probably best imo. We can make a seperate card for properly dealing with warnings on both backend and frontend

Peedee2002:
if sano doesnt get around to reviewing this PR, ill add changes to get the green to show up correctly - since my changes are now in dev.

hdphuong:
yup, sounds good to me 👍

Snuzzn (approved):
sorry, missed this. Lgtm!

PR #306, opened 2022-04-03
Leo/navbar/menu overflow fix

leonardo-fan:
- fixed a bug where zooming in and out of page would make navbar pages collapse even when screen size was large
- fixed bug where accessing side menu from hamburger menu did not show the report a bug button when screen size is increased


imagine-hussain (approved):
lgtm

PR #308, opened 2022-04-04
Leo/planner cart/course links

leonardo-fan:
- added links back to courses in planner cart
- fixed bug where the highlighted header page didn't change when not changing via the header (useEffect using the window path address)


frontend/src/pages/CourseSelector/SearchCourse.jsx
```diff
@@ -8,7 +8,7 @@ import { useDebounce } from "../../hooks/useDebounce";
 
 export default function SearchCourse() {
   const [value, setValue] = useState("");
-  const debouncedSearchTerm = useDebounce(value, 200);
+  const debouncedSearchTerm = useDebounce(value, 500);
```
Snuzzn: I recently changed this to 200 to reduce the delay :).

Snuzzn (approved):
lgtm!

frontend/src/pages/CourseSelector/SearchCourse.jsx
leonardo-fan: Ohh I see - I think when I merged dev it didn't take in the new changes for some reason, I'll go in to fix them up

PR #321, opened 2022-04-05
Mae/circles 309 proper warnings

hdphuong:
- Propagate warnings and handbook_note from the backend to the frontend
- Divide warnings into 2 types: important vs not so important warnings.

https://user-images.githubusercontent.com/63543383/161741060-ca5ccbbd-6557-4a36-8d21-6562e5cae3ac.mov


Peedee2002 (approved):
yoo hype ez

frontend/src/reducers/plannerReducer.js
```diff
@@ -111,9 +113,12 @@ const plannerReducer = (state = initialState, action) => {
 
     case "TOGGLE_WARNINGS":
       let coursesCpy = new Map(state.courses);
+      console.log(coursesCpy);
```
leonardo-fan: can remove for less console clutter, but otherwise looks good! 😎

hdphuong: Oops i forgot. Thanks for pointing that out 🙏

PR #327, opened 2022-04-09
CF-154 reset tabs after resetting degree planner

gtangelo:
Fix bug where course selector tabs would not be cleared when user resets the degree planner.


Peedee2002 (approved):
nice!

PR #328, opened 2022-04-09
Peter/cf 167 eslint

Peedee2002:
Ive added a lot of changes, but not enough to completely satisfy ESLint. Ive 
- added a config for ESLint
-  tried to fix up what i can by
     -  cleaning up the easy warnings
     -  getting rid of the JS compiler warnings too. 

I have added a test to the pipeline to ensure that the frontend compiles in CI.

just so im not stupid, please pull this branch and validate it - just to be sure it works.


frontend/.eslintrc.js
```diff
@@ -0,0 +1,28 @@
...
+    },
+    ecmaVersion: "latest",
+    sourceType: "module",
+  },
+  plugins: [
+    "react",
+  ],
+  rules: {
+    quotes: ["error", "double"],
+    "react/jsx-props-no-spreading": [0],
```
gtangelo: Maybe instead of using `[0]`, use `["off"]` to make it consistent and readably for configuring the eslint rules (same goes for `[2]` to `["error"]`)?

gtangelo (changes requested):
Note: Having issues running the frontend via docker due to some errors (I think mostly relating to some import issues with default and non default exports i.e. some components are imported as a non-default components when the component are specified as default exports).

So maybe merging into master might not be a good idea right now if the frontend can't be run lol

Peedee2002:
> 

Oh god I thought I dealt with those. I'll fix that then.

Peedee2002:
ok done messing with this branch :)

frontend/src/pages/DegreeWizard/steps/DegreeStep.jsx
```diff
@@ -31,15 +31,15 @@ export const DegreeStep = ({ incrementStep, currStep }) => {
...
+      }),
     );
   };
 
   const props = useSpring(springProps);
 
   return (
     <animated.div style={props}>
-      <div className="steps-heading-container">
+      <div className={styles.stepsHeadingContainer}>
```
gtangelo: Don't think importing styles as an object would work since its not a styled component. Perhaps just use the css class directly

gtangelo (changes requested):
Thanks for doing this Peter! Everything at the code level seems to be good. There seems to be some bugs in the FE lol

- Not too sure if this is a FE/BE bug but this error message seems a bit too long.
image
- For the help popup on the term selector, not entirely sure why the css styles are different to prod. I guess in `HelpMenu.jsx`, it was using the `class` attribute rather than `className` which caused antd css to mess up? Could you create a ticket on the FE board so that the FE team could fix this?
image

frontend/src/pages/DegreeWizard/steps/DegreeStep.jsx
gtangelo: Causes this css styling issue 
image

Peedee2002:
@gtangelo for some of these bugs (eg the 405 error and getCourse causing a crash, and possibly the "inaccurate" errors) i suspect this is because you didnt re-build the backend. You need to kill whatever backend youre currently running and then run `docker compose build backend`
when i did that those errors went away.

Also, that 405 bug was supposed to get caught in CI, so if it persists, ill be very worried lol
I did fix the css issues though.

gtangelo:
Oh yep your right @Peedee2002. My bad on that part. I've updated my comments to the bugs I see on my end

gtangelo:
Also weird thing I notice is that when you add the course to the term planner, the CourseDescription component seems to be rerendered everytime when you add the course via the `Add to planner` button

Peedee2002:
> Also weird thing I notice is that when you add the course to the term planner, the CourseDescription component seems to be rerendered everytime when you add the course via the `Add to planner` button

Yea, that needs to happen because the "getCoursesUnlockedWhenTaken" endpoint needs to be recalled (right at the bottom) so we need a rerender to make that happen.

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -91,17 +86,18 @@ export default function CourseDescription({ structure }) {
...
-      });
+      if (!err) {
+        setCoursesPathTo({
+          direct_unlock: data.direct_unlock,
+          indirect_unlock: data.indirect_unlock,
+        });
+      }
     };
 
     setpageLoaded(false);
```
gtangelo, line 103: I checked prod and when clicking the `Add to planner` button, everything seems smooth. Where as with this branch, clicking the `Add to planner` button, the course description rerenders and briefly shows the loading skeleton due to the `pageLoaded` state (as it was set to false before setting it to true after finishing calling the endpoint). May not offer the best UX experience as the page loading is dependent on the network speed.

This may have been a bug before you eslinted everything and it wasn't picked up as there were less dependencies in the useEffect call?

Peedee2002: yea, i think that was it. We can probably figure out a way to make it not refresh if we modularise a bit more. I can try my hand at fixing that for now.

gtangelo: Yea all right. I was thinking since this was kinda out of the scope of the eslint ticket (since its more of a logic issue), we can try to create a bug ticket for it for the FE team to fix it in another branch, since this ticket is already big enough haha. But totally up fine fr you to make the call on that 😄. Otherwise, everything with the eslint changes looks good to me!

gtangelo (approved):

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
Peedee2002: @gtangelo https://compclub.atlassian.net/browse/CF-186

PR #332, opened 2022-04-10
Gabriel/cs/cf 163 close all tabs

gtangelo:
Added a dropdown menu to the course tabs to allow users to close all tabs when clicking the `Close all tabs` button.

image
image
image


Peedee2002 (approved):
nice and simple 😎

frontend/src/pages/CourseSelector/courseTabs/CourseTabs.jsx
```diff
@@ -53,6 +67,12 @@ export const CourseTabs = () => {
           )}
         </Droppable>
       </DragDropContext>
+      {
+        !!tabs.length &&
+        <Dropdown overlay={TabDropdownMenu}>
+          <Button className="cs-tabs-dropdown-btn" icon={<EllipsisOutlined />}/>
```
Peedee2002: not sure if this icon is the best option though. possibly we just have it be a single click delete?

gtangelo: Hmm yea.. There's not much UI convention for an 'Close all tabs' button (not that I know of). On chrome, you can close tabs either with a shortcut or a right click on the tab. I thought having it like an ellipsis icon would give a similar meaning to a right click.

If we want to go with a single click delete approach, I feel that a very powerful action to mass delete tabs if the user did not intend for that to happen.

I guess another approach is to change the icon to a single click delete but instead a warning popup appears something similar to this?
image

I prefer either using the elipsis and having a dropdown to close all tabs or changing the icon and having a popup for a warning. Thoughts?

Peedee2002: I think it depends on if we want to add more functionality under those ellipses.

If we do, then keeping it this way is reasonable, but if not, then I think changing the icon makes the most sense

gtangelo:
Made close all tabs as a popup instead
image

Peedee2002 (approved):

PR #338, opened 2022-04-11
switch from regex to fuzzy matching

joooooooooooooooooooooooooooooooooooosh:
This now includes weightings based on the user data


backend/requirements.txt
```diff
@@ -5,6 +5,7 @@ charset-normalizer==2.0.12
 click==8.1.2
 dnspython==2.2.1
 fastapi==0.75.1
+fuzzywuzzy==0.18.0
```
imagine-hussain, line 8: fuzzywuzzy uses the Levenshtein distance for weightings and is considerably slower without the package and returns this error:
UserWarning: Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning

Adding the latest version of it into the requirements should fix this.

imagine-hussain (changes requested):

backend/server/routers/courses.py
```diff
@@ -1,6 +1,7 @@
 import re
```
imagine-hussain: This import is now unused.

backend/server/routers/courses.py
```diff
@@ -129,23 +131,11 @@ def search(string):
...
 
-    if not code_query and not title_query:
-        for year in sorted(ARCHIVED_YEARS, reverse=True):
-            code_query = list(archivesDB[str(year)].find({"code": {"$regex": pat}}))
-            title_query = list(archivesDB[str(year)].find({"title": {"$regex": pat}}))
-            if code_query or title_query:
-                break
-
-    return {
-        course["code"]: course["title"] for course in chain(code_query, title_query)
```
imagine-hussain, line 154: `from itertools import chain` is no longer needed on line 2

Peedee2002:
you seem to be (predictably) failing some search tests - youre gonna need to update them to match the new behaviour.

Peedee2002:
you may need to change the Dockerfile to correctly install your dependancies in docker. 
EDIT: im dumb you did that already

backend/server/routers/courses.py
```diff
@@ -129,23 +134,36 @@ def search(string):
...
-    }
+    # TODO: can you have a minor with no major in Circles?
+    if len(specialisations) == 1:
+        major = specialisations[0]
+    elif len(specialisations) == 2:
+        major = specialisations[0]
+        minor = specialisations[1]
+
+    structure = getStructure(userData.program, major, minor)['structure']
```
Peedee2002: this is an annoying comment from me - but im curious if you can also do `getStructure(userData.progam, *specialisations)` I think that works and gets rid of the messiness from earlier. If it doesnt - then im happy to add default args to getStructure to make that work.

backend/server/routers/courses.py
```diff
@@ -129,23 +135,45 @@ def search(string):
...
+    if ALL_COURSES is None:
+        ALL_COURSES = [{"title": course["title"], "code": course["code"]} for course in coursesCOL.find()]
+
+    # TODO: consider ways to search legacy courses
+    # TODO: can you have a minor without a major selected?
+    #       will wreak havoc on the argument order with *specialisations
+    specialisations = list(userData.specialisations.keys())
+    structure = getStructure(userData.program, *specialisations)['structure']
+
+    top_results = sorted(ALL_COURSES, reverse=True,
```
Peedee2002: so we will probably need to add legacy courses back. to do this, you should chain the iterables together, and give a low/minimal weighting to those legacy courses.

Peedee2002 (changes requested):
yup after adding back legacy courses and the merge conflict, we should be good!

backend/server/routers/courses.py
joooooooooooooooooooooooooooooooooooosh: The whole legacy courses structure is going to need an overhaul for this. It worked for the previous implementation since you'd only check legacy years if the course didn't exist in the current year, but since we're fuzzy matching and will have to add all the previous years courses this gets problematic due to all of the duplicates from courses that have been in the handbook for multiple years.

joooooooooooooooooooooooooooooooooooosh: A side effect of all the duplicates is that the search call is noticeably slower when running the weighting algos over every course from every year

joooooooooooooooooooooooooooooooooooosh: Unless you have a straightforward solution, this might be better left as a separate issue.

Peedee2002: hmm would it be simple to do a `fetch_latest_offering` for all courses? So basically you look through the archive, and if the course is not currently in the list of all courses, its added - with the list traversed backwards. This might be expensive, but its probably only going to need to be done once. This helps get rid of duplicates. Note: looking through the archive shouldnt add too many courses, so we shouldnt see that much of a performance hit on the scoring side

joooooooooooooooooooooooooooooooooooosh: I'll look into it. While I'm doing this, do we want to indicate to the user that the course is a legacy course somehow?

Peedee2002: @joooooooooooooooooooooooooooooooooooosh I think we do this in 2 ways - we have a note in the course selector and another in degree planner (which is a warning). Should we add an icon in search as well?

joooooooooooooooooooooooooooooooooooosh: potentially, i'll think about it

backend/server/routers/programs.py
```diff
@@ -136,11 +136,12 @@ def getMinors(programCode: str):
     return {"minors": minrs}
 
 def convertSubgroupObjectToCoursesDict(object: str, description: str|list[str]) -> dict[str, str]:
-    """ gets a subgroup object (format layed out in the processor) and fetches the exact courses its referring to """
+    """ gets a subgroup object (format laid out in the processor) and fetches the exact courses its referring to """
     if " or " in object:
         return {c: description[index] for index, c in enumerate(object.split(" or "))}
     elif not re.match(r"[A-Z]{4}[0-9]{4}", object):
-        return search(object)
+        # TODO(josh): ensure this is accurate
```
joooooooooooooooooooooooooooooooooooosh: @Peedee2002 when does this actually get used? I can't quite figure out the intended purpose of this

Peedee2002: its an artifact of how we store program data. Some strings show up like:
"COMP9" - this indicates any level 9 course
"COMP3121 or COMP3821" - this means that we can pick one of two. 
This is supposed to match via regex.

Maybe the correct solution is to push this logic to the data section actually.

joooooooooooooooooooooooooooooooooooosh: Yeah that might be a good idea but sounds like I should bring back the regex based search for this case in the meantime

Peedee2002: > Yeah that might be a good idea but sounds like I should bring back the regex based search for this case in the meantime

sounds good. We should make it unavailable to the frontend tho.

joooooooooooooooooooooooooooooooooooosh: If you mean not providing an endpoint to the function, this is already the case

joooooooooooooooooooooooooooooooooooosh: Are there any cases where this would require matching on the title of the course or is it only ever the course codes?

Peedee2002: only codes.

backend/algorithms/exampleUnselectCourse.json
```diff
@@ -2,7 +2,7 @@
     "userData": {
```
imagine-hussain: Looks like `exampleUnselectCourse.json` just contains one example user. There already exists `exampleUsers.json` in the same directory as: `/backend/algorithmns/exampleUsers.json`
Ideally, this user would get added to that file instead so that this file can be deleted.

backend/server/tests/courses/test_searchCourse.py
```diff
@@ -1,20 +1,27 @@
 import requests
+import json
+
+with open("algorithms/exampleUnselectCourse.json") as f:
```
imagine-hussain: pt2 to the comment earlier about using `exampleUsers.json` instead

imagine-hussain (changes requested):

backend/server/routers/courses.py
```diff
@@ -136,23 +180,47 @@ def search(string):
...
+                         key=lambda course: fuzzy_match(course, search_string)
+                         )[:100]
+    weighted_results = sorted(top_results, reverse=True,
+                              key=lambda course: weight_course(course,
+                                                               search_string,
+                                                               structure, 
+                                                               *specialisations)
+                              )[:30]
+
+    return {course[0]: course[1] for course in weighted_results}
```
Peedee2002: This is mean, but you should use tuple decomposition here.

backend/server/routers/courses.py
```diff
@@ -322,5 +393,56 @@ def coursesUnlockedWhenTaken(userData: UserData, courseToBeTaken: str):
...
+               sum([fuzz.partial_ratio(title.lower(), word)
+                       for word in search_term.split(' ')]))
+
+def weight_course(course: tuple, search_term: str, structure: dict,
+                  major_code: Optional[str] = None, minor_code: Optional[str] = None):
+    """ Gives the course a weighting based on the relevance to the user's degree """
+    weight = fuzzy_match(course, search_term)
+    (code, title) = course
+
+    # TODO: integrate electives weighting
```
Peedee2002: Does this comment still matter?

backend/server/routers/programs.py
```diff
@@ -136,22 +136,28 @@ def getMinors(programCode: str):
...
-    """returns the added courses"""
+    """ Returns the added courses """
     structure[type][container["title"]] = {}
     item = structure[type][container["title"]]
 
     item["UOC"] = container["credits_to_complete"]
     item["courses"] = {}
     for object, description in container["courses"].items():
+        if " or " in object:
+            courses_mentioned = {c: description[index] for index, c in enumerate(object.split(" or "))}
```
Peedee2002: Where is this variable used?

backend/server/routers/courses.py
joooooooooooooooooooooooooooooooooooosh: Wasn't sure how to do that in a list comprehension

backend/server/routers/courses.py
joooooooooooooooooooooooooooooooooooosh: To a degree, we could consider weighting electives differently to core courses but it's probably not important enough for a comment

backend/server/routers/programs.py
joooooooooooooooooooooooooooooooooooosh: bro i swear i never wrote this part of the code it just appeared there lmao

joooooooooooooooooooooooooooooooooooosh: but ur right this looks like useless code

joooooooooooooooooooooooooooooooooooosh: literally have no idea how it got there

backend/server/routers/courses.py
Peedee2002:
```python
return {code: name for code, name in weighted_results}
```

joooooooooooooooooooooooooooooooooooosh: oh yeah im just dumb

backend/server/routers/courses.py
```diff
@@ -322,5 +393,56 @@ def coursesUnlockedWhenTaken(userData: UserData, courseToBeTaken: str):
...
+    (code, title) = course
+
+    # either match against a course code, or match many words against the title
+    # (not necessarily in the same order as the title)
+    search_term = search_term.lower()
+    if re.match('[a-z]{4}[0-9]', search_term):
+        return fuzz.ratio(code.lower(), search_term)
+
+    return max(fuzz.ratio(code.lower(), search_term),
+               sum([fuzz.partial_ratio(title.lower(), word)
```
Peedee2002: You don't need the brackets here. You can pass a generator directly into sum. I usually wouldn't care, but this may be memory intensive if we decide to store the whole list in memory.

joooooooooooooooooooooooooooooooooooosh: cool

backend/server/routers/courses.py
```diff
@@ -136,23 +180,47 @@ def search(string):
...
+
+def regex_search(search_string: str):
+    """ 
+    Uses the search string as a regex to match all courses with an exact pattern.
+    """
+
+    pat = re.compile(search_string, re.I)
+    courses = list(coursesCOL.find({"code": {"$regex": pat}}))
+
+    # TODO: do we want to always include matching legacy courses (excluding duplicates)?
```
Peedee2002, line 216: The answer to this question is yes. We should however score them slightly lower.

joooooooooooooooooooooooooooooooooooosh: These aren't being scored, this is the lookup function used in addSubgroupContainer

imagine-hussain (approved):

Peedee2002 (approved):
Noice

PR #346, opened 2022-04-12
sozz debug warning for "isAccurate"

hdphuong:
- when a course is added to the planner, its data is retrieved through getCourse endpoint which doesn't have "isAccurate" so it's default to false and display warning 
==>> set default value to True


Peedee2002:
is there a reason we cant fetch this information if the FE needs it? We can add an endpoint isUnlocked that gets if a single course is unlocked. That might be useful for the FE. Also, it wont lead to code duplication because getAllUnlocked still is the same - ut willl just call isUnlocked.

hdphuong:
yeah we can, but only for isAccurate tho. 
cos for isUnlocked, there will be a tons of warnings in the unplanned column at set up since user hasn’t planned any course. 

Will fix this tmr :D

Peedee2002:
> yeah we can, but only for isAccurate tho. cos for isUnlocked, there will be a tons of warnings in the unplanned column at set up since user hasn’t planned any course.
> 
> Will fix this tmr :D

ah - so this is populated later. If thats the case then im happy.

Peedee2002 (approved):

gtangelo (approved):

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -135,7 +135,7 @@ export default function CourseDescription({ structure }) {
         isUnlocked: true,
         warnings: [],
         handbook_note: "",
-        isAccurate: true,
```
Peedee2002: Hmm - neither of these is correct tbh. The user can still put the order incorrectly (e.g. the user has done 1511, but only after 1521. This should return false, but instead this will be true - because the order is not correctly considered using this method.)

Peedee2002 (changes requested):
Hmm... we probably need to chat about this more - we can't determine if the course is accurate or not in the course selector. Is this nessesary? If it's a wierd issue with our architecture, we probably should apply a bandaid fix and come back to it - but we should think about if there is a better way here. @gtangelo any ideas?

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
hdphuong: hmm you should look again at the files changed, pretty sure `isAccurate = course.is_accurate` now. Not sure why it's still displayed as `isAccurate: true` here.

Peedee2002: Yup fair, nvm I misunderstood

Peedee2002 (approved):
Nice!

PR #348, opened 2022-04-12
Circles 302 courses unlocked refactor direct indirect

imagine-hussain:
Refactored `coursesUnlockedWhenTaken` to differentiate between courses unlocked directly through doing the given course and, those unlocked through other requirements such as UOC.
See example on frontend: The user has taken two course and after taking `comp2521`, they will unlock `TABL2170` due to its `18UOC` requirement.
image

Also, minor style changes as per pylint.


backend/server/routers/courses.py
```diff
@@ -280,7 +290,9 @@ def unselectCourse(userData: UserData, lockedCourses: list, unselectedCourse: st
                     "content": {
                         "application/json": {
                             "example": {
-                                "courses_unlocked_when_taken": ["COMP2511, COMP3311"]
+                                "courses_unlocked_when_taken": ["COMP2511, COMP3311"],
```
joooooooooooooooooooooooooooooooooooosh: is COMP3311 supposed to fall under one of the direct/indirect categories?

backend/server/routers/courses.py
```diff
@@ -297,9 +309,17 @@ def coursesUnlockedWhenTaken(userData: UserData, courseToBeTaken: str):
...
     courses_now_unlocked = unlocked_set(getAllUnlocked(user)['courses_state'])
+    new_courses = courses_now_unlocked - courses_initially_unlocked
+
+    ## Differentiate direct and indirect unlocks
+    path_to = set(getCourse(courseToBeTaken)['path_to'])
+    direct_unlock = new_courses.intersection(path_to)
+    indirect_unlock = new_courses - direct_unlock
 
     return {
         'courses_unlocked_when_taken' : sorted(list(courses_now_unlocked - courses_initially_unlocked)),
```
joooooooooooooooooooooooooooooooooooosh: 1.`courses_unlocked_when_taken` can just be `new_courses`
2. what's the reason for returning `courses_unlocked_when_taken` when it appears to be just a duplicate of `direct_unlock + indirect_unlock`?

joooooooooooooooooooooooooooooooooooosh (changes requested):

backend/server/routers/courses.py
imagine-hussain: Had it initially to maintain compatability with FE but you're right, not needed anymore.

backend/server/routers/courses.py
imagine-hussain: Fixed, added to direct

backend/server/routers/courses.py
```diff
@@ -19,6 +23,7 @@
 
 @router.get("/")
 def apiIndex():
+    """ returns the index of the courses API """
```
joooooooooooooooooooooooooooooooooooosh: tiny nitpick but docstrings should probably be capitalised e.g. Returns the index of the courses API

backend/server/routers/courses.py
```diff
@@ -93,7 +98,8 @@ def fixUserData(userData: dict):
     },
 )
 def getCourse(courseCode: str):
-    """get info about a course given courseCode
+    """
+    get info about a course given courseCode
```
joooooooooooooooooooooooooooooooooooosh: same here

backend/server/routers/courses.py
joooooooooooooooooooooooooooooooooooosh: this line is now irrelevant in the documentation since that's not included in the response

imagine-hussain: Updated

backend/server/routers/courses.py
imagine-hussain: done

backend/server/routers/courses.py
imagine-hussain: fixed

joooooooooooooooooooooooooooooooooooosh (approved):

PR #355, opened 2022-04-16
luke/circles-288 -- Integrate Enrolment Numbers

banaboi:
Opened this up now to get some initial thoughts on the changes. Added a progress bar under course details that shows the current capacity and enrolments as a percentage. The current data gets pulled from 2022-T2, maybe we want to have a drop down to select which term the user is concerned with? Also made a slight change to how pageLoaded is set to accommodate the other api calls that have been added after getCourse. 

Also ive noticed https://timetable.csesoc.unsw.edu.au/api/terms/2022-T2/courses/.. is down for some reason? Currently im just running the timetable api on localhost:3001 for now

image


frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -26,6 +26,43 @@ const CourseAttribute = ({ title, content }) => {
...
+  : bgColor = "green";
+
+  const Parentdiv = {
+    height: height,
+    width: '100%',
+    backgroundColor: 'whitesmoke',
+    borderRadius: 40,
+  }
+
+  const Childdiv = {
```
Peedee2002: all of these should be in the CSS file.

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -98,20 +136,44 @@ export default function CourseDescription({ structure }) {
...
+      getCourseCapacityById(id);
     }
+    setPageLoaded(true);
   }, [id]);
 
+  const getCapacityAndEnrolment = (data) => {
+    let enrolmentCapacityData = {};
+    for (let i = 0; i < data.classes.length; i++) {
+      if (data.classes[i].activity === "Lecture") {
+        enrolmentCapacityData = data.classes[i].courseEnrolment;
```
Peedee2002: this is an ok workaround - but you should notice this will fail if a course doesnt run lectures (eg SENG2991 https://timetable.csesoc.app/api/terms/2022-T1/courses/SENG2991). Also, this assumes that there is only 1 lecture. (eg MATH1141 https://timetable.csesoc.app/api/terms/2022-T1/courses/MATH1141). You should come up with a way to deal with both those cases.

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -280,6 +343,17 @@ export default function CourseDescription({ structure }) {
                   </Tag>
                 );
               })}
+            <Title level={3} className="text cs-final-attr">
+              Capacity
+            </Title>
+            {courseCapacity.enrolments ? (
+                <ProgressBar
+                  progress={Math.round((courseCapacity.enrolments/courseCapacity.capacity) * 100)}
```
Peedee2002: this is nice - but we should also have text to tell them the exact capacity as well.

imagine-hussain:
Is it supposed to have nothing to show at the moment?
Is this because registrations are not open? If so, during what times will this feature typically show a number.
image

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -280,6 +343,17 @@ export default function CourseDescription({ structure }) {
...
+                  height = {30}
+                />
+            ) : (
+              <Text>Nothing to show</Text>
```
imagine-hussain: Is it supposed to have nothing to show at the moment?
Is this because registrations are not open? If so, during what times will this feature typically show a number.
image
For context, that's COMP2521 but this result appears for all courses.

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -280,6 +343,17 @@ export default function CourseDescription({ structure }) {
                   </Tag>
                 );
               })}
+            <Title level={3} className="text cs-final-attr">
+              Capacity
```
imagine-hussain: This should probably also include what term the capacity is being shown for since it's not immediately clear whether during something like Week 10 of a term if we're showing the capacity for the current term or the next.

imagine-hussain (changes requested):

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -26,6 +27,22 @@ const CourseAttribute = ({ title, content }) => {
   );
 };
 
+const ProgressBar = ({progress,height}) => {
```
gtangelo: This component should be seperated to another file under the `components` folder with its own js and css/less files included.

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -15,6 +15,7 @@ import { setCourse } from "../../../actions/coursesActions";
 import { BsPlusLg } from "react-icons/bs";
 
 const { Title, Text } = Typography;
+const term = "2022-T2";
```
gtangelo: I think it should probably be better if this variable is defined as a `REACT_APP_` env var or create a file in the root of the `src` folder (something like `config.js`) that exports this variable. Would be easier to know how to change the `term` value if another dev is not familiar on how to change the term later on. What do you think?

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -26,6 +27,22 @@ const CourseAttribute = ({ title, content }) => {
...
 
+const ProgressBar = ({progress,height}) => {
+  let bgColor = "";
+  progress >= 75 ? bgColor = "red"
+  : progress >= 45 ? bgColor = "orange"
+  : progress >= 25 ? bgColor = "yellow"
+  : bgColor = "green";
+
+  return (
+  <div className="progressBarWrapper" height={height}>
```
gtangelo: Nitpick: For the `height`, `width` and `backgroundColor` you could just use the `style` attributes and just create a style object to pass. i.e. `style={{height: height}}`. Just a personal preference tho 😄

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -98,20 +116,47 @@ export default function CourseDescription({ structure }) {
...
+        setCoursesPathTo({
+          direct_unlock: data.direct_unlock,
+          indirect_unlock: data.indirect_unlock,
+        });
+    };
+
+    const getCourseCapacityById = async (id) => {
+      const [data, err] = await axiosRequest(
+        "get",
+        `https://timetable.csesoc.app/api/terms/${term}/courses/${id}`
```
gtangelo: Think we should also make the base url (`timetable.csesoc.app`) configurable as well similar to `term` by making it as a `REACT_APP_` env var or an exported variable as discussed with `term` above.

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -98,20 +116,47 @@ export default function CourseDescription({ structure }) {
...
 
+  const getCapacityAndEnrolment = (data) => {
+    let enrolmentCapacityData = {
+      enrolments: 0,
+      capacity: 0,
+    };
+    for (let i = 0; i < data.classes.length; i++) {
+      if (data.classes[i].activity === "Lecture" || data.classes[i].activity === "Seminar" || data.classes[i].activity === "Thesis Research") {
+        enrolmentCapacityData.enrolments += data.classes[i].courseEnrolment.enrolments;
+        enrolmentCapacityData.capacity += data.classes[i].courseEnrolment.enrolments;
```
gtangelo: Correct me if I'm wrong but wouldn't `enrolmentCapacityData.enrolments` and `enrolmentCapacityData.capacity` end up as the same value after the loop? Since both are adding `data.classes[i].courseEnrolment.enrolments`.

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -200,15 +245,19 @@ export default function CourseDescription({ structure }) {
                 </Dropdown.Button>
               )}
             </div>
-            {course.is_legacy && <Text strong> NOTE: this course is discontinued - if an equivalent course exists for it, please pick that instead </Text>}
+            {course.is_legacy && (
+              <Text strong>
+                {" "}
```
gtangelo: Could you remove `{" "}` if the space is not needed? Would be better style wise!

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -280,6 +326,17 @@ export default function CourseDescription({ structure }) {
                   </Tag>
                 );
               })}
+            <Title level={3} className="text cs-final-attr">
```
gtangelo: Would it be better to include the title for Capacity in the conditional rendering below? So if there is no `courseCapacity.enrolments`, then we don't show this at all (instead of showing the text `Nothing to show`)? Let me know your thoughts.

frontend/src/pages/CourseSelector/courseDescription/courseDescription.less
```diff
@@ -90,6 +90,20 @@
   gap: 20px;
 }
 
+.progressBarWrapper {
+  width: '100%';
+  background-color: 'whitesmoke';
```
gtangelo: Can you specify the hex code instead? Also, for `border-radius` could you specify the unit of measurement aswell? i.e. `px`

frontend/src/pages/CourseSelector/courseDescription/courseDescription.less
```diff
@@ -90,6 +90,20 @@
...
+  border-radius: 40;
+}
 
+.progressBar {
+  height: '100%';
+  border-radius:40;
```
gtangelo: Could you specify the units being used i.e. `px`?

frontend/src/pages/CourseSelector/courseDescription/courseDescription.less
```diff
@@ -90,6 +90,20 @@
...
+  text-align: 'right'
+}
 
-
+.progressText {
+  padding: 10;
```
gtangelo: For `padding`, could you specify the units being used i.e. `px`? And also, for `color`, replace the value to its hex code i.e. `#000`

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -26,6 +27,22 @@ const CourseAttribute = ({ title, content }) => {
   );
 };
 
+const ProgressBar = ({progress,height}) => {
+  let bgColor = "";
+  progress >= 75 ? bgColor = "red"
```
gtangelo: These ternary statement is a bit confusing to read. Could you be able to simplify the logic? haha.

Something like this would work
```
const bgColor = "green"
// would be a cascading effect to assign the bgColor
progress >= 25 && (bgColor = "yellow")
progress >= 45 && (bgColor = "orange")
progress >= 75 && (bgColor = "red")
```

Also, if you have time, could you find similar hex codes to the yellow, orange and red color? I feel like the default css color options (i.e. red, yellow) have a higher constrast to what our main color theme is (generally a lighter and softer color feel)

gtangelo (changes requested):
Sorry if this review is a bit harsh haha. Just wanted to enforce good code standards, styling and conventions in the FE repo. Otherwise, everything is looking good! Pretty impressive to code up something like this coming from the BE team 🤩.

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
banaboi: Yeah that's a typo. ignore this for now. This is a rough amendment. Need to think about the logic a bit more

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
banaboi: Yeah this is better. Have put this in

frontend/src/pages/CourseSelector/CourseDescription/CourseDescription.jsx
```diff
@@ -50,11 +52,43 @@ const CourseDescription = () => {
...
+        if (
+          data.classes[i].activity === "Lecture"
+          || data.classes[i].activity === "Seminar"
+          || data.classes[i].activity === "Thesis Research"
+        ) {
+          enrolmentCapacityData.enrolments
+            += data.classes[i].courseEnrolment.enrolments;
+          enrolmentCapacityData.capacity
+            += data.classes[i].courseEnrolment.capacity;
+        }
```
banaboi: Ignore this for now still brainstorming some ideas

frontend/src/components/ProgressBar/ProgressBar.jsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import "./progressBar.less";
```
gtangelo: could you rename this less file to `index.less` instead?

frontend/src/pages/CourseSelector/CourseDescription/CourseDescription.jsx
```diff
@@ -192,6 +226,19 @@ const CourseDescription = () => {
               })
               : "None"}
           </div>
+          {courseCapacity !== {} && (
```
gtangelo: I think if `courseCapacity` is `{}`, I think it may still trigger this conditional rendering. JS is weird in that `{} === {}` is still false (I believe it has to do with the object memory addresses). You could try to do something like `Object.keys(courseCapacity).length > 0` instead?

frontend/src/components/ProgressBar/progressBar.less
```diff
@@ -0,0 +1,17 @@
+.progressBarWrapper {
+  width: '100%';
```
gtangelo, line 2: css values do not have '' in the css value. i.e. should be `width: 100%` rather than `width: '100%'`

gtangelo (changes requested):
Lgtm with the code!

Just need to fix the css! Also, not entirely sure if enrolment should be in another column. Think its better to be placed righ under `Offering Terms`?

image

frontend/src/components/ProgressBar/progressBar.less
banaboi: Yeah forgot to change it when moving it over from the react code. Fixed now

frontend/src/pages/CourseSelector/CourseDescription/CourseDescription.jsx
banaboi: I think courseCapacity.enrolements && ... would have also worked but that's interesting. Went with your suggestion

frontend/src/components/ProgressBar/ProgressBar.jsx
banaboi: Done

banaboi: And in regards to adding it to the other column, didn't intend for that. Will investigate different screen sizes

banaboi: fixed

frontend/src/components/ProgressBar/index.less
```diff
@@ -0,0 +1,16 @@
...
+.progressBar {
+  height: 100%;
+  border-radius: 40px;
+}
+
+.progressText {
+  padding: 10px;
+  color: #000;
+  font-weight: 1000px;
+}
```
imagine-hussain: can we get a newline here

frontend/src/components/ProgressBar/ProgressBar.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+  return (
+    <div className="progressBarWrapper">
+      <div
+        className="progressBar"
+        style={progressBarStyle}
+      >
+        <span
+          className="progressText"
+        >{`${progress}%`}
```
imagine-hussain: If we're going with the `3 colour` model then, I think it's probably more clear if we move the `progressBarStyle` to a `.less` file and have the progress variable just denote a `className`.

It may also be worth adding some sort of gradient in the future. I'll approve w/out the gradient on this PR but, may be nice to add to the backlog.

frontend/src/components/ProgressBar/index.jsx
```diff
@@ -0,0 +1,3 @@
+import ProgressBar from "./ProgressBar";
```
imagine-hussain, line 1: Is this something that we have to build ourselves? It may be easier to borrow the progress bars from `antd` just for future maintenance and aesthetic consistency.

https://ant.design/components/progress/

They've got a few selections in there that just require passing in a percentage. lmk your thoughts on those and if they satisfy our needs

imagine-hussain (changes requested):
The functionality here looks really nicely done.
I've just got an inquiry here about making our own progress bar vs using antd and a minor comment about moving styling to style sheets.

frontend/src/components/ProgressBar/index.less
banaboi: added

frontend/src/components/ProgressBar/index.jsx
banaboi: Looks good, added

frontend/src/components/ProgressBar/ProgressBar.jsx
banaboi: This has been changed based on the antd component. Let me know your thourghts

imagine-hussain (approved):
Looks great!

Peedee2002:
@gtangelo mind reviewing?

Peedee2002 (approved):
noice

frontend/src/config.js
```diff
@@ -0,0 +1,3 @@
+const REACT_APP_TERM = "2022-T2";
```
gtangelo: Maybe add these to `constants.js` instead as direct exports?

gtangelo (approved):
lgtm!

frontend/src/config.js
Peedee2002: Wait @banaboi I just noticed this - you should use `getlatestterm` and the `date` function to get the term and year, respectively

PR #363, opened 2022-04-18
Cf 185 collapse direct indirect unlocks

imagine-hussain:
Made elements of courseDescription like the requirements, courses done to unlock this course, and the direct / indirect unlocks collapsible.
Elements can be collapsed by clicking on their title. Hovering over the title will change the colour of the background and indicate it can be clicked on it.
FEEDBACK needed for the background colour on hover. I think it *may* be too dark at the moment but am not good enough at UI to choose a colour that would look nice.

https://user-images.githubusercontent.com/93496985/163794326-ac739cbb-7b2c-4a8a-8e10-a4542b4ad3fe.mp4

More questions to consider:
- Should overview section also should be collapsible?
- Does the filled in button for dropdown look fine or, should we be using something like the attached image?
image


frontend/src/pages/CourseSelector/courseDescription/Collapsible.jsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+
+const CollapsibleButton = ({ isCollapsed, setIsCollapsed }) => {
+  const rotation = isCollapsed ? 0 : -90;
+  
+  const classes = (isCollapsed) ? "collapsible-button collapsible-button-collapsed" : "collapsible-button";
+  
+  return (
+    <div className={classes}>
```
gtangelo: Since `classes` is used once, might as well put the ternary statement inside the `className` itself?

frontend/src/pages/CourseSelector/courseDescription/Collapsible.jsx
```diff
@@ -0,0 +1,51 @@
...
+      <svg
```
gtangelo: For the icon, would it better to use [antd's native icons](https://ant.design/components/icon/) instead? Since it might closely resememble the antd theme. Just a suggestion though!

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -112,6 +114,21 @@ export default function CourseDescription({ structure }) {
     }
   }, [id]);
 
+  const [collapseRequirements, setCollapseRequirements] = useState(false);
+  const [collapseOverview, setCollapseOverview] = useState(false);
+  const [collapseDone, setCollapseDone] = useState(false);
+  const [collapseDirectUnlock, setCollapseDirectUnlock] = useState(false);
+  const [collapseIndirectUnlock, setCollapseIndirectUnlock] = useState(true);
+
+  const toggleCollapse = (state, setState) => {
```
gtangelo: Is this `toggleCollapse` being used anywhere in this file?

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -211,50 +228,66 @@ export default function CourseDescription({ structure }) {
                 />
               </Text>
             </Space>
-            <Title level={3} className="text">
-              Requirements
-            </Title>
+              <CollapsibleHeader
```
gtangelo: I'm not entirely sure if we should use our own component for collapsible. The antd theme provides a `collapse` [component](https://ant.design/components/collapse/#header) which might be an alternative to this. May be simpler to work with and abit cleaner as it seems like we don't have to worry bout managing states and reinventing the wheel.

gtangelo (changes requested):
Nice work! Though I think using antd [collapse components](https://ant.design/components/collapse/#header) might prove advantageous that creating our own component? Let me know what you think!

Not entirely sure how the collapsible looks atm. It looks a bit squashed imo, perhaps thats because of the added underline to the header now? Adding some spacing between each section might make it seem more spacious? It makes sense to make direct and indirect courses, and courses to unlock this course collapsible. But not entirely sure if it fits with `Requirements`. I feel like the point of a collapse action is to hide unrelevant information or if the information is alot for the user to take it. Maybe `Requirements` does not fit that criteria?

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
imagine-hussain: Was being used originally but the responsibility was moved into the component when cleaning up code.
Resolved.

frontend/src/pages/CourseSelector/courseDescription/Collapsible.jsx
imagine-hussain: Resolved.

joooooooooooooooooooooooooooooooooooosh:
> Does the filled in button for dropdown look fine or, should we be using something like the attached image?

I think using something like the image you've attached would be a better fit, especially as it matches better with the drop down arrows used in the left sidebar

frontend/src/pages/CourseSelector/courseDescription/Collapsible.jsx
imagine-hussain: I'll probably end up going with one of the following then. 
I'm considering the `RightOutlined` due to @joooooooooooooooooooooooooooooooooooosh 's suggestion. lmk if you have a strong preference towards one or the other.
image

joooooooooooooooooooooooooooooooooooosh: Using RightOutlined seems a bit weird when the sidebar uses up/down arrows for collapsible containers

imagine-hussain:
@gtangelo 
>  think using antd [collapse components](https://ant.design/components/collapse/#header) might prove advantageous that creating our own component? Let me know what you think!

I think the issue here is that we pass the collapsible `heading` as a prop so we lose much of the control over styling.
Also, the components have outlined and backgrounds. The main conflict here would be losing the `Title` styling that we've created. I think the default component is nice but doesn't work as well for when what we're trying to collapse is such a large segment in the page rather than just inline content.

> Not entirely sure how the collapsible looks atm. It looks a bit squashed imo, perhaps thats because of the added underline to the header now? Adding some spacing between each section might make it seem more spacious?

Definitely can do this

> It makes sense to make direct and indirect courses, and courses to unlock this course collapsible. But not entirely sure if it fits with `Requirements`. I feel like the point of a collapse action is to hide unrelevant information or if the information is alot for the user to take it. Maybe `Requirements` does not fit that criteria?

My thought process here was that we'd also make `Overview` collapsible because for some courses the overview text is pretty huge, bigger than the unlocks. Then if that's the case, it would look strange if every element is collapsible except for overview.

For context: the `antd` collapsibles:
image

image

gtangelo:
@imagine-hussain 

> I think the issue here is that we pass the collapsible heading as a prop so we lose much of the control over styling.
Also, the components have outlined and backgrounds. The main conflict here would be losing the Title styling that we've created. I think the default component is nice but doesn't work as well for when what we're trying to collapse is such a large segment in the page rather than just inline content.

Ah I see. I was thinking of using the `ghost` attribute + custom styling to the `Panel` component for the header to achieve a similar look.
image
Either way, I'm happy for you implementation to be used if it is easier and achives what you need! 😄 Though a note of refactoring, for the `CollapsibleHeader` component, rather than passing in `isCollapsed` and `setIsCollapsed`, would you be able to include the state management stuff within the component? I know you're need the state for managing displaying the content under the header, but this could be passed down via the `children` prop into the `CollapsibleHeader`? Right now, I feel that there's a bit too much repetition in displaying the collapsible headers with managing the all of the states in the parent component which the logic could be abstracted out into the `CollapsibleHeader` component?

> My thought process here was that we'd also make Overview collapsible because for some courses the overview text is pretty huge, bigger than the unlocks. Then if that's the case, it would look strange if every element is collapsible except for overview.

Oh right. Yep having `Overview` should be collapsible would be good if this is the case.

frontend/src/pages/CourseSelector/courseDescription/Collapsible.jsx
gtangelo: Happy with the `RightOutlined` icon :D

imagine-hussain:
@gtangelo 
> Could be passed down via the children prop into the CollapsibleHeader?
Was trying to figure out how to do this but TIL. Has been converted to use children and all state management is now self contained.

- Overview is now also collapsible.
- Changed colour of highlight on hover to an opacity closer to 50% (lmk if this needs changing, very open to feedback on what the styling should look like on hover)
- Adding spacing between collapsible elements.
- Using `LeftOutline` element from antd.

@joooooooooooooooooooooooooooooooooooosh 
> Using RightOutlined seems a bit weird when the sidebar uses up/down arrows for collapsible container
I tried the up / down  but I don't think that animation looks a little over the top for a component of this size. 

Pictured:
1. Default View on a course Description
2. Some collapsed elements + hover on overview
3. Current build

image 1

image 2

image 3

frontend/package.json
```diff
@@ -15,6 +15,7 @@
     "craco-less": "1.20.0",
     "framer-motion": "6.3.3",
     "moment": "^2.29.3",
+    "package.json": "^2.0.1",
```
gtangelo: Oh any reason why this was included (and is this needed)? lol If not, could we remove it.

frontend/src/pages/CourseSelector/main.less
```diff
@@ -63,4 +63,50 @@
   }
 }
 
+.collapsible-header {
```
gtangelo: Could these css classes be included in a seperate less file? Maybe place these stylings into a file like `Collapsible.less` for now.

gtangelo (approved):
I like this approach! Nice work!

Some minor things:
For the hover effect, I feel like making the background change color might not look that great? Instead, maybe just change the text color to our purple hex code when hovered instead of the background? Reason for this is because you already changed the cursor on hover (which already tells the user that this can be clicked on) + with the added text color change on hover, I feel that it's conveys enough information to the user that this item is collapsible.

This is pretty subjective though so I'm happy idm what approach is taken either way. Up to your discretion to which you would like :D.

BUT, if you do decide to keep the background, some improvements:
- For the background color, you could try a lighter variant. Try something like `rgb(195, 195, 195)` as the bg color. Wouldn't have a strong color effect to what is present right now.
- Add `padding-right` for `collapsible-header` class? Right now the arrow is right to the edge of the collapsible container and is more apparent with the bg color on hover. 
- Just a query, any reason why we are positioning the arrow to the right of the container? We could try positioning it left just before the title, though not entirely sure if this will look great.

Just to restate that this structure and your code is great stuff! Just some nitpicks here and there so this PR may have seem abit harsh haha

frontend/package.json
imagine-hussain: I didn't change anything here so it may have come by merging in main or another branch.

imagine-hussain: I'll delete and re-merge dev into this branch.

Peedee2002: @imagine-hussain this could have happened if you ran `npm install` wierdly (eg running `npm install package.json` - in npm, you just need to run `npm install` to install your packages. If you also want to respect what is in `package-lock.json`, you can run `npm ci`), which is what i suspect - you can probably remove this from your package.json and run npm install again to fix your package-lock.json.

imagine-hussain:
> For the hover effect, I feel like making the background change colour might not look that great? Instead, maybe just change the text colour to our purple hex code when hovered instead of the background? Reason for this is because you already changed the cursor on hover (which already tells the user that this can be clicked on) + with the added text colour change on hover, I feel that it's conveys enough information to the user that this item is collapsible.

Went with the purple hover, looks much better.

> Just a query, any reason why we are positioning the arrow to the right of the container? We could try positioning it left just before the title, though not entirely sure if this will look great.

Was to do with alignment of the collapsible content below the header. Moved it to left now and fixed  the alignment issues.

image

image

PR #397, opened 2022-04-26
CF-190 Refactor to used redux toolkit

gtangelo:
Previously, we were using `createStore` for redux. However, this approach has been made deprecated and the recommended approach now is to use the redux toolkit.

Therefore, this PR contains refactoring on the redux structure on the FE using the redux toolkit approach, which simplifies overall usage. Also, some logic of the reducers have been simplified/logic bugs spotted in the refactoring have been fixed.


frontend/src/reducers/plannerSlice.js
```diff
@@ -0,0 +1,292 @@
...
+          state.courses[code].handbookNote = "";
+          state.courses[code].isAccurate = true;
+        }
+      });
+      localStorage.setItem("planner", JSON.stringify(state));
+    },
+    toggleSummer: (state) => {
+      state.isSummerEnabled = !state.isSummerEnabled;
+
+      // TODO: Unsure if we should keep the course in the summer column if hidden
```
Peedee2002, line 180: what is going on here?

Peedee2002 (approved):
looking up and down this PR, i like it a lot! nice.

frontend/src/reducers/plannerSlice.js
gtangelo: So there's an edge case that I encountered while testing out the term planner. Let's say that a user has the summer term toggled on when planning their courses. The user drags their courses into the summer column. However, the user decides to not show the summer column anymore and toggles the summer term to not be shown. There's two intents to consider:
- If the user simply does not want to do a summer term in the degree, should all the planned courses in the summer term be moved back to unplanned? The current implementation does not does this and the summer courses are simply left in the summer column (which is hidden). This means that the user can't drap/plan their degree for courses in the summer column without toggling the summer term back on.
- If the user simply wants to hide the summer term, toggle off the summer column and toggle on the summer column after, should we save the summer courses?

I still need to see if there's a better way to handle this but this would be out of the scope of the ticket I feel (and just want to merge the redux changes into dev as soon lol). Personally, I think the first option would be simplier but any thoughts on this is welcomed :D

Peedee2002: hmm im not too fussed by this - i think its fine to leave the courses in the summer term. as long as we get the dragndrop to unplanned working the difference in the approaches should be minimal.

PR #404, opened 2022-04-27
TIL-negative-lookbehind

hdphuong:
- same behaviours as previous PR but better regex


Peedee2002 (approved):
this is wizardry, but thats ok lmao

PR #407, opened 2022-04-27
Gabriel/cs/cf 181 add to planner transition

gtangelo:
Changes made:
- Refactored file structure and cleaned up course description components
- Fix bugs associated with the add to planner button (CF-174 - text placement when removing course, CF-186 - flashing when course is added)
- CF-181 - Added correct transition when adding course to planner i.e.

image
image
image
image


frontend/src/pages/CourseSelector/courseDescription/AddToPlannerButton.jsx
```diff
@@ -0,0 +1,95 @@
...
+  const coursesInPlanner = useSelector((state) => state.planner.courses);
+  const course = useSelector((state) => state.courses.course);
+
+  const id = tabs[active];
+  const dispatch = useDispatch();
+  const [addedCourseInPlanner, setAddedCourseInPlanner] = useState(!!coursesInPlanner[id]);
+  const [loading, setLoading] = useState(false);
+
+  useEffect(() => {
+    if (!!coursesInPlanner[id] !== addedCourseInPlanner) {
```
Peedee2002: this is a guard, so use an early return

frontend/src/pages/CourseSelector/courseDescription/AddToPlannerButton.jsx
```diff
@@ -0,0 +1,95 @@
...
+      if (coursesInPlanner[id]) {
+        // course is being added
+        setLoading(true);
+        setTimeout(() => {
+          setLoading(false);
+          setAddedCourseInPlanner(true);
+        }, 1000);
+      } else if (!coursesInPlanner[id]) {
```
Peedee2002: these two clauses are the same in every respect except for setAddedCourseInPlanner - use a turnary statement to differentiate between the two

frontend/src/pages/CourseSelector/courseDescription/AddToPlannerButton.jsx
```diff
@@ -0,0 +1,95 @@
...
+        prereqs: course.raw_requirements,
+        isLegacy: course.is_legacy,
+        isUnlocked: true,
+        warnings: [],
+        handbookNote: course.handbook_note,
+        isAccurate: course.is_accurate,
+      },
+    };
+    dispatch(addToUnplanned(data));
+    setLoading(true);
```
Peedee2002: this block is used a lot - refactor into a function.

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
```diff
@@ -119,81 +68,18 @@ const CourseDescription = ({ structure }) => {
...
-                  overlay={() => (
-                    <PlannerDropdown
-                      courseCode={course.code}
-                      structure={structure}
-                      addToPlanner={addToPlanner}
-                    />
-                  )}
-                  loading={loading}
-                  onClick={() => {
-                    addToPlanner("Uncategorised");
```
Peedee2002, line 189: where did this logic go?

Peedee2002 (changes requested):
a few stylistic changes here, but i like the change :)

frontend/src/pages/CourseSelector/courseDescription/CourseDescription.jsx
gtangelo: Removed the dropdown for now since we aren't considering the course's type (i.e. Core Course, Gen Ed) for the term planner atm. Will revisit this down the road

gtangelo:
This may create a conflict with your changes on the deselect so feel free to merge yours in first then I can fix any potential issues that may arise! @Peedee2002

gtangelo:
Updated changes:
- reuse logic of `unschedule` reducer in `unscheduleAll` in `plannerSlice.js` rather than repeating code to unschedule all courses
- moved logic of removing several courses into a reducer function. This fixes the lag present with framer motion if the course removed removes alot of courses that require it. i.e. previously, removing 'COMP1511' causes some lag with the course selector menu dropdown when the user has already added several comp courses

Peedee2002 (approved):
looks good to me

PR #408, opened 2022-04-27
CF-195 CF-192 Update to use antd theme icons and change to lock icon on term planner

gtangelo:
Changes made:
- Update most icons to use antd theme to keep it consistent
- Restructured ContextMenu component and css files into a seperate folder
- Term planner now displays a lock icon instead of a tick
image


imagine-hussain:
For the lock, could we overlay it over a round grey-ish circle in the same manner as the ticks?
It would prevent the lock from looking disconnected from the term box but, it may look worse due to the sizing issues we'd encounter trying to get the lock without making the container too big.

Peedee2002 (approved):

imagine-hussain (approved):

gtangelo:
> lock from looking disconnected from the term box but, it may look worse due to the sizing issues we'd encounter trying to get the lock without making the container

I've updated the lock icon to have a background. Let me know what you think! @imagine-hussain 
image

Think it looks good with the background!

imagine-hussain:
> > lock from looking disconnected from the term box but, it may look worse due to the sizing issues we'd encounter trying to get the lock without making the container
> 
> I've updated the lock icon to have a background. Let me know what you think! @imagine-hussain image
> 
> Think it looks good with the background!

I quite like the way this turned out.
Looks good to merge @gtangelo .

PR #415, opened 2022-04-29
Leo/tp/CF-175-help-popup-bugs

leonardo-fan:
- help gif hover shadow transition
- tippy font size fix
- made option header a fixed header so when scrolling it doesn't disappear 
image


Peedee2002:
pics or it didnt happen 😆

frontend/src/pages/TermPlanner/optionsHeader/main.jsx
```diff
@@ -68,55 +68,61 @@ const OptionsHeader = ({
...
-          </button>
-        </Tooltip>
+        </div>
       </div>
 
       <Tippy
         content={<HelpMenu />}
         moveTransition="transform 0.2s ease-out"
-        interactive
+        interactive="true"
```
Peedee2002: lol does tippy require you to pass true as a string? thats so troll

frontend/src/pages/TermPlanner/main.less
```diff
@@ -1,23 +1,23 @@
...
+    height: calc(100vh - var(--navbar-height) - var(--option-header-height));
+    padding-top: 0.5em;
     overflow: auto;
   }
 
   .plannerContainer {
     display: flex;
     // align-items: center;
-    margin-top: 0.5em;
+    margin: 1em 1em 1em 1em;
```
gtangelo: Just a note, for units, might be better of to go with rem instead of em? rem is sizing relative to the root while em is relative to the parent node. So if the parent size changes, this could be changed aswell which may not be intend. Personally, I always go with rem when I need it as it keep things tied relative to the root node.

Also, for margin and padding, I usually stick with using px instead so that may be a better option? :)

Though if using em is the best in this case, go for it!

frontend/src/pages/TermPlanner/main.less
```diff
@@ -1,23 +1,23 @@
 html,
 body {
   height: 100%;
+  --option-header-height: 65px;
```
gtangelo: Nice! Like the usage of a css var. Maybe define it inside `App.less` instead? Would be easier to find all css variables when we need to change it in the near future!

frontend/src/pages/TermPlanner/main.less
```diff
@@ -206,8 +208,9 @@ body {
     image-rendering: auto;
     border-radius: 12px;
     cursor: help;
+    transition: box-shadow 0.5s;
```
gtangelo: could make the speed 0.3s instead? Feels like the hover animation is a bit slow on 0.5s. Also, good to specify the transition effect for readability. So `transition: box-shadow 0.3s ease` would work!

gtangelo (approved):
General Feedback on UI:
I like how the icons/settings/options are now fixed! Really good idea to have them there always. Though not entirely sure on the background/option header design. IMO I feel like the option header + navbar makes it look like its taking too much screen space? I played around and some suggestions could include:
- making the background just white instead?
- decreasing the spacing/height in the option header?

Let me know how it goes but the easiet option I feel would just make the background white? 😄 

Just some general nitpick comments I made so nothing too major needs to change.

Thanks for fixing these issues! 🚀 

(Don't forget to attach pictures next time in your PRs 👀 )

frontend/src/pages/TermPlanner/optionsHeader/main.jsx
leonardo-fan: oops don't think i need that there ahaha

frontend/src/pages/TermPlanner/main.less
leonardo-fan: ohh yup makes sense!

frontend/src/pages/TermPlanner/main.less
leonardo-fan: oo sounds good :)

leonardo-fan:
> General Feedback on UI: I like how the icons/settings/options are now fixed! Really good idea to have them there always. Though not entirely sure on the background/option header design. IMO I feel like the option header + navbar makes it look like its taking too much screen space? I played around and some suggestions could include:
> 
> * making the background just white instead?
> * decreasing the spacing/height in the option header?
> 
> Let me know how it goes but the easiet option I feel would just make the background white? 😄
> 
> Just some general nitpick comments I made so nothing too major needs to change.
> 
> Thanks for fixing these issues! 🚀
> 
> (Don't forget to attach pictures next time in your PRs 👀 )

Thanks for feedback y'alll, won't forget to add a picture when I make a pr next time oopsies. 

Regarding making the header take up less screen space, what do you think about giving a bottom border just to signify that the header part isn't scrollable/will cut off content if scrolled @gtangelo? (attached an example picture in the pr)

gtangelo:
@leonardo-fan It actually looks good! Yep I'm happy with that approach 😁. Feel free to merge into dev when your ready :D

PR #428, opened 2022-05-03
fixed the warning for isoffered through some restructuring

Peedee2002:
changed supression to simply set a flag for the frontend to deal with - this is because we want the frontend to be able to show its own warnings. This does cause extra complexity though - let me know what you all think about that.
also, i fixed some dumbass bugs i found while testing supression - i had an "idk how react works" moment for a little bit there.
warnings off (this is the behaviour a user probably expects):
image
warnings off but past the current term:
image
warnings normal:
image

also, i fixed a bug where the update for terms wouldnt occur properly on toggle:
image
then we have an invalid supressed:
image
after it is no longer supressed:
image


frontend/src/reducers/plannerSlice.js
```diff
@@ -85,13 +85,14 @@ const plannerSlice = createSlice({
...
           const {
-            is_accurate: isAccurate, unlocked, warnings, handbook_note: handbookNote,
+            is_accurate: isAccurate, unlocked, warnings, handbook_note: handbookNote, supressed,
           } = action.payload[course];
 
           state.courses[course].isAccurate = isAccurate;
           state.courses[course].isUnlocked = unlocked;
           state.courses[course].warnings = warnings;
           state.courses[course].handbookNote = handbookNote;
+          state.courses[course].supressed = supressed;
```
gtangelo, line 95: Adding `supressed` in the courses for the planner state shouldn't need a refresh on the data structure right (i.e. it would still be compatible with courses that does not have the `supressed` field)?

gtangelo (approved):

Joelrwest (approved):
Looks gooooooood to me

hdphuong (approved):
Just 2 questions:
Line 64 in `DraggableCourse.jsx` don't we need to suppress that one alert as well? 
And, wdym by "warnings off but past the current term"?

Other than that, lgtm!

Peedee2002:
> Just 2 questions: Line 64 in `DraggableCourse.jsx` don't we need to suppress that one alert as well? And, wdym by "warnings off but past the current term"?
> 
> Other than that, lgtm!

for the alert stuff, i mean that the supression is only supposed to work for courses that are before the current term - but the current term was being calculated incorrectly. 

For like 64, i think i do supress it on line 35.

frontend/src/reducers/plannerSlice.js
Peedee2002: hmm - i dont think it matters because we immediately call `validateTermPlanner` which will fill in this data on the fly. So no need to increment,

PR #452, opened 2022-05-09
fix(docs): update README screenshot link

lhjt:
This PR fixes the broken image reference in the primary README file.


gtangelo (approved):
Oh yea oops my bad. Didn't realise we reference an image in the readme.

PR #453, opened 2022-05-09
Leo/tp/cf 188 make unplanned column fixed

leonardo-fan:
- unplanned column now always visible (absolute) despite scrolling the main container
- overflow on y axis
- support for different screen sizes
- support for summer term and normal term sizes
- stylistic changes

https://user-images.githubusercontent.com/90736577/172995904-d27524ed-f9fd-4bba-bed7-a5eae01373a4.mov


frontend/src/pages/TermPlanner/TermPlanner.jsx
```diff
@@ -170,7 +170,7 @@ const TermPlanner = () => {
               <div className="gridItem">Term 1</div>
               <div className="gridItem">Term 2</div>
               <div className="gridItem">Term 3</div>
-              <div className="gridItem">Unplanned</div>
+              <div className="gridItem" />
```
gtangelo: Maybe add a comment saying that this empty div is for adding the spacing for the fix unplanned column?

gtangelo (changes requested):
Didn't realise it would need js calc for make it fixed haha. Yeet everything looks good to me!

There's one bug I found:
- There's a bug where if there are not any courses in the unplanned column, the width shrinks:
image

leonardo-fan:
Nice, sure will fix those things now!

Peedee2002 (approved):
lgtm

leonardo-fan:
refer to new description and demo :)

frontend/src/pages/TermPlanner/UnplannedColumn/index.less
```diff
@@ -1,18 +1,75 @@
...
+  top: calc(
+    var(--navbar-height) + 
+    var(--option-header-height) + 
+    var(--tp-main-container-padding) + 
+    var(--tp-planner-container-margin) + 
+    var(--tp-term-box-margin) +
+    var(--tp-grid-item-font-size)
+  );
+  height: calc(
+    100vh - (
```
gtangelo, line 43: oof alot of calcs haha. Not visually readable but I think this is the simpliest approach to do the unplanned column. Maybe could add a brief comment explaining why it's using calc?

leonardo-fan:
ahh, just found a bit of a weird thing because of the absolute positioning when the screen width gets too small, not too sure how to fix it 🆘 

https://user-images.githubusercontent.com/90736577/172997727-652c26a7-544d-4b9c-8d5a-8e6aee704a3a.mov

frontend/src/pages/TermPlanner/UnplannedColumn/index.less
```diff
@@ -1,18 +1,75 @@
...
+  align-items: center;
+}
+
+.summerUnplannedContainer {
+  grid-column-start: 6;
+}
+
+.unplannedTitle {
+  position: absolute;
+  font-weight: bold;
```
gtangelo: UI change, change it to `500` or `600` rather than `bold`? Seems a bit bold imo

frontend/src/pages/TermPlanner/UnplannedColumn/index.less
```diff
@@ -1,18 +1,75 @@
...
+      3 * var(--tp-planner-container-margin) + 
+      2 * var(--tp-term-box-margin) + 
+      var(--tp-grid-item-font-size)
+    )
+  );
+  min-width: 20em;
+  max-width: 20em;
   margin: 1em;
-  border: 1px solid #D9D9D9;
+  border: 2px solid #9254de;
```
gtangelo: I like the border-color, but imo the px should be between 1.5px to 2px? 2px feels a bit to thick for the border. But totally up to you on which looks best. Don't mind either way :D

gtangelo (approved):
Very nice! lgtm! This looks simplier compared to the ref approach but the css calc could be confusing (not a bad thing, I don't disagree with the approach but I guess it's a con of using css instead haha). Still pretty nice to see the unplanned column tho!!

I'm happy for this to be merged now 🥳. But some minor things/bugs I noticed (not a deal breaker and more of edge cases).

@leonardo-fan if you just want to merged this in and not want to fix the bugs below, pls create tickets just to take note of it in the backlog and we can address it in a later time 😅.

- Is it possible to change the background of the overflow scroll to a white color rather than a grey like color?
image
- When summer term is enabled, unplanned is not padded for higher res screens ripp (only when no courses have been planned yet). I'm happy for this to be merged without fixing it. If you don't want to fix it, create a bug ticket for it on Jira. If you do want to fix it, fix it haha. Up to u :)
image
- When dragging the course back to the unplanned column, the x-scroll appears for a second until it has been placed. Not a huge breaker. Likewise, create a bug ticket on Jira for someone else to fix it later on. But I believe it should just be a css change i.e. add overflow-x to none?
image

Nice work!! 😎 

EDIT: Just saw ur comment on the smaller screens, don't think we need to worry about smaller screens lol. Just create a bug ticket on jira in the backlog to make a mental note of it.

PR #455, opened 2022-05-10
Uzaif/cf 180

Uzaif-Sheikh:


frontend/src/components/plannerCart/PlannerCart.jsx
```diff
@@ -79,9 +79,25 @@ export const PlannerCart = () => {
       setCode("");
     }, 3500);
   };
+  const cartRef = useRef();
+
+  useEffect(() => {
+    const handlerClickOutside = (e) => {
+      console.log(e.target);
```
gtangelo: removed the console log statement

frontend/src/components/plannerCart/PlannerCart.jsx
```diff
@@ -79,9 +79,25 @@ export const PlannerCart = () => {
...
+      if (cartRef?.current?.contains(e.target) === false) {
```
gtangelo: could be simplified to `!cartRef?.current?.contains(e.target)`?

gtangelo (changes requested):
I think there's currently a bug where if you delete a course in the planner cart, then the planner cart close and doesn't actually delete the course. This is most likely the fact that the delete popup is considered outside of the planner cart element, but we would not want this behaviour at all.

Perhaps we could check if the element clicked outside of the planner cart is not also a popover?

Edit: Also merge conflicts :P

gtangelo (changes requested):
Code level looks good :), however these bugs are still present:

- When clicking the course on the planner cart which redirects to the course select page, the planner cart modal is still open. It should be closed after the redirect to the course occurs.
- When deleting a course, the cart closes and does not actually deletes the course. The planner cart should still remain open when a course is deleted from the planner cart. I think you can simply add another check to the if statement in `handlerClickOutside` function to detect if the target is in the delete popup or not.

gtangelo:
@Uzaif-Sheikh Feel free to make this PR as a draft for now if the bugs found can't be easily fixed😄

gtangelo (approved):

PR #457, opened 2022-05-11
done with cleaning - added some removal of 'equivilant' redundant information

Peedee2002:
I also added some cleaning of UNSW global as a common pattern. The equivalent stuff is what was actually screwing up some of the co-req strings after mae's changes.

I also made pytest work as expected in vscode - add:
```
    "python.testing.cwd": "${workspaceFolder}/backend"
```
to your settings.json and that will let you correctly see all the tests
image


backend/data/final_data/conditionsProcessed.json
```diff
@@ -4679,11 +4679,11 @@
     },
     "DPBS1100": {
         "original": "UNSW Global Diplomas only (7001, 7002, 7003, 7004)<br/><br/>",
-        "processed": "UNSW Global Diplomas only (7001, 7002, 7003, 7004)"
+        "processed": "(7001 || 7002 || 7003 || 7004)"
```
joooooooooooooooooooooooooooooooooooosh, line 4682: Do we need to include some kind of note that you need to be enrolled in a UNSW Global Diploma or is that handled already by something else?

joooooooooooooooooooooooooooooooooooosh: oh nvm those are programs being listed, not courses

joooooooooooooooooooooooooooooooooooosh (approved):

backend/data/processors/conditions_preprocessing.py
```diff
@@ -118,16 +119,21 @@ def preprocess_conditions():
...
         processed = re.sub(excl_string.group(1), "", processed)
 
+    if equiv_string:
+        processed = re.sub(equiv_string.group(1), "", processed)
+
     # Remove exclusion string appearing after prerequisite plaintext, typically
     # at the end of the enrolment rule
-    processed = re.sub(r"excl.*", "", processed, flags=re.IGNORECASE)
+    processed = re.sub(r"((\.|,)?\s)?excl.*", "", processed, flags=re.IGNORECASE)
+    processed = re.sub(r"((\.|,)?\s)?equivalent:.*", "", processed, flags=re.IGNORECASE)
```
hdphuong: editted: don't worry about the removed part in the original comment :D i'm dumb 

probs a good idea to use `equiv` instead of `equivalent:` since they may use `equivalence` or `equivalent courses`?

hdphuong (approved):
Noice! Great to see errors.json go down 700s lines :)

backend/data/processors/conditions_preprocessing.py
joooooooooooooooooooooooooooooooooooosh: i would be surprised if `equivalence` was used anywhere but `equivalent courses` almost certainly is used somewhere

backend/data/final_data/conditionsProcessed.json
```diff
@@ -10031,11 +10031,11 @@
     },
     "SOCW3012": {
         "original": "Prerequisite: ARTS1753 or ARTS1870 or ARTS1871 and PSYC1001, SOCW2006, SOCW2008, SOCW2009, SOCW2100, SOSS/SRAP1000, SOSS/SRAP1001,  ARTS2871 or SOSS/SRAP2002, and enrolment in a Social Work program. Equivalent: SOCW3001 and SOCW3011<br/><br/>",
-        "processed": "ARTS1753 || ARTS1870 || ARTS1871 && PSYC1001 || SOCW2006 || SOCW2008 || SOCW2009 || SOCW2100 || SOSS/SRAP1000 || SOSS/SRAP1001 || ARTS2871 || SOSS/SRAP2002 && (4033) . Equivalent: SOCW3001 && SOCW3011"
+        "processed": "ARTS1753 || ARTS1870 || ARTS1871 && PSYC1001 || SOCW2006 || SOCW2008 || SOCW2009 || SOCW2100 || SOSS/SRAP1000 || SOSS/SRAP1001 || ARTS2871 || SOSS/SRAP2002 && (4033)"
```
Joelrwest, line 10034: Unrelated to your changes, but how does the order of operations work with something like this? Seems infeasible for a computer to figure it out without manual oversight, right?

Joelrwest (approved):

backend/data/processors/conditions_preprocessing.py
Peedee2002: Yup will add some variance for that

backend/data/final_data/conditionsProcessed.json
Peedee2002: Yea, this is something the parser can arbitrarily decide - but either way a string this doomed should probably be manually fixes anyways

PR #460, opened 2022-05-11
Cf 205 navigation item highlight does not showing correctly and updated help gifs

leonardo-fan:
- change to fix bug where highlight didn't change when navigating from planner cart
- updated the images and gifs for help button
links fix

image


frontend/src/components/Header/Header.jsx
```diff
@@ -39,10 +39,11 @@ const Header = () => {
   const isSmall = useMediaQuery("(max-width: 1000px)");
   const [showDrawer, setShowDrawer] = useState(false);
   const [current, setCurrent] = useState(getCurrentPath());
+  const url = window.location.pathname;
```
gtangelo: I think you can also use the 'useLocation' hook to get the pathname aswell. Though this is just an FYI since it's basically the same functionality anyways :)

gtangelo (approved):
Noice lgtm!

PR #461, opened 2022-05-12
Uzaif/cf 207

Uzaif-Sheikh:
Locked courses are shown in redux, icon change and added tooltip


frontend/src/pages/CourseSelector/CourseSelector.jsx
```diff
@@ -1,22 +1,30 @@
...
 import CourseTabs from "./CourseTabs";
 import PageTemplate from "../../components/PageTemplate";
 import "./index.less";
 import CourseSearchBar from "./CourseSearchBar";
+import { toggleCourseLock } from "../../reducers/coursesSlice";
 
 const CourseSelector = () => {
   const [structure, setStructure] = useState({});
-  const [showLockedCourses, setShowLockedCourses] = useState(false);
+  const dispatch = useDispatch();
```
Peedee2002, line 16: oh, sorry i didnt think to make this part of localstorage. Nice.

Peedee2002 (approved):

frontend/src/pages/CourseSelector/CourseSelector.jsx
```diff
@@ -43,18 +51,23 @@ const CourseSelector = () => {
...
-            className="cs-toggle-locked"
-            onChange={() => setShowLockedCourses((prev) => !prev)}
-            checkedChildren="locked courses shown"
-            unCheckedChildren="locked courses hidden"
-          />
+          <Tooltip placement="topLeft" title={isLockedEnabled ? "Hide locked courses" : "Show locked courses"}>
+            <Switch
+              defaultChecked={isLockedEnabled}
+              className="cs-toggle-locked"
+              onChange={() => {
```
gtangelo, line 58: nitpick: just reference the function rather than creating an anon function to trigger onChange. i.e. `onChange={handleChange}` or add the function directly since its just one line like `onChange={() => dispatch(toggleCourseLock())`.

frontend/src/reducers/coursesSlice.js
```diff
@@ -1,26 +1,43 @@
...
 
-const initialState = {
+let initialState = {
   courses: {},
   course: {},
+  isLockedEnabled: false,
 };
 
+const courses = JSON.parse(localStorage.getItem("courses"));
+if (courses && courses.version === initialState.version) {
```
gtangelo, line 10: The if, else statement should be changed. I see that you haven't included the `version` field in the `initialState` object, which accessing `courses.version` would be null. Since the courses redux state is not complex (unlike the planner), we do not need to validate the structure version. Instead, something like this would be okay:
```
const courses = JSON.parse(localStorage.getItem("courses"));
if (courses) initialState = courses
```

gtangelo (commented):
I know you already merged this in to dev @Uzaif-Sheikh, but could you fix up the version field for the coursesSlice? It doesn't really make sense to include that if,else logic if the version field is not present.

frontend/src/reducers/coursesSlice.js
gtangelo: EDIT: Now that I think about it, I think I need to streamline how we are saving to local storage for every redux update. Ignore this then haha

PR #464, opened 2022-05-14
Fix Runprocessor bug and rerun processors

imagine-hussain:
- `runprocessors` was passing in `None` as the year by default which then caused `create_payload` to take the year as `None` by default rather than the `currentYear` which lead to undesired behaviour.
- Reran processors.


joooooooooooooooooooooooooooooooooooosh:
if we're going to merge it from this branch name then make sure the branch isn't auto-deleted

joooooooooooooooooooooooooooooooooooosh:
why is there a whole bunch of unrelated frontend commits? was this branch created off of another branch?

imagine-hussain:
> why is there a whole bunch of unrelated frontend commits? was this branch created off of another branch?
Not sure who created the branch but, that could be the case. Though, it wouldn't make sense as to why those are only showing up from 40 mins ago, as the branch would have been created far before then.

The more likely explanation seems would be that dev didn't merge in fully - there's conflicts in `errors.json` and `conditions.pkl`; perhaps resolving these will make it rebase properly? Not 100% sure though.

joooooooooooooooooooooooooooooooooooosh (approved):

PR #465, opened 2022-05-14
Gabriel/cf 218 redux persist

gtangelo:
Changes made:
- Removed direction interaction of local storage in reducer functions. Instead, we are now using `react-persist` to save local storage for us.
- Removed need to reset modal popup when data structure in local storage changes. Instead, it just redirects the user to the degree wizard.


Uzaif-Sheikh (approved):
lgtm.

PR #466, opened 2022-05-14
Gabriel/setup progression checker

gtangelo:
Changes made:
- Restructured the progression checker files to be more inline with Course Selector and Term Planner
- Re-enable in progress features (i.e. dark mode and progression checker) to be shown via an env variable. NOTE: This is just adding back the theme toggle and progression checker since we will most likely be working on it soon.
image


frontend/src/App.jsx
```diff
@@ -1,18 +1,30 @@
...
 import "./axios";
 import Header from "./components/Header";
 
 const App = () => {
   const [loading, setLoading] = useState(true);
+  const theme = useSelector((state) => state.theme);
+
+  useEffect(() => {
+    if (theme === "light") {
+      document.body.classList.remove("dark");
```
leonardo-fan: can use classList.replace here

frontend/src/components/Header/Header.jsx
```diff
@@ -79,13 +81,15 @@ const Header = () => {
...
               <span>Progression Checker</span>
-              <Link to="/progression-checker"></Link>
-            </Menu.Item> */}
+              <Link to="/progression-checker" />
+            </Menu.Item>
+            )}
           </Menu>
           <PlannerCart />
-          {/* <ThemeToggle /> */}
+          { inDev && <ThemeToggle /> }
```
leonardo-fan: small styling thing but inconsistent spacing in braces compared to some other files like drawercontent

frontend/src/pages/ProgressionChecker/ProgressionChecker.jsx
```diff
@@ -38,31 +39,36 @@ const degreeData = {
...
-  };
+  const [structure, setStructure] = useState({});
 
   useEffect(() => {
-    fetchDegree();
-  }, []);
+    // get structure of degree
+    const fetchStructure = async () => {
+      try {
+        const res = await axios.get(`/programs/getStructure/${programCode}/${specialisation}${minor && `/${minor}`}`);
```
leonardo-fan, line 54: could do a ternary for the url to use beforehand to avoid the nesting just to make it more readable

leonardo-fan (approved):
Nice nice the initial groundwork looks good to me apart from those small styling things I put in comments but will approve anyway. 

Know this is also just setup, so probably not too applicable for this pr, but might just give some thoughts on some things for future tickets:
- Don't see a need for the extra customisation grouping in the menu drawer
- Icon for switching to light/dark mode in drawer not centred vertically
- header toggle for light/dark spacing is a bit inconsistent to the planner cart icon and rest of nav
- constants for the light and dark colours
- bug in prog checker with list overlapping headers (in the dummy data the marine science header is overlapped)
image

gtangelo:
Ah yep nice feedback! Yep will create cards for those to be worked on in the future! Thanks for the review Leo! :D

PR #470, opened 2022-05-15
Circles 298 remove extraneous comm data

sonar3d:
added handbook notes, left some COMM manual fixes


backend/data/processors/conditions_preprocessing.py
```diff
@@ -432,6 +442,134 @@ def map_word_to_program_type(processed, regex_word, type):
...
+    # Final 2 terms note
+    processed, number_of_subs = re.subn(
+        r"It is strongly recommended that students only complete this course within their final 2 terms of study", r"", processed, flags=re.IGNORECASE
+    )
+    if number_of_subs > 0:
+        note += "Recommended to take this course within their final 2 terms of study."
+
+    # Business school students notes
+    processed, number_of_subs = re.subn(
+        r"Only available to single and double degree Business School students in Term 1. It will be offered to non-Business School students in Terms 2 and 3.<br/><br/>", r"", processed, flags=re.IGNORECASE
```
Peedee2002: there should be a way to fetch the term number using code instead of having lots of code duplication here.

backend/data/processors/manual_fixes/COMMFixes.py
```diff
@@ -27,38 +27,38 @@
...
     CONDITIONS["COMM1110"] = COMM_1110_1120_1140(CONDITIONS["COMM1110"])
-    CONDITIONS["COMM1120"] = COMM_1110_1120_1140(CONDITIONS["COMM1120"])
-    CONDITIONS["COMM1140"] = COMM_1110_1120_1140(CONDITIONS["COMM1140"])
+    #CONDITIONS["COMM1120"] = COMM_1110_1120_1140(CONDITIONS["COMM1120"])
+    #CONDITIONS["COMM1140"] = COMM_1110_1120_1140(CONDITIONS["COMM1140"])
     CONDITIONS["COMM1150"] = COMM_1150(CONDITIONS["COMM1150"])
-    CONDITIONS["COMM1170"] = COMM_1170_1180(CONDITIONS["COMM1170"])
-    CONDITIONS["COMM1180"] = COMM_1170_1180(CONDITIONS["COMM1180"])
-    CONDITIONS["COMM1190"] = COMM_1190(CONDITIONS["COMM1190"])
+    #CONDITIONS["COMM1170"] = COMM_1170_1180(CONDITIONS["COMM1170"])
```
Peedee2002: feel free to completely delete these lines :)

Peedee2002 (changes requested):
overall, pretty good job!
As well as the small comments i made, it would also be good if you re-ran the processors
(run `python -m runproccessors`) to make sure your changes have an effect.

backend/data/processors/conditions_preprocessing.py
```diff
@@ -432,6 +442,134 @@ def map_word_to_program_type(processed, regex_word, type):
...
+    r"Its recommended to seek a progression check prior to application.", r"", processed, flags=re.IGNORECASE
+    ) 
+    if number_of_subs > 0:
+        note += "It is recommended to do a progression check prior to enrolling."
+
+    processed, number_of_subs = re.subn(
+    r"This course is by application only. Please visit Business School website for more information<br/><br/>", r"", processed, flags=re.IGNORECASE
+    ) 
+    if number_of_subs > 0:
+        note += "This course is by application only. Please visit Business School website for more information<br/><br/>"
```
Peedee2002: avoid keeping the HTML tags. Ideally, we should run this function after they are removed.

backend/data/processors/conditions_preprocessing.py
```diff
@@ -34,6 +34,11 @@ def preprocess_conditions():
         conditions["original"] = original
         processed = original
 
+
+        processed = remove_extraneous_comm_data(conditions, processed)[1]
+        if remove_extraneous_comm_data(conditions, original)[0] != "":
+            conditions["handbook note"] = remove_extraneous_comm_data(conditions, original)[0]
```
Peedee2002: afaik this should be handbook_note, with an underscore, not a space.

backend/data/processors/conditions_preprocessing.py
```diff
@@ -432,6 +442,134 @@ def map_word_to_program_type(processed, regex_word, type):
...
+    processed, number_of_subs = re.subn(
+        r"on Good Standing and", r"", processed, flags=re.IGNORECASE
+    )
+    if number_of_subs > 0:
+        note += "Students must be in Good Academic Standing."
+
+    processed, number_of_subs = re.subn(
+        r"Good Standing", r"", processed, flags=re.IGNORECASE
+    )
+    if number_of_subs > 0:
```
imagine-hussain: There's quite a bit of excessive whitespace in this file, at the end of lines. This is a pretty inconsequential nitpick but, it is one of those things that our pylint picks up.

If you're on vscode, you don't need to manually clean this, just open up the command palette and search for `Trim Trailing Whitespace`. `cmd` `k` + `cmd x` is a shortcut for it too.

---------
The pylint output (starts line 36 of the most recent check as of writing this):
> backend/data/processors/conditions_preprocessing.py:371:0: C0303: Trailing whitespace (trailing-whitespace)
backend/data/processors/conditions_preprocessing.py:466:0: C0303: Trailing whitespace (trailing-whitespace)
backend/data/processors/conditions_preprocessing.py:481:0: C0303: Trailing whitespace (trailing-whitespace)
backend/data/processors/conditions_preprocessing.py:487:0: C0303: Trailing whitespace (trailing-whitespace)
backend/data/processors/conditions_preprocessing.py:490:5: C0303: Trailing whitespace (trailing-whitespace)
backend/data/processors/conditions_preprocessing.py:497:5: C0303: Trailing whitespace (trailing-whitespace)
backend/data/processors/conditions_preprocessing.py:503:5: C0303: Trailing whitespace (trailing-whitespace)

backend/data/processors/conditions_preprocessing.py
```diff
@@ -432,6 +442,134 @@ def map_word_to_program_type(processed, regex_word, type):
         flags=re.IGNORECASE,
     )  # hard to capture a generic case?
 
+#adds handbook notes and removes phrasing
```
imagine-hussain: Another small style nitpick but, ideally this comment would be inside the function as a docstring.
It may also be worth including an example in the docstring as though I'm able to tell what this does based off the details of the pull request, I don't know if it would be immediately obvious to someone looking at this function after the fact.

backend/data/processors/conditions_preprocessing.py
```diff
@@ -432,6 +442,134 @@ def map_word_to_program_type(processed, regex_word, type):
...
+def remove_extraneous_comm_data(conditions, processed):
```
imagine-hussain: Is the `conditions` argument being used anywhere in this function?

imagine-hussain (changes requested):
Sorry if this review comes off super nitpicky, the code looks good to me; I just thought the style issues were worth addressing since they're relatively easy fixes and were becoming a non-negligible proportion of the pylint report.

backend/data/processors/conditions_preprocessing.py
joooooooooooooooooooooooooooooooooooosh: also:
1. is the second argument mean to change from being `processed` in the first one to `original` in the next two?
2. consider only making one call in lines 39 and 40 and just storing the result in a variable, also this way you can name that variable to better communicate the intention of this bit of code

backend/data/processors/conditions_preprocessing.py
```diff
@@ -405,13 +412,16 @@ def strip_bracket_spaces(processed):
...
-# Phase 4: Common patterns
+# Phase 5: Common patterns
 # -------------------
 def uoc_in_business_school(processed):
     r"""Converts \d+UOC offered by the UNSW Business School to \d+UOC in F Business"""
     processed = re.sub(
         r"(\d+UOC) offered by the UNSW Business School", r"\1 in F Business", processed
     )
+    processed = re.sub(
+        r"(\d+UOC) of Business courses", r"\1 in ZBUS && COMM#", processed
```
joooooooooooooooooooooooooooooooooooosh: Why is the `&& COMM#` added here? The string being replaced doesn't seem to require the user to be enrolled in a certain program

backend/data/processors/conditions_preprocessing.py
```diff
@@ -432,6 +442,134 @@ def map_word_to_program_type(processed, regex_word, type):
...
+    r"Only available to single and double degree Business School students in Term 2. It will be offered to non-Business School students in Terms 1 and 3.<br/><br/>", r"", processed, flags=re.IGNORECASE
+    )
+    if number_of_subs > 0:
+        note += "Only available to single and double degree Business School students in Term 2. It will be offered to non-Business School students in Terms 1 and 3."
+    
+    processed, number_of_subs = re.subn(
+    r"Only available to single and double degree Business School students in Term 2. It will be offered to non-Business School students in Term 3.<br/><br/>", r"", processed, flags=re.IGNORECASE
+    ) 
+    if number_of_subs > 0:
+        note += "Only available to single and double degree Business School students in Term 2. It will be offered to non-Business School students in Term 3.<br/><br/>"
```
joooooooooooooooooooooooooooooooooooosh: Make sure you take the HTML out of the handbook note
edit: look at peter's comment about the same thing

backend/data/processors/conditions_preprocessing.py
joooooooooooooooooooooooooooooooooooosh: actually yeah isn't this function being run after HTML deletion? HTML deletion is one of the first stages in the processing pipeline so this will probably mean that the `re.subn` calls won't match anything

backend/data/processors/conditions_preprocessing.py
```diff
@@ -432,6 +442,134 @@ def map_word_to_program_type(processed, regex_word, type):
...
+
+    if re.search("good academic standing", processed, flags=re.IGNORECASE) != None:
+        note += "Students must be in good academic standing."
+    
+    processed = re.sub(
+    r"be in good academic standing", r"", processed, flags=re.IGNORECASE
+    )
+    
+    processed, number_of_subs = re.subn(
+    r"in their final year of a single or double Commerce degree", r"", processed, flags=re.IGNORECASE
```
joooooooooooooooooooooooooooooooooooosh: You could probably make the substitution string in some of these be "COMM#" like you have in line 544 - the note is still needed since this condition doesn't check what year they're in but it's still an implicit requirement

backend/data/processors/conditions_preprocessing.py
```diff
@@ -432,6 +442,134 @@ def map_word_to_program_type(processed, regex_word, type):
...
+    processed = re.sub(
+    r"be enrolled in a Commerce Program", r"COMM#", processed, flags=re.IGNORECASE
+    )
+    
+    # need to change this/review these
+    processed = re.sub(
+    r"myBCom First Year Portfolio", r"", processed, flags=re.IGNORECASE
+    )
+    processed = re.sub(
+    r"with ", r"", processed, flags=re.IGNORECASE
```
joooooooooooooooooooooooooooooooooooosh: What was this one for? It's generic enough that we need to be extra careful we aren't accidentally breaking anything else i think - not that we can't have it if it's needed

backend/data/processors/conditions_preprocessing.py
```diff
@@ -432,6 +442,134 @@ def map_word_to_program_type(processed, regex_word, type):
...
+    )
+    processed, number_of_subs = re.subn(
+    r"completed 30 UOC of Integrated First Year Core<br/><br/>", r"", processed, flags=re.IGNORECASE
```
joooooooooooooooooooooooooooooooooooosh: This case is going to be handled by CIRCLES-262 so I'd recommend deleting this and just keeping the manual fix for COMM1999 around so that we don't have some redundant substitutions after 262 is finished

backend/data/processors/conditions_preprocessing.py
```diff
@@ -432,6 +442,134 @@ def map_word_to_program_type(processed, regex_word, type):
...
+    ) 
+    if number_of_subs > 0:
+        note += "You must have completed 30UOC of Integrated First Year Core"
+
+    processed, number_of_subs = re.subn(
+    r"First Year core", r"", processed, flags=re.IGNORECASE
```
joooooooooooooooooooooooooooooooooooosh: Same here - just delete this one and keep the manual fix for COMM3900 around

backend/data/processors/manual_fixes/COMMFixes.py
joooooooooooooooooooooooooooooooooooosh: as well as their corresponding functions

joooooooooooooooooooooooooooooooooooosh (changes requested):
Thanks for helping automate the handbook note process - it's very tedious to do manually
I've left quite a few comments but they're all fairly little things on their own - also make sure you delete the unused manual fixes after you're done and rerun the processors to ensure new stuff hasn't broken

backend/data/processors/conditions_preprocessing.py
sonar3d: this was for specific courses that had these strings in their final processed string, definitely needs to be changed though

backend/data/processors/conditions_preprocessing.py
sonar3d: nope forgot to delete this thanks

backend/data/processors/conditions_preprocessing.py
sonar3d: removed, will move it to after the remove HTML function is run and check everything is ok

backend/data/processors/conditions_preprocessing.py
sonar3d: thanks for pointing this out, it was specifically for COMM3999 that required a commerce degree, i'll change this now

backend/data/processors/conditions_preprocessing.py
```diff
@@ -518,5 +524,118 @@ def unsw_global_degrees(processed):
...
+    )
+
+    # need to change this/review these
+    # fix comm3999
+    processed = re.sub(
+        r"(\d+UOC) of Business courses", r"\1 in ZBUS && COMM#", processed
+    )
+    # delete semi-colon
+    processed = re.sub(
+    r"; ", r"", processed, flags=re.IGNORECASE
```
imagine-hussain: Should this also replace the space after the semicolon or just the semi-colon itself?

backend/data/processors/conditions_preprocessing.py
```diff
@@ -518,5 +524,118 @@ def unsw_global_degrees(processed):
...
+    r"Only available to single and double degree Business School students in Term 2. It will be offered to non-Business School students in Term 3.", r"", processed, flags=re.IGNORECASE
+    )
+    if number_of_subs > 0:
+        note += "Only available to single and double degree Business School students in Term 2. It will be offered to non-Business School students in Term 3."
+
+    # Application and progression check notes
+    processed, number_of_subs = re.subn(
+    r"This course is by application only. Please contact the Co-op office for more information", r"", processed, flags=re.IGNORECASE
+    )
+    if number_of_subs > 0:
```
Peedee2002: this pattern of `if number_of_subs` is a little repetitive, you should make a general function to handle that logic.

```
def add_note_if_found(processed, find, note, replace_with):
    processed, number_of_subs = re.subn(find, r"", processed, flags=re.IGNORECASE)
    return processed, note + replace_with if number_of_subs > 0 else processed, note
```

you can call this function like:
```
processed, note = add_note_if_found(processed, Only available to single and double degree Business School students in Term " + terms_numbers[0] + ". It will be offered to non-Business School students in Terms " + terms_numbers[1] +" and " + terms_numbers[2] +".", note, Only available to single and double degree Business School students in Term 2. It will be offered to non-Business School students in Term 3.")
```

backend/data/processors/conditions_preprocessing.py
```diff
@@ -518,5 +524,118 @@ def unsw_global_degrees(processed):
 
 
 
+def remove_extraneous_comm_data(processed):
+    """Adds handbook notes and removes phrasing
+    example: """
+    note = ""
+    # Academic Standing note
+    processed, number_of_subs = re.subn(
```
Peedee2002: this one and the next one can be collapsed into 1 line using regex:
```
r"(Students must be)? [io]n Good Academic Standing (and)?"
```
this can be done a lot for the following examples.

Peedee2002 (approved):
Nice!

Peedee2002 (approved):

imagine-hussain (approved):
lgtm

backend/data/processors/conditions_preprocessing.py
```diff
@@ -516,7 +523,58 @@ def unsw_global_degrees(processed):
...
+    )
+    processed = re.sub(
+    r"be enrolled in a Commerce Program", r"COMM#", processed, flags=re.IGNORECASE
+    )
 
+    # need to change this/review these
+    processed = re.sub(
+        r"(\d+UOC) of Business courses", r"\1 in ZBUS && COMM#", processed
+    )
+    # delete semi-colon
```
joooooooooooooooooooooooooooooooooooosh, line 573: Very minor but this could probably be in the deletions section

joooooooooooooooooooooooooooooooooooosh (approved):

backend/data/processors/conditions_preprocessing.py
Peedee2002: I think there is a wierd ordering issue - we can ignore that problem for now.

PR #483, opened 2022-05-21
Luke/circles 322

banaboi:
Created a python script which will run the whole app without much use of docker (the database is started using it but that's it). The script will also log all console output from the frontend and backend into a .log file called 'debug'. The script is currently running at the top directory but that can change if we want.

image
Here's a sample from debug.log


joooooooooooooooooooooooooooooooooooosh (approved):

imagine-hussain (changes requested):
Can we get the `debug.log` (or really all `.log` files) added to the git ignore?
Otherwise, the script LGTM.

hdphuong (approved):

Peedee2002 (approved):

imagine-hussain (approved):
lgtm

Peedee2002:
have a look at the pylint issues tho

PR #484, opened 2022-05-21
CF-144 Allow entering marks

imagine-hussain:
<h2>Update - 29 May</h2>

Fixed most of the bugs from last time. The FE is effectively fully functional

<h3>Done & Working</h3>

- Create EditMarks component to allow user to edit their mark for a certain course
  - for privacy reasons, there is a toggle for marks -> the `export planner` will or won't show the marks on the `.png` or `.jpeg` output, depending on whether the user has elected to hide or show their marks
  - Error and success messages on update work
  - Dispatching the mark works - now also for letter grades (which is why the queue exists)
  - 1Ostensibly no migration needed since `mark` is checked for undefined on render in `<Marks` components
  - Re-validate the planner once marks are dispatched
  - pass data to BE via `util.js`

<h3>not so cool</h3>

- Kebab Menu won't render the ContextMenu - has been commented out but, the code has been left in
- <s> even though backend is (ostensibly) getting the data, warnings seem not not go away yet</s>

<s>
<h3>PR Description [Outdated]</h3>
- Create a `Marks` component to show course marks
- Add `EditMarks` component to allow a user to edit their mark for a course in the term planner
- dispatch the mark entered - via `plannerSlice.js`
- Add `Edit Mark` option to the context menu
- add error and success messages to input / output 

- Kebab menu: added component, currently unused - won't invoke the contextMenu

**TODO**
- Figure out why `const marksState  = useSelector(state => state);` fails inside `Marks.jsx` (current have a stub of mark = 100)
- pass mark to backed in `utils.js`
- Add a version migration function?</s>


imagine-hussain:
image

imagine-hussain:
image

image

image

image

image

frontend/src/components/EditMarks/EditMarks.jsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <>
+    <div className="edit-mark">
+      <div className="edit-mark-head">
+      </div>
+      <Input
+        onChange={handleInputChange}
+        onKeyDown={handleKeyDown}
+        placeholder="Enter Mark"
+      />
```
lhjt: Is it possible for this to be a `number` input component instead of allowing all inputs?

imagine-hussain: > Is it possible for this to be a `number` input component instead of allowing all inputs?

I let it allow all inputs so that the user can also enter a letter grade like `PS`, `FL`, `DN`, `HD`;
This gets validated inside of the handleConfirm function which then chooses if we should produce a success or error message.

lhjt: Ohh alright; sounds good 👍

imagine-hussain:
<h2>Update - 28 May</h2>

Fixed most of the bugs from last time. The FE is effectively fully functional

<h3>Done & Working</h3>

- Create EditMarks component to allow user to edit their mark for a certain course
  - for privacy reasons, there is a toggle for marks -> the `export planner` will or won't show the marks on the `.png` or `.jpeg` output, depending on whether the user has elected to hide or show their marks
  - Error and success messages on update work
  - Dispatching the mark works - now also for letter grades (which is why the queue exists)
  - 1Ostensibly no migration needed since `mark` is checked for undefined on render in `<Marks` components
  - Re-validate the planner once marks are dispatched
  - pass data to BE via `util.js`

<h3>not so cool</h3>


- Kebab Menu won't render the ContextMenu - has been commented out but, the code has been left in
- even though backend is (ostensibly) getting the data, warnings seem not not go away yet

<h3>Demo</h3>


https://user-images.githubusercontent.com/93496985/170739422-8493f182-2f2d-4d5e-9140-6e0fe8c961c0.mov

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
```diff
@@ -23,7 +27,7 @@ const ContextMenu = ({ code, plannedFor }) => {
     dispatch(removeCourse(code));
     validateTermPlanner(
       dispatch,
-      { years, startYear, completedTerms },
+      { years, startYear, completedTerm },
```
imagine-hussain: wait this is bad @imagine-hussain, pls remember to fix

frontend/src/reducers/coursesSlice.js
```diff
@@ -1,25 +1,38 @@
 import { createSlice } from "@reduxjs/toolkit";
 
-const initialState = {
+let initialState = {
```
gtangelo: Could this file be reset back to what we have on `dev`? Think the changes may have been a result of a conflict change but we are no longer using localstorage directly within our redux slices. Instead, persistance is done through `redux-persist`

frontend/src/components/KebabMenu/KebabMenu.jsx
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+import "./index.less";
+
+const KebabMenuIcon = () => (
```
gtangelo: Was there any reason to create our own kebab menu icon? Could we use the `<MoreOutlined />` icon from antd icons? https://ant.design/components/icon/
image

frontend/src/components/Marks/Marks.jsx
```diff
@@ -0,0 +1,22 @@
...
+  const { Text } = Typography;
+  const { mark } = useSelector((state) => state.planner.courses[courseCode]);
+
+  return (
+    <div>
+      <Text strong className="text">
+        Mark:
+      </Text>
+      <Text className="text marks-val">
+        {(mark === null || mark === undefined || mark === "") ? "N/A" : mark}
```
gtangelo, line 12: Could be simplified to `{mark || "N/A"}`?

frontend/src/components/Marks/index.less
```diff
@@ -0,0 +1,15 @@
+
+.marks-cont {
+    display: flex;
+    flex-direction: row;
+    color: green;
```
gtangelo: could we use hex code here? Same goes with `border` below for black

frontend/src/components/Marks/index.less
```diff
@@ -0,0 +1,15 @@
...
+    border: solid black 2px;
+}
+
+.marks-val {
+    margin-left: 0.8ch;
```
gtangelo, line 9: Nitpick: What is the `ch` unit lol. Never seen it haha. Prefer if we're using either `px` here.

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
```diff
@@ -1,19 +1,23 @@
...
+import EditMarks from "../EditMarks";
+
 import "./index.less";
 
 const ContextMenu = ({ code, plannedFor }) => {
   const dispatch = useDispatch();
   const navigate = useNavigate();
-  const { years, startYear, completedTerms } = useSelector((state) => state.planner);
+
+  const { years, startYear, completedTerm } = useSelector((state) => state.planner);
```
gtangelo: `completedTerm` should be `completedTerms`?

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
```diff
@@ -41,21 +45,92 @@ const ContextMenu = ({ code, plannedFor }) => {
...
+        {plannedFor && (
+          <Item onClick={handleUnschedule}>
+            <FaRegCalendarTimes className="context-menu-icon" /> Unschedule
+          </Item>
+        )}
+        <Item onClick={handleDelete}>
+          <DeleteFilled className="context-menu-icon" /> Delete from Planner
+        </Item>
+        <Item onClick={showEditMark}>
+          <FaEdit className="contextMenuIcon" /> Edit mark
```
gtangelo: Could we use `<EditFilled />` icon instead from antd icons? https://ant.design/components/icon/ Also, may want to fix the className to `context-menu-icon` instead.

frontend/src/pages/TermPlanner/DraggableCourse/index.less
```diff
@@ -16,6 +16,15 @@
   line-height: 1.5715;
   background-color: #d9d9d9;
 }
+@media (max-width: 1400px) {
+  .course {
+    border-radius: 0.2em;
+  }
+}
+
+// .course * {
```
gtangelo, line 25: Should this css be removed since its being commented out?

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
```diff
@@ -41,21 +45,92 @@ const ContextMenu = ({ code, plannedFor }) => {
...
-      <Item onClick={handleDelete}>
-        <DeleteFilled className="context-menu-icon" /> Delete from Planner
-      </Item>
-      <Item onClick={handleInfo}>
-        <InfoCircleFilled className="context-menu-icon" />
-        View Info
-      </Item>
-    </Menu>
+      </Menu>
+      <Modal
```
gtangelo: I think it would be better to encapsulate the logic of the edit marks modal within a component alongside `EditMarks`? Rather than making `EditMarks` a seperate component, instead, make the edit modal component as it's own component? This way we can encasulate all associated logic (i.e. states, handle functions, etc.) inside that component file.

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
```diff
@@ -41,21 +45,92 @@ const ContextMenu = ({ code, plannedFor }) => {
...
+    setIsEditMarkVisible(true);
+  };
+
+  const validLetterGrades = ["SY", "PS", "CR", "DN", "HD"];
+  const [markInputBuf, setMarkInputBuf] = useState(
+    useSelector((state) => state.planner.courses[code].mark),
+  );
+
+  const handleConfirmEditMark = (mark) => {
+    const attemptedMark = (` ${mark}`).slice(1).replace(" ", "");
```
gtangelo: I'm a bit confused here. Why is `mark` contained in a template literal string with a space and then sliced? Couldn't we just alter `mark` directly?

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
```diff
@@ -41,21 +45,92 @@ const ContextMenu = ({ code, plannedFor }) => {
...
+        onOk={handleConfirmEditMark}
+        onCancel={handleCancelEditMark}
+        width="300px"
+      >
+        <EditMarks
+          courseCode={code}
+          inputBuffer={markInputBuf}
+          letterGrades={validLetterGrades}
+          setInputBuffer={setMarkInputBuf}
+          handleConfirm={queueEditMark}
```
gtangelo: I'm a bit confused on why we are indirectly calling `handleConfirmEditMark` through `queueEditMark` mark? Imo adds a bit of complexity. Couldn't we just directly call `handleConfirmEditMark` instead?

frontend/src/pages/TermPlanner/LetterGradeButton/LetterGradeButton.jsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import { Button } from "antd";
+
+const LetterGradeButton = ({ letterGrade, setInputBuffer, handleConfirm }) => {
```
gtangelo: Since `LetterGradeButton` is only used once, could just include it inside the `EditMarks` folder rather than making a seperate component folder for it :)

gtangelo (changes requested):

frontend/src/pages/TermPlanner/LetterGradeButton/LetterGradeButton.jsx
imagine-hussain: Can definitely move the folder. If it's only being used once, should we just remove the component all together and just move the lettergrade responsibility entirely inside `EditMarks`?

frontend/src/pages/TermPlanner/DraggableCourse/DraggableCourse.jsx
```diff
@@ -70,17 +75,29 @@ const DraggableCourse = ({ code, index }) => {
...
+                    courseCode={code}
+                  />
+                ) : null}
+              </div>
             </div>
+            {/* <div className="kebab-container">
+              <KebabMenu
+                onClick={displayContextMenu}
+              />
+            </div> */}
```
lhjt: Should this still be included? (Likewise with the import on line 11)

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
Peedee2002: yup now this is more sane

frontend/src/components/Marks/Marks.jsx
Peedee2002: unfortunately not, because 0 is falsey.

Peedee2002:
update - marks are now correctly sent and updating :)

Peedee2002:
@gtangelo if you're happy with the core functionality, we can approve this PR. If there is any refinement needed (as in something like: the kebab menu is missing but the code we do have here is ok) we can hand that off to FE. But if there are style issues, let Hussain know :)

frontend/src/components/KebabMenu/KebabMenu.jsx
imagine-hussain: I just made our own because I didnt find it after looking for `kebab` or under the `menu` section. Since `MoreOutlined` exists, this component is no longer needed and will be removed.

frontend/src/components/Marks/index.less
imagine-hussain: `1ch` is equivalent to the width of the `0` character. i.e., `em` but for horizontal alignment.

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
imagine-hussain: To create a new string and avoid mutating the buffer set by `useState`.

Peedee2002 (approved):
lgtm based on the code I've worked with

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
Peedee2002: Wait I can't read

imagine-hussain: > I'm a bit confused on why we are indirectly calling `handleConfirmEditMark` through `queueEditMark` mark? Imo adds a bit of complexity. Couldn't we just directly call `handleConfirmEditMark` instead?

If you take the following
```jsx
const [var, setVar] = useState();

return <Component var={var} setVar={setVar} />
```
If the child component wants to call `setVar` and then callback a function from the parent like `handleConfirmEditMark`, that function will actually still use the old value of `var` because it doesn't get updated till the next reload which means, we will dispatch an older version of the input.

This isn't really an issue when the user types in input, since the act of typing in input (which is what sets the buffer) and, confirming your input, are two separate actions. However, when it comes to `letterGrade`, we want to save the user a click by cutting out the need to press `confirm` after clicking on a grade.

However, if we simply call `handleConfirmEditMark` immediately, the `ContextMenu` will not have refreshed and we will dispatch the old value already in the input. This will make it so that u actually need to click twice on the `letterGrade` to successfully update the mark.

We can avoid this by having all `confirm` actions signal that the mark is ready to be updated. Then, as soon as `ContextMenu` refreshes due to this, the mark will be dispatched and the `EditMarks` menu will be closed on success.

@Peedee2002 This is the queue I was referring to on Sunday.
 @gtangelo

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
imagine-hussain: Do you mean a hierarchy that looks like
`ContextMenu` -> `EditMarksModal`

or 

`ContextMenu` -> `EditMarksWrapper` -> `EditMarks`, where `EditMarksWrapper` contains all the logic for mark editing.

The former would require us to build our own functionality for a `Modal` component (or at least the footer with the confirm/cancel buttons), which I don't take much issue with. Though it would require a bit of `css`, it may be preferable to the latter because it cuts out a need for another wrapper around `EditMarks`.

The only reason so much of the functionality exists within `ContextMenu` rather than `EditMarks` is due by default, the functions for edit and closing of the `Modal` must be passed into the `Modal` itself.

I have no issue with either, just lmk which is preferable.

frontend/src/pages/TermPlanner/DraggableCourse/index.less
imagine-hussain: removed

frontend/src/components/Marks/index.less
imagine-hussain: done

frontend/src/pages/TermPlanner/LetterGradeButton/LetterGradeButton.jsx
gtangelo: Yep that would be better i reckon :)

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
imagine-hussain: The `slice` here is redundant if we just use `replaceAll` (which is the preferred behaviour i wanted anyways) so it's been removed to just
```jsx
const attemptedMark = (` ${mark}).replaceAll(" ", "");
```

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
gtangelo: Okay I now get what you're trying to do @imagine-hussain. You're approach is definitely valid! I guess what confused me was the fact that there was alot of one liner functions (which may or may not add a bit of bloat to the code) that can be just moved to as anonymous functions instead. Another thing that confused me was the variable name for `inputBuffer` (nothing wrong with the name itself, just never seen it named as a 'buffer' for an input value before lol - usually just seen it named more explicit i.e. in this case, somthing like `mark` or `markVal`.


> I think it would be better to encapsulate the logic of the edit marks modal within a component alongside EditMarks? Rather than making EditMarks a seperate component, instead, make the edit modal component as it's own component? This way we can encasulate all associated logic (i.e. states, handle functions, etc.) inside that component file.

I guess a more straightforward way to implement to `EditMarks` component would be something like this below (this is very rough code so it may need some improving if u want to use it lol)? This is also in relation to my comment made in that we should have the `EditMarksModal` component which encapsulates the whole editing marks logic.

```
const EditMarksModal = ({
  code, visible, setVisible,
}) => {
  const [mark, setMark] = useState(useSelector((state) => state.planner.courses[code].mark));
  const dispatch = useDispatch();
  const letterGrades = ["FL", "PS", "CR", "DN", "HD"];

  const handleConfirmEditMark = () => {
    if (
      (mark.isNaN && !letterGrades.includes(mark))
      || (parseFloat(mark) < 0 || parseFloat(mark) > 100)
    ) {
      return message.error("Could not update mark. Please enter a valid mark or letter grade");
    }
    dispatch(updateCourseMark({
      code,
      mark,
    }));
    setVisible(false);
    return message.success("Mark Updated");
  };

  const handleEnter = (e) => {
    if (e.key === "Enter") {
      handleConfirmEditMark();
    }
  };

  const handleConfirmLetterGradeMark = (grade) => {
    dispatch(updateCourseMark({
      code,
      mark: grade,
    }));
    setMark(grade);
    setVisible(false);
    return message.success("Mark Updated");
  };

  return (
    <Modal
      title={`Edit Mark: ${code}`}
      visible={visible}
      onOk={handleConfirmEditMark}
      onCancel={() => setVisible(false)}
      width="300px"
    >
      <div className="edit-mark">
        <div className="edit-mark-head" />
        <Input
          onChange={(e) => setMark(e.target.value)}
          onKeyDown={handleEnter}
          placeholder="Enter Mark"
          value={mark}
        />
        <div className="letter-grade-container">
          {
            letterGrades.map((letterGrade) => (
              <Button onClick={() => handleConfirmLetterGradeMark(letterGrade)} className="letter-grade-button">
                {letterGrade}
              </Button>
            ))
          }
        </div>
      </div>
    </Modal>
  );
};
```

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.jsx
gtangelo: See comment below. I think we can just declare the `Modal` component and it's associated functions inside `EditMarks`? :)

gtangelo:
btw @imagine-hussain, but what was the reason to make the marks input to accept all chacters rather than making it a number type again?

gtangelo (approved):

lhjt:
> btw @imagine-hussain, but what was the reason to make the marks input to accept all chacters rather than making it a number type again?

I believe it was to allow users to enter `HD` etc. directly into the field.

PR #487, opened 2022-05-23
Modifying programs_processing.py to get program structures of more degrees

Joelrwest:
Made some good progress on modifying programs_processing.py and `programsProcessed.json`. Just wanted some input on what I've done so far + ideas for how to format the weirder requirements. Have a glance at `programsProcessed.json` to see current progress.

What I've done:
- Made `programs_processing.py` (hopefully) massively more readable
- Notes are now available for each section. E.g., in CS for the major the note "COMPA1 is the default stream, and will......."
- Made locations of majors, minors and honours consistent across all degrees
- Slight changes in formatting 'other' requirements (there's still a lot to do though...)
- Completely ditched the free elective section as some programs don't even have it. Instead, we're just going to have to make some system to figure out how many free electives are left

Lingering problems/uncertainty that I'd like some input on:
- Biggest issue is the 'other' requirements (basically lumping all of `limit_rule` and `non_spec_data` into one bucket). I think the best place to see what I'm talking about it probably Science (3970) (https://www.handbook.unsw.edu.au/undergraduate/programs/2022/3970?year=2022).
        - I've been ignoring this whilst I fixed the other sections, but can anyone thing of a way to structure them that makes sense (see my current attempt in `non_spec_data`).
        - The types vary massively. It currently deals with core courses, electives with a list of courses, and electives with a list of faculties
        - It might need to be split across multiple sections instead of lumping them into `non_spec_data` as I'd originally hoped...
        - There are some small bugs with the way core courses are formatted as seen in Biotechnology honours (3053) (https://www.handbook.unsw.edu.au/undergraduate/programs/2022/3053).
        - For core courses, I've tried to mimic the way that they're represented in `specialisationsProcessed.json` in the hopes that we can reuse functionality. An I had were, what is `levels`?
- Happy with the current way it's structured? Any feedback/improvements?
- Commerce (3502) (https://www.handbook.unsw.edu.au/undergraduate/programs/2022/3502) is weird... especially myBcom which is a core requirement but has 0UOC courses. Currently, it's just dealt with by setting the `credits_to_complete` section to `null`.
- Do we want access to the 'program structure' bit of the handbook? (an easy addition)
- I think there's no shot we're going to be able to capture all cases with this, so each program will need 5 mins of a humans time as a final check
- I've certainly broken stuff with the changes, but I'll focus on them later

Sorry for the mega long post, let me know if anything is unclear


hdphuong:
Nice work refactoring programs_processing.py 🔥

Some comments:
1. The "notes" is breaking a lot of stuff so I think we should store them separately instead of in the same place with the courses/majors. If we keep them there, we'll need to make a lot of minor changes to a lot of places to make it work, which is not good practice!
2. As I mentioned in the meeting, the UOC stuff can be set as 0 when there is no information found.
3. I like the way `non_spec_data` is handled. You're right about the chance to reuse an existing function. We could integrate that with the FE just by adding these 2 lines of code to `getStructure()`:
```
    for container in programsResult['components']['non_spec_data']:
        addSubgroupContainer(structure, "General", container, [])
```

BTW, if you wanna see the changes you made reflected on the FE, check out branch `mae/api_for_new_structure`...
The FE's `general` only works when there are courses in non_spec_data and it doesn't work when there are wacky stuff like "any Aviation courses" (this is just me daydreaming but maybe we could Collab with cselectives and recommend people courses here?)

Peedee2002:
I completely agree with mae's comments here, and find your changes to be sensible.
Notes should definitely be seperated, not placed in the dictionary with the course elements IMO. I know that this data is stored this way in other parts of the code, but i find that code gross too lmao.
This can be as simple as just having a `courses` dictionary ie:
```
                "honours": {
                    "Engineering (Honours)": {
                        "notes": "Students must complete at least one of the specialisations below.",
                        "courses": {
                              "CVENBH": "Environmental Engineering",
                              "NAVLAH": "Naval Architecture",
                               "SOLAAH": "Photovoltaics and Solar Energy",
                               .....
                    }
```
 Also, thanks for changing stuff to snake_case. I understand why it was it camel case before but since this is a python project it just makes more sense that way.
1. as for splitting non-spec-data... I don't think that that is much of an issue. The problem we are trying to solve with splitting majors and minors is this:
we have minors and majors, which are connected to specialisations code. We want to make this specialisations code modular, meaning that programs has to have a way to figure out where we need to start fetching specialisations. This means that we need specific ways to tell where these minors and majors are. For the rest, we can just treat them as normal containers. 
NOTE: the only exception is in core courses. We should be able to tell they are core, no matter where they are. I dont nessesarily know what this entails yet (this is part of @joooooooooooooooooooooooooooooooooooosh's card) but i dont see this being too problematic, because getStructure should eventually dump it in a spot that sorta looks like "core courses"

2. The other thing is the wierder non-spec data in science. What we can do here is define a conversion here (similar to CS where we did COMP9 -> any level 9 comp course) which converts it to an easy thing for `getStructure`  to deal with. Have a look at how it was done for COMP, and if that method was not cursed, then we can accept that for everything else. If it is cursed, a solution we can have is one that is similar to how @imagine-hussain is doing free electives, which is using the search to figure out exactly what we need to fetch. There should be a way to analyse the packet to figure out the search value. You will have to ask hussain how he gets the packets.  We can also hardcode this if need be (my solution may be too complex so fair enough lol).

3. setting it to 0 should be fine, as long as we indicate that its core. I think that we will need a more elaborate way to deal with cores when we present that to the FE - since we will need to tell them what courses they actually need to do (without relying on UOC alone) - we will worry about that later.
4. sure

GOOD WORK!!!!!!!! This architectural stuff is an easy way to fry one's brain, but the changes you made should be great in supporting a wide range of degrees.

Peedee2002:
another thing i noticed after exploring:
this request: GET 
```
https://www.handbook.unsw.edu.au/api/content/render/false/query/+contentType:unsw_peducational_area%20+conHost:f59fc109-4aaa-40e0-bdcc-7039d31533f8%20+deleted:false%20+working:true%20+live:true%20+languageId:1%20/orderBy/modDate%20desc/limit/0
```
gives a response that maps formal school names to their abbreviation,
 and this one:
```
https://www.handbook.unsw.edu.au/api/content/render/false/query/+contentType:unsw_pOrgUnits%20+conHost:f59fc109-4aaa-40e0-bdcc-7039d31533f8%20+deleted:false%20+working:true%20+live:true%20+languageId:1%20+unsw_pOrgUnits.type:*faculty*%20+unsw_pOrgUnits.Active:*1*/orderBy/modDate%20desc/limit/0
```
does the same for the faculties. It seems that this is used to give you the conversion from "any X course" to X's code. This can make it pretty easy to deal with, as opposed to the more complex method used by gen eds.

Joelrwest:
> 1. The "notes" is breaking a lot of stuff so I think we should store them separately instead of in the same place with the courses/majors. If we keep them there, we'll need to make a lot of minor changes to a lot of places to make it work, which is not good practice!

Yep sounds good, will move them as Peter described

> 2. As I mentioned in the meeting, the UOC stuff can be set as 0 when there is no information found.

Perfect

> 3. I like the way `non_spec_data` is handled. You're right about the chance to reuse an existing function. We could integrate that with the FE just by adding these 2 lines of code to `getStructure()`:

I'll just keep doing more of the same for `non_spec_data` then

> it doesn't work when there are wacky stuff like "any Aviation courses"

Yeah I was thinking about how to deal with this... does there exist a mapping that goes from faculty -> faculty code? E.g., "Aviation" -> "AVIA". If that was possible, instead of storing a list of "any xxxxxx course", I could structure it the same as the core courses. E.g., the "any Aviation courses" would be structured as { *other courses* , "AVIA????": "any Aviation course", *other courses* } where we treat "AVIA????" as the 'code' and "any Aviation course" as the 'title'? Not sure if this is clear at all so let me know if I haven't explained what I'm thinking very well.

Joelrwest:
> Notes should definitely be seperated, not placed in the dictionary with the course elements IMO.

Yep sounds good, will basically copy this structure

>    NOTE: the only exception is in core courses. We should be able to tell they are core, no matter where they are. I dont nessesarily know what this entails yet (this is part of @joooooooooooooooooooooooooooooooooooosh's card) but i dont see this being too problematic, because getStructure should eventually dump it in a spot that sorta looks like "core courses"

I haven't tested it rigorously or really paid too much attention to it yet, but there's a Boolean that says whether it's core or not which I think is enough

> 2. The other thing is the wierder non-spec data in science. What we can do here is define a conversion here (similar to CS where we did COMP9 -> any level 9 comp course) which converts it to an easy thing for `getStructure`  to deal with. Have a look at how it was done for COMP, and if that method was not cursed, then we can accept that for everything else. If it is cursed, a solution we can have is one that is similar to how @imagine-hussain is doing free electives, which is using the search to figure out exactly what we need to fetch. There should be a way to analyse the packet to figure out the search value. You will have to ask hussain how he gets the packets.  We can also hardcode this if need be (my solution may be too complex so fair enough lol).

See the other reply to Mae's comment, what if instead of "any xxxxxx course" we just made it so that we can have wildcarded course codes like we currently have with specialisations? That would mean that something like "any Aviation course" would become a course with code "AVIA????" and title "any Aviation course", seem reasonable?

Joelrwest:
> gives a response that maps formal school names to their abbreviation, and this one:

Oops apparently I'm a little slow ahhahaha, you've already had the idea I just explained. Yep this sounds like a good idea - will look into it. The main concern I have at the moment is that this card is getting pretty long, but not really sure how to change that without doing a halfway job

PR #491, opened 2022-05-26
Fixed page loading behaviour after all requests are feteched

gtangelo:
Changes made:
- Fixed behaviour of managing state of `pageLoading` to account for async request


frontend/src/pages/CourseSelector/CourseDescription/CourseDescription.jsx
```diff
@@ -85,12 +85,14 @@ const CourseDescription = () => {
       }
     };
 
+    const fetchCourseData = async () => {
+      await Promise.all([getCourse(), getPathToCoursesById(id), getCourseCapacityById(id)]);
+      setPageLoaded(true);
```
Peedee2002, line 90: ah so that was the bug - i made a hotfix but couldnt find the root cause. Nice job spotting it!

Peedee2002 (approved):

PR #492, opened 2022-05-26
Uzaif/cf 138 handle 404 routes

Uzaif-Sheikh:
Added 404 page.
404-Screenshot


frontend/package.json
```diff
@@ -13,7 +13,7 @@
     "antd": "4.20.6",
     "axios": "0.27.2",
     "craco-less": "1.20.0",
-    "framer-motion": "6.3.3",
+    "framer-motion": "^4.1.17",
```
gtangelo: Any reason why we are downgrading the framer-motion version. If not, can we revert it back?

frontend/src/App.jsx
```diff
@@ -11,6 +11,7 @@ import "./App.less";
 import PageLoading from "./components/PageLoading";
 import "./axios";
 import Header from "./components/Header";
+import PageNotFound from "./components/PageNotFound/PageNotFound";
```
gtangelo: `import PageNotFound from "./components/PageNotFound` instead. We want to import the default export from the folder :)

EDIT: Actually, it would make sense for this component to be located under `pages` instead? Similar to CourseSelector, DegreeWizard, etc since it's technically a page.

frontend/src/components/PageNotFound/PageNotFound.jsx
```diff
@@ -0,0 +1,57 @@
...
+          <img alt="circles-logo" src={circlesLogo} width="200" height="200" />
+        </div>
+        <div className="box">
+          <img alt="circles-logo" src={circlesLogo} width="200" height="200" />
+        </div>
+      </div>
+      <div className="box-container" />
+    </div>
+    <div className="text404">
+      <Title id="headd">404</Title>
```
gtangelo: Could rename the id to something a bit clearer like `not-found` instead?

frontend/src/components/PageNotFound/PageNotFound.jsx
```diff
@@ -0,0 +1,57 @@
...
+      <Title level={2} id="page-text">PAGE NOT FOUND</Title>
```
gtangelo: Nitpick: Would prefer using classes over id here :)

frontend/src/components/PageNotFound/PageNotFound.jsx
```diff
@@ -0,0 +1,57 @@
...
+    </div>
```
gtangelo: We could also add another text underneath to redirect the user back to the home page? i.e.
```
404
Page not found
Click here to return back
```
or smth

frontend/src/components/PageNotFound/index.less
```diff
@@ -0,0 +1,106 @@
...
+  grid-template-columns: auto auto auto auto;
+  background-image: linear-gradient(to right, #ff512f, #f09819);
+  padding: 10px;
+  height: 100%;
+  width: 100%;
+}
+
+.text404 {
+  position: absolute;
+  top: 44%;
```
gtangelo: Think if you want to center it precisely using position absolute, you can do:
```
position: absolute;
top: 50%;
left: 50%;
transform: translate(-50%, -50%);
```

frontend/src/components/PageNotFound/index.less
```diff
@@ -0,0 +1,106 @@
...
+  left: 41%;
+  margin-right: -50%;
+  display: flex;
+  align-items: center;
+  flex-direction: column;
+}
+
+#headd {
+  font-size: 7em;
+  color: gray;
```
gtangelo: Could we use the hex code for the color attribute?

frontend/src/components/PageNotFound/index.less
```diff
@@ -0,0 +1,106 @@
...
+#page-text {
+  font-size: 2em;
+  color: gray;
+  margin: 0px;
+}
+
+@media only screen and (max-width: 470px) {
+  .text404 {
+    position: absolute;
+    top: 44%;
```
gtangelo: Same thing in the above comment, where you can use a different css to center

gtangelo (changes requested):
Sorry if this review seems harsh haha. 404 page looks good.

One thing of note, could we disable the circular animation of the logo for now? May be a bit disorientating to look at 😅. Otherwise, good work for a 404 page!

frontend/src/App.jsx
```diff
@@ -12,6 +12,7 @@ import "./App.less";
 import PageLoading from "./components/PageLoading";
 import "./axios";
 import Header from "./components/Header";
+import PageNotFound from "./pages/PageNotFound/PageNotFound";
```
gtangelo: do `./pages/PageNotFound` instead

frontend/src/pages/PageNotFound/PageNotFound.jsx
```diff
@@ -0,0 +1,63 @@
...
+          <div className="box">
+            <img alt="circles-logo" src={circlesLogo} width="200" height="200" />
+          </div>
+        </div>
+        <div className="box-container" />
+      </div>
+      <div className="text404">
+        <Title className="not-found">404</Title>
+        <Title level={2} className="page-text">PAGE NOT FOUND</Title>
+        <Button type="primary" size="large" onClick={() => navigate("/degree-wizard")}>Go Back Home</Button>
```
gtangelo: Think u can just redirect to the root route `/` instead?

gtangelo (approved):
just those fixes and lgtm!

PR #493, opened 2022-05-28
Circles 268 insufficient dependancies

Peedee2002:
image
current state^

so, this is all cool, but this PR still needs a lot of work to go into production:
1. we need a better way to split up nodes - its not currently very good, and i generated the above image just by dragging stuff. We ideally want a way that is better.
2. the backend gets spammed on this page atm - we may want to not make that happen. Think about this when we come up with which methods to use
3. this is a proof of concept, so there is no functionality (no adding courses etc)

Overall - i like the state that it is in, but it will take some time to flush this out. 
Also, feel free to talk to james about his old code, and weigh the pros and cons.

probably needs many eyes to look at - this PR will probably take a while.

NOTE: we want the state of the PR to reach a state where the FE can take over from me - not that we can actually deploy this feature. This feature will stay in dev for a while im assuming.


frontend/src/pages/GraphicalSelector/GraphicalSelector.jsx
```diff
@@ -0,0 +1,74 @@
...
+  let graph = null;
+  const loadGraph = () => {
+    if (!graph) {
+      graph = new G6.Graph({
+        // eslint-disable-next-line react/no-find-dom-node
+        container: findDOMNode(ref.current),
+        width: 1920,
+        height: 870,
+        modes: {
+          default: ["drag-canvas", "drag-node"],
```
imagine-hussain, line 22: In terms of cleaning up distribution of nodes, does the `collapse-expand` conflict with the draggability?
If not, allowing the user to collapse nodes dependent on a current node could be pretty useful for decluttering.

frontend/src/pages/GraphicalSelector/GraphicalSelector.jsx
```diff
@@ -0,0 +1,74 @@
...
+        await axios.get(`/programs/getStructure/${programCode}/${majors.join("+")}${minor && `/${minor}`}`)
+      ).data;
+      const courseList = (
+        Object.values(res.structure)
+          .flatMap((specialisation) => Object.values(specialisation)
+            .filter((spec) => typeof spec === "object" && spec.courses)
+            .flatMap((spec) => Object.keys(spec.courses)))
+      );
+      setCourses(courseList);
+      const res2 = await Promise.all(courseList.map((c) => axios.get(`/courses/getPathFrom/${c}`)));
```
imagine-hussain, line 53: Can we pre-compute the graph structure in the BE, store that data structure in our  database and expose that through an API instead of needing to loop over every course and remake the graph on the FE.
It could save quite some computation on the FE esp for slower devices.

From what I can tell so far, this structure isnt contingent on the user planner so we could just store this data and only refresh it when it's required which could save network usage, which is probably a bigger issue than the computation issue brought up earlier.

In that case, we could store a `lastUpdated` value on both the FE and BE and then we first need to make a call to check if an update is required - which will usually return false - and then call for the data structure again only if that value is true.

frontend/src/hooks/useWindowDimensions.js
```diff
@@ -0,0 +1,26 @@
+import { useState, useEffect } from "react";
+// https://stackoverflow.com/questions/36862334/get-viewport-window-height-in-reactjs
+const getWindowDimensions = () => {
+  const { innerWidth: width, innerHeight: height } = window;
```
imagine-hussain: not sure if it would work but, could this line be removed so that this function just returns
```jsx
{
  width: window.innerWidth,
  height: window.innerHeight,
}
```

frontend/src/pages/GraphicalSelector/GraphicalSelector.jsx
Peedee2002: The thing i am worried about is that we may decide to use another library (like react-flow-renderer) if that turns out to be preferred by the FE team. If we solidify our structure, ill be happy to do this.

frontend/src/hooks/useWindowDimensions.js
Peedee2002: oh i realised i stopped using this file. Ill remove it.

frontend/src/pages/GraphicalSelector/GraphicalSelector.jsx
Peedee2002: yea - im not so sure if it works with this preset. putting it there didnt really do much.

Peedee2002:
another thing i noticed - ive basically pushed both of path_from and path_to to the API layer - we may consider moving them back lower down, but this requires a dependancy with conditions.pkl lower down the stack. Idk if that is the best idea, but it may be necessary.

backend/algorithms/objects/conditions.py
```diff
@@ -36,6 +37,21 @@ def validate(self, user: User) -> tuple[bool, list[str]]:
...
+
+    def beneficial(self, user: User,  course: dict[str, Tuple[int, int]]) -> bool:
+        """ checks if 'course' is able to meet any *more* subtrees' requirements """
+        course_name = list(course.keys())[0]
+        if self.validate(user)[0] or user.has_taken_course(course_name):
+            return False
+        user.add_courses(course)
+        answer = self.validate(user)[0]
+        user.pop_course(course_name)
+        return answer
```
Joelrwest, line 54: It's clear to me at least what's happening in this function (although I haven't really worked on this bit of the code), possibly some comments would help?

backend/algorithms/objects/conditions.py
```diff
@@ -265,7 +309,11 @@ def __init__(self, exclusion: str):
         self.exclusion = exclusion
 
     def validate(self, user: User) -> tuple[bool, list[str]]:
-        return not user.has_taken_course(self.exclusion), []
+        is_valid = not user.has_taken_course(self.exclusion)
+        return is_valid, [] if is_valid else [f"{self.exclusion} is an exclusion course for this one"]
```
Joelrwest: Function typing of tuple[bool, list[str]] isn't always true since it's sometimes just list[str] - is it possible to always return the same type or not really?

backend/algorithms/objects/conditions.py
```diff
@@ -317,6 +368,15 @@ def validate(self, user: User) -> tuple[bool, list[str]]:
 
         return satisfied, sum(warnings, [])  # warnings are flattened
 
+    def is_path_to(self, course: str) -> bool:
+        return any(condition.is_path_to(course) for condition in  self.conditions)
```
Joelrwest: Double space between "in" and "self.conditions"

Joelrwest (changes requested):
LGTM, I'm pretty illiterate on the FE stuff though

frontend/src/pages/GraphicalSelector/GraphicalSelector.jsx
imagine-hussain: Sounds good; we can just keep it in mind for once the FE structure gets solidified.

gtangelo (approved):
Just reviewing code from the frontend since I can't understand BE that well rip

This is a great proof of concept for a graphical selector. Obviously, some things needs to be fleshed out (i.e. nodes, graph structure, user flow, etc) but I think this is a great start!

In regards to the BE request spam, perhaps this could be avoided if we just slowly load up each node when a user has selected a course (i.e. how the og circles loads up new course nodes). Though then again, if we want the option to show all course nodes for the degree, may need to think about the structure for that.

I'm happy for the FE team to take direction after this unless BE wants to work on the graphical selector as well :). But otherwise, would be okay for this to be merged in to dev for the FE side since this feature shouldn't appear on prod anyways due to the `REACT_APP_ENV` config

backend/data/final_data/conditionsProcessed.json
```diff
@@ -8490,33 +8490,33 @@
     },
     "MGMT4101": {
         "original": "MGMT2102 and completion of 72 UOC (if you are enrolled the Commerce International [program 3558], completion of the Commerce Overseas Program (Exchange) is also required)<br/><br/>",
-        "processed": "4501",
-        "handbook_note": "enrolment in the International Business Honours plan is required"
+        "processed": "MGMT2102 && 72UOC && 3558",
+        "handbook_note": "Commerce Overseas Program (Exchange) is also required"
```
Joelrwest: Why was a manual fix done for these that doesn't seem to match the original requirement?
- https://www.handbook.unsw.edu.au/undergraduate/courses/2022/MGMT4101/
- https://www.handbook.unsw.edu.au/undergraduate/courses/2022/MGMT4104/

backend/data/processors/manual_fixes/MGMTFixes.py
```diff
@@ -122,30 +119,13 @@ def MGMT_3730():
...
-    "processed": "Program 4501 && the Human Resource Management Honours plan"
-    """
+def MGMT_4101_4500_4501_4104_4738_4739():
     # Have not been able to locate a code for the HRM honours plan.
     return {
         "original": CONDITIONS["MGMT3110"]["original"],
-        "processed": "4501",
-        "handbook_note": "enrolment in the Human Resource Management Honours plan is required"
+        "processed": "MGMT2102 && 72UOC && 3558",
+        "handbook_note": "Commerce Overseas Program (Exchange) is also required"
```
Joelrwest: Here is the manual fix I was talking about

Joelrwest (approved):

PR #498, opened 2022-05-30
Leo/pg/cf 219 & cf 60 grid view display fix and additions

leonardo-fan:
- made gridview and removed previous list view
- added css to highlight added courses that have been completed in previous terms

image


gtangelo (approved):
lgtm! Awesome work for this! :)

PR #499, opened 2022-05-30
cf-214: Course Menu sidebar to display course names

kellieyau:
added course title and code to course menu container on full screen and course code only if reduced window size

reduced window size:
image

full window size:
image


gtangelo (approved):
lgtm!

PR #500, opened 2022-05-30
added button to switch between views

leonardo-fan:
- button that when pressed switches whether table or grid view is shown (went a bit off ticket spec because thought it'd be more visible if the text was in the button instead of a tooltip but let me know what you think)
Screen Shot 2022-05-31 at 12 03 41 am
Screen Shot 2022-05-31 at 12 03 53 am


gtangelo (approved):
lgtm!

PR #504, opened 2022-06-03
add condition for core courses

joooooooooooooooooooooooooooooooooooosh:


backend/algorithms/objects/user.py
```diff
@@ -21,6 +21,7 @@ def __init__(self, data = None):
         self.program: str = None
         self.specialisations: dict[str, int] = {}
         self.year: int = 0
+        self.core_courses: list[str] = []
```
Peedee2002, line 24: Is there currently a way to populate this?

Peedee2002 (changes requested):
Nice, but we are also gonna need some tests for user.has_taken_cores()

This can be pretty simple because the code isn't complex, but yea. Also, make sure that the cores field can be initialised properly when passing a user json. I can't tell if youre currently handling that case.

backend/algorithms/objects/user.py
joooooooooooooooooooooooooooooooooooosh: no that'll be in the next stage where we add this functionality to the frontend

joooooooooooooooooooooooooooooooooooosh: this pr is just outlining the structure etc

joooooooooooooooooooooooooooooooooooosh:
> Nice, but we are also gonna need some tests for user.has_taken_cores()

do you mean has_taken_course() or completed_core()?
has_taken_course() was already there we just changed it to utilize hussain's marks pr

Peedee2002:
> > Nice, but we are also gonna need some tests for user.has_taken_cores()
> 
> do you mean has_taken_course() or completed_core()? has_taken_course() was already there we just changed it to utilize hussain's marks pr

Ah yes sorry I meant both. Just 1 test for a course with less than 50% and another few tests for core.

backend/algorithms/objects/user.py
Peedee2002: Ah fair

PR #505, opened 2022-06-04
fixed bug 400 when no minor is selected

hdphuong:
i love javascirpt 🥰


Peedee2002 (approved):
Lol I don't get the difference but that's ok

PR #509, opened 2022-06-04
CF-231 Use react-helmet and removed unused public json files

gtangelo:
- removed unused json files
- Used react-helmet to try improving seo


Uzaif-Sheikh (approved):
Looks good.

PR #510, opened 2022-06-04
fix bug to show loading skeleton when fetching new course data

gtangelo:
- show loading skeleton on course description when refetching new course data


kellieyau:
So just to cinfrm the logic:
1. useEffect is no longer required because it should show the loading skeleton when a new course is requested not necessarily when the page is rendered?
2. also "c" is now added as a parameter in getCourse as it is the new course? 
3. I dont get why id was used in the orginal code

gtangelo:
We're still using useEffect in the code. Just cleaning up the code by moving the functions around :)

Adding the param "c" in getCourse to make it consistent with the other functions.

kellieyau (commented):
yep looks good :D

kellieyau (approved):

PR #513, opened 2022-06-05
Feature/cf 228 add styled components

gtangelo:
- In future, we want to migrate the use of less to styled components due to the benefits that it brings (i.e. encapsulation, will probably make it easier to integrate dark mode, etc.)
- This PR is just adding `styled-components` for use as a dependency.
- Migrated some simple components to use styled components over less


leonardo-fan (approved):
Lgtm 😄  Had to do a quick tutorial on styled components before reviewing lul. Question though, with the styles.js files, is it be better to do ```export default { constThing }``` or ```export default constThing``` or does it not matter?

kellieyau (approved):
I think this lgtm, does this mean we'll officially have dark mode now 👀

xincczn:
LGTM! In my limited knowledge, doesn't look dodgy to me. :>

xincczn (approved):

gtangelo:
@leonardo-fan This is from my experience but generally, I think it's a bit cleaner to export components as part of an object i.e. `export default { component1, component2 }`. This is basically exporting an object that contains `component1` and `component2` as the key-pair values for that object. So you could do `import S from './styles'`, and since we are exporting it as the default, `S` would be `S = { component1, component2 }`. So you can access styled components like `<S.component1/>` or `<S.component2/>`.

The other approach is directly exporting it i.e. `export component1; export component2;` This is also fine, but you would need to import it directly i.e. `import { component1, component2 } from './styles.js'`. But as you can imagine, if you have like a bunch of exports, there's gonna be a lot of code lines just importing styled components XD.

TLDR: it doesn't matter functionality wise but `export default { component1, component2 }` just makes it easier to import components since it would be access via the default import and doesn't add that many lines compared if you have to import alot of components, if that makes sense LOL.

@kellieyau yee hopefully, this approach should make it easier to configure dark mode later on! :D 

@xincczn totally not dodgy code 👀

PR #520, opened 2022-06-09
Tests/circles 320 evaluate cores token

imagine-hussain:
tests for `User.has_taken` to ensure correctness for the the new marks-compatible paradigm

Thoughts:
- Many of these would be very ripe for property based testing (if that's something we're moving to eventually - i remember james saying we have a preference for it); This would however required tests to get access to the course list

### DIFF
--- a/backend/algorithms/exampleUsers.json
+++ b/backend/algorithms/exampleUsers.json
@@ -6,8 +6,8 @@
             "ACCTA2": 1
         },
         "courses": {
-            "COMP1511": [6, null],
-            "COMP1521": [6, null]
+            "COMP1511": [6, 84],
+            "COMP1521": [6, 57]
         },
         "year": 3
     },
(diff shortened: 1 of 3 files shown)


Peedee2002:
Yes, I love property based testing. I don't think it requires access to the course list either (make up marks for fake courses for example)

imagine-hussain:
> Yes, I love property based testing. I don't think it requires access to the course list either (make up marks for fake courses for example)

We should make a card for converting tests to be property based.

imagine-hussain:
> > Yes, I love property based testing. I don't think it requires access to the course list either (make up marks for fake courses for example)
> 
> We should make a card for converting tests to be property based.

also, now that I think about it, you're right, we can use any made-up courses we like

imagine-hussain:
Also, for some reason CI isn't re-running so, proof of pass
image

ollibowers:
looks insanely good to me!

ollibowers (approved):

joooooooooooooooooooooooooooooooooooosh (approved):

PR #522, opened 2022-06-09
better load_json_tests

imagine-hussain:
user loading tests
- test initialisation
- ensure that load_json on empty dict or missing elements does not crash nor, change the user object


ollibowers (approved):
lets get this merged!

PR #526, opened 2022-06-11
fix: Changed minor step button text for better userflow.

xincczn:
Allow minor step button to display "Skip" instead of "Next" when no minors are selected.


gtangelo (approved):
lgtm!

PR #527, opened 2022-06-12
Moved Header to styled components and restructure PageTemplate component layout

gtangelo:
Changes made:
- moved `Header` component to use styled components
- removed the use of antd `Layout` component since it was not needed
- Streamlined component structure for each page to include `PageTemplate` and `Header` in respective pages components rather in `App.jsx`


ollibowers (approved):
very exciting, lgtm

PR #533, opened 2022-06-13
feat: added expanded support for many more degrees

Peedee2002:
so i added support to many more degrees, including ones that are honours only, like engineering. This required a slight re-tweaking of the app, which included ignoring the difference between the types of specialities. So, i refactored stuff to make that work across the FE and BE. the change to localstorage is trivial, so ill probably not need to break their build and instead allow for a migration to occur.

This is still a little bit of a WIP because i think my FE design can be improved slightly (especially with the new degrees). Ill look more into that soon.


frontend/package.json
```diff
@@ -77,6 +77,6 @@
     "eslint-plugin-react": "7.30.0",
     "eslint-plugin-react-hooks": "4.5.0",
     "eslint-plugin-simple-import-sort": "7.0.0",
-    "react-error-overlay": "6.0.11"
+    "react-error-overlay": "6.0.9"
```
Peedee2002, line 80: this is annoying - but it is the only way to make sure hot-reload works as expected.

frontend/src/pages/DegreeWizard/DegreeWizard.jsx
```diff
@@ -108,34 +112,23 @@ const DegreeWizard = () => {
...
+                        currStep={currStep}
+                        incrementStep={incrementStep}
+                      />
+                    )
+                      : index + 1 === stepList.length ? <StartBrowsingStep />
+                        : <SpecialisationStep incrementStep={incrementStep} type={stepName} />
+                }
+              </div>
+            )
+          ))}
```
Peedee2002: this can be less ugly somehow i think - does an FE person have any better ideas?

frontend/src/pages/CourseSelector/CourseSelector.jsx
```diff
@@ -16,7 +16,7 @@ const CourseSelector = () => {
   const dispatch = useDispatch();
```
Peedee2002, line 17: in this file, i was also thinking that we should find a way to sort the containers - possibly by level? not sure.

frontend/src/reducers/degreeSlice.js
```diff
@@ -3,8 +3,7 @@ import { createSlice } from "@reduxjs/toolkit";
 const initialState = {
   programCode: "",
   programName: "",
-  majors: [],
-  minors: [],
+  specs: [],
```
Peedee2002, line 6: this is the localstorage thing i messed up - i think its a trivial thing to fix using a migration though, right?

frontend/package.json
zax-xyz: alternatively you could just migrate to react-scripts v5 (we had the [same problem in chaos](https://github.com/csesoc/chaos/pull/149), the only [breaking change from 4.x to 5.x](https://github.com/facebook/create-react-app/blob/main/CHANGELOG.md#breaking-changes) that should really matter is the change in eslint parsers)

Peedee2002: Yea the problem we have with that is Craco not being able to support CRA 5 (this rabbit hole goes deep lmao)

zax-xyz: ok surely just ditch CRA and migrate to [vite](https://vitejs.dev) (i am going to peer pressure chaos to do the same)

Peedee2002: I was shilling for nextjs as well lmao

@gtangelo plz

frontend/src/pages/DegreeWizard/DegreeWizard.jsx
imagine-hussain: Perhaps, could pre-compute these conditionals and store their values or even the entire component being rendered here in a variabel

backend/data/final_data/conditionsTokens.json
```diff
@@ -26119,50 +26090,40 @@
...
+        "CCDF#",
         ")"
     ],
     "ZEIT4902": [
         "(",
-        "Enrolment",
-        "in",
-        "CDF",
-        "Program",
+        "CCDF#",
```
imagine-hussain, line 25379: this is very nice improvement

backend/server/routers/specialisations.py
```diff
@@ -0,0 +1,81 @@
...
+    responses={
+        400: {
+            "description": "The given program code could not be found in the database",
+        },
+        200: {
+            "description": "Returns all majors to the given code",
+            "content": {
+                "application/json": {
+                    "example": {
+                        "spec": {
```
hdphuong, line 44: shouldn’t it be `majors` here?

backend/server/routers/programs.py
```diff
@@ -281,12 +199,7 @@ def getStructure(
     if major:
         majors_l = major.split("+") if "+" in major else [major]
         for m in majors_l:
```
hdphuong: nitpicky but replacing `major` with `spec` here cos better naming

hdphuong (approved):
Amazing work 💯 I agree that grouping minors, majors, honours to specialisations is a great idea. 

Just one thing: the ranking of the search function is not working for honours (yet). But it’s not breaking anything so maybe another PR? 

(Also, can we capitalise the first letter in “**M**ajors for [something]”? It looks off non-capitalised)

Other than that, LGTM!

frontend/src/pages/DegreeWizard/DegreeWizard.jsx
zax-xyz: Perhaps could just extract this into a generic `Step` component where you can use regular `if` statements to conditionally render instead of doing wacky ternary conditions inside the jsx. I suppose hussain's suggestion of precomputation could also work, with something like a `steps[index] ?? <SpecialisationStep ...`

Personally I would also use `.slice` before `map` here instead of having that `currStep >= index + 1` condition adding an extra layer of complexity (btw actually took me a while to realise you were rendering the first n, not the last n, because of the `>=` confusing me - `index` being on the right is strange imo but that's just my personal style preference)

Also should probably be giving the divs here a `key` prop? :p

frontend/src/reducers/degreeSlice.js
gtangelo: should be easy to create a migration function between versions. Though you may want to test it beforehand if it migrates correctly

frontend/src/pages/DegreeWizard/steps/SpecialisationStep.jsx
```diff
@@ -3,51 +3,49 @@ import { useDispatch, useSelector } from "react-redux";
...
         <Title level={4} className="text">
-          What do you major in?
+          What are your {type}?
         </Title>
-        {majors.length > 0 && currStep === 3 && (
-          <Button type="primary" onClick={incrementStep}>
-            Next
-          </Button>
-        )}
+        <Button type="primary" onClick={incrementStep}>
```
gtangelo: Hmm could we have back check if it is in the current step when to render the button? So we don't display `Next` button when moving to the next step for the previous step.

frontend/src/pages/DegreeWizard/steps/SpecialisationStep.jsx
```diff
@@ -3,51 +3,49 @@ import { useDispatch, useSelector } from "react-redux";
...
       </div>
 
       <Menu
         className="degree-specialisations"
-        onSelect={(e) => dispatch(addMajor(e.key))}
-        onDeselect={(e) => dispatch(removeMajor(e.key))}
-        selectedKeys={majors}
+        onSelect={(e) => dispatch(addSpecialisation(e.key))}
+        onDeselect={(e) => dispatch(removeSpecialisation(e.key))}
+        selectedKeys={specs}
```
gtangelo, line 43: Nitpick (Dont need to fix this - just pointing it out): Since we are using `specs` for the selected keys, for example, if this was a minor step, `selectedKeys` would contain the major `spec` as part of it. Since `specs` is globalised outside of each step and contains selected keys for majors, minors and honours. 

This doesn't really affect things, maybe for an edge case where if a major and minor contains the same key, then it would affect each other options when selecting? Though, most likely every specialisation is unique so we wouldn't have this issue?

frontend/src/pages/DegreeWizard/steps/SpecialisationStep.jsx
```diff
@@ -3,51 +3,49 @@ import { useDispatch, useSelector } from "react-redux";
...
         defaultOpenKeys={["0"]}
         mode="inline"
       >
         {Object.keys(options).map((sub, index) => (
           <Menu.SubMenu
             key={index}
-            title={`Majors for ${sub}`}
+            title={`${type} for ${sub}`}
```
gtangelo: Ooof anyway to capitalise the first letter? i.e. Instead of `majors for Computer Science`, it could be `Majors for Computer Science`?

frontend/src/pages/DegreeWizard/DegreeWizard.jsx
gtangelo: Couldn't you have it where you use `map` on `steps` rather than `stepList`? Whilst also keep the old way of rendering the year, degree and start browsing step (by hard coding it)?

Something like this:
```
{steps.map((stepName, index) => (
    currStep >= index + 3 && (
    <div className="step-content" id={stepName}>
      <SpecialisationStep incrementStep={incrementStep} type={stepName} />
    </div>
    )
  ))}
```
May not be the best approach but i feel its more readable than nested ternary statements 😅

gtangelo (approved):
Nice refactoring and addition of specialisation. Lgtm in terms of the FE side. Just address the comments made and update the redux persist version and everything can be merged in for the frontend.

- [x] address PR comments
- [x] update local storage redux persist config (either migrate or hard reset)

backend/server/routers/specialisations.py
Peedee2002: nope - ive unified this so it can be a major, minor, or honours depending on request.

frontend/src/pages/DegreeWizard/steps/SpecialisationStep.jsx
Peedee2002: they cant have the same key, so no worries

Joelrwest (approved):

Peedee2002:
One thing that's bad in this PR - the new stuff we support are much less well supported. There are likely bugs in a few of them at least. One I noticed was the Phys bug, which was missing brackets. It's gonna get a lot harder to keep bugs at bay if we merge this since the scope is much larger. Maybe I need to amend the disclaimer to tell people that we are dumdums.

joooooooooooooooooooooooooooooooooooosh (approved):

PR #537, opened 2022-06-15
feat(misc): added darkmode toggle to small navbar

ollibowers:
- Added darkmode toggle to the small screen navbar.

- Removed the customization box from the drawer.

### DIFF
--- a/frontend/src/components/Header/Header.jsx
+++ b/frontend/src/components/Header/Header.jsx
@@ -20,6 +20,57 @@ const Header = () => {
   const [showDrawer, setShowDrawer] = useState(false);
   const { pathname } = useLocation();
 
+  const smallHeader = (
+    <S.HeaderContent>
+      {inDev && <ThemeToggle />}
+      <div style={{ margin: "10px" }}>
+        <Button
+          type="primary"
+          size="large"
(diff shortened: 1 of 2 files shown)


imagine-hussain:
lgtm

imagine-hussain (approved):
lgtm; good work

frontend/src/components/Header/Header.jsx
```diff
@@ -20,6 +20,57 @@ const Header = () => {
...
+        />
+      </div>
+    </S.HeaderContent>
+  );
+
+  const largeHeader = (
+    <S.HeaderContent>
+      <Menu
+        theme="dark"
+        selectedKeys={[pathname.split("/")[1]]}
```
zax-xyz, line 41: you could use the optional `limit` parameter for [split](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split) here so that it doesn't split the whole string, to marginally improve performance to a perhaps negligible degree

ollibowers: Thanks, will make this to another ticket

ollibowers: I did not write this code.

PR #539, opened 2022-06-17
Leo/pc/cf 251 grid view distinguish future planned courses

leonardo-fan:
- show future courses in grid view
- refactored planner course formatter util
- refactored course badge ternaries into function

https://user-images.githubusercontent.com/90736577/174219205-fc0ee688-938f-4fa8-9b35-5e62997303e3.mov


frontend/src/pages/ProgressionChecker/GridView/GridView.jsx
```diff
@@ -75,14 +81,92 @@ const GridView = ({ isLoading, structure }) => {
...
+    setPlannerCourses(getFormattedPlannerCourses(years, startYear, courses));
     generateGridStructure();
   }, [isLoading, structure, years, startYear, courses]);
 
   const handleCourseLink = (courseCode) => {
     navigate("/course-selector");
     dispatch(addTab(courseCode));
   };
+
+  const courseBadge = (course) => {
```
gtangelo: Delegate this reposibility into a seperate component file instead?

frontend/src/pages/ProgressionChecker/GridView/GridView.jsx
```diff
@@ -75,14 +81,92 @@ const GridView = ({ isLoading, structure }) => {
...
+          onClick={() => handleCourseLink(course.key)}
+        >
+          {course.key}: {course.title}
+        </Button>
+      );
+    }
+
+    // for future courses planned
+    // course.past can be undefined if not in term planner thus check for false
+    if (course.past === false) {
```
gtangelo: Could just do `course.past && course.planned`?

frontend/src/pages/ProgressionChecker/GridView/GridView.jsx
```diff
@@ -75,14 +81,92 @@ const GridView = ({ isLoading, structure }) => {
...
+      return (
+        <Badge
+          key={course.key}
+          count={(
+            <Tooltip title="Course added but not planned">
+              <div className="courseBadgeIcon">!</div>
+            </Tooltip>
+          )}
+        >
+          <Button
```
gtangelo: The button could be probably extracted to a seperate functional component since you are basically using the same params? Make's it easier to make any future changes through one component rather than copying and pasting.

OR

This could be abstracted out somehow? Like maybe have a config object, do the logic of the if statements above. Right now, there is a lot of duplicated jsx lying around.

frontend/src/pages/ProgressionChecker/GridView/GridView.jsx
```diff
@@ -39,7 +40,10 @@ const GridView = ({ isLoading, structure }) => {
               newGridLayout[group][subgroup].push({
                 key: courseCode,
                 title: subgroupStructure.courses[courseCode],
-                completed: courseCode in pastCourses,
+                planned: courseCode in plannerCourses,
```
gtangelo: wait what is the point in having `planned` when we can just use the `not` operator on `unplanned` and vice versa?

frontend/src/pages/ProgressionChecker/GridView/GridView.jsx
```diff
@@ -56,7 +60,9 @@ const GridView = ({ isLoading, structure }) => {
                 newGridLayout[group][subgroup].push({
                   key: courseCode,
                   title: courseData.title,
-                  completed: courseCode in pastCourses,
+                  planned: courseCode in plannerCourses,
+                  past: plannerCourses[courseCode].past,
```
gtangelo, line 52: Maybe we can remove keeping track of the `past` field here? Don't think it's necessary to keep track of it where we can just calculate on the fly (inside `courseBadge`)?

Not 100% sure about if this is possible tho. Up to you if you want to keep this field in or not. :)

gtangelo (changes requested):
Good changes. Think there could be some improvements with the code tho :)

frontend/src/pages/ProgressionChecker/GridView/GridView.jsx
leonardo-fan: doesn't seem like this works because by removing planned I'm using ```!unplanned``` and since it evaluates to true even when the course isn't planned, the incorrect courses are getting classified as future planned courses

frontend/src/pages/ProgressionChecker/GridView/GridView.jsx
leonardo-fan: planned is if it is in the termPlanner while unplanned is if it is in the unplanned column of the termPlanner - ```!unplanned``` would capture everything outside the unplanned column i.e. planned courses and courses not in the termPlanner.

frontend/src/pages/ProgressionChecker/GridView/GridView.jsx
leonardo-fan: could calculate inside courseBadge but I'm using the logic from the ```getFormattedPlannerCourses``` function that loops through the planner, would have to repeat using that inside the courseBadge

frontend/src/pages/ProgressionChecker/GridView/GridView.jsx
gtangelo: Ohhh okie get what you mean then

frontend/src/pages/ProgressionChecker/GridView/GridView.jsx
gtangelo: Oh interesting I see. That's weird LOL

gtangelo (approved):
@leonardo-fan I'm approving this PR so feel free to merge it in whenever. Though I think the only improvement would be the course bade component to make :)

leonardo-fan:
> @leonardo-fan I'm approving this PR so feel free to merge it in whenever. Though I think the only improvement would be the course bade component to make :)

sweet, yee was just about to push it hehe

PR #540, opened 2022-06-18
Got rid of line that prints warnings in programs_processing.py

Joelrwest:


Peedee2002 (changes requested):
I don't think we should completely remove this warning. Can we instead store it as json somewhere?

Joelrwest:
> I don't think we should completely remove this warning. Can we instead store it as json somewhere?

Yep it's stored in each programs "processing_warnings", which has more specific warnings. I guess I could also make a JSON to isolate the warnings instead of being mixed in with all the other program information (like it is now)?

Peedee2002 (approved):
Oh having it in the program is even better. I'm happy with removing this then

PR #549, opened 2022-06-23
Elliot/circles 326

sixohtoo:
CIRCLES-326
Implemented multiterm courses and added a new attribute to the frontend and backend to determine whether a course is a 'multiterm course'.

image


gtangelo (changes requested):
~~Haven't looked at the code just yet, played it on local but just an FYI, there seems to be some bugs I found with the term planner:~~
- ~~splice error~~
image
- ~~cant rearrange the order of courses within a term box at some points~~

Ooops nevermnind, didnt rebuild the BE, will test/look at the PR later

gtangelo:
Think there are some bugs with the dragging indexes/positions in a term box? 😅 
https://user-images.githubusercontent.com/65492346/175433904-34c0f117-6dbf-4f4b-ab58-a160f01a908a.mp4

frontend/src/pages/CourseSelector/CourseDescription/AddToPlannerButton.jsx
```diff
@@ -48,6 +48,7 @@ const AddToPlannerButton = () => {
         handbookNote: course.handbook_note,
         isAccurate: course.is_accurate,
         mark: null,
+        isMultiterm: course.is_multiterm,
```
gtangelo: I think because you added an extra attribute, you may need to update the redux persist version? https://compclub.atlassian.net/wiki/spaces/C/pages/2235465792/Updating+redux-persist

frontend/src/reducers/plannerSlice.js
```diff
@@ -78,20 +79,90 @@ const plannerSlice = createSlice({
       state.unplanned = state.unplanned.filter(
         (c) => c !== course,
       );
-      state.years[destRow][destTerm].splice(destIndex, 0, course);
+
+      if (state.courses[course].isMultiterm) {
+        const { UOC: uoc, termsOffered } = state.courses[course];
+        const termsList = getTermsList(destTerm, uoc, termsOffered, state.isSummerEnabled);
+
+        // Add ccourse multiple times
```
gtangelo: nitpick: type for `ccourse`

frontend/src/pages/TermPlanner/TermBox/TermBox.jsx
```diff
@@ -64,10 +64,11 @@ const TermBox = ({
             {coursesList.map(
               (code, index) => (
                 <DraggableCourse
-                  key={code}
+                  key={`${code}${term}`}
                   code={code}
                   index={index}
                   showMarks={showMarks}
+                  term={term || ""}
```
gtangelo: Why is this `||` an empty string?

frontend/src/pages/TermPlanner/utils.js
```diff
@@ -82,4 +82,80 @@ const prepareCoursesForValidation = (planner, degree, suppress) => {
   return payload;
 };
 
-export { getMostRecentPastTerm, parseMarkToInt, prepareCoursesForValidation };
+// Calculated lcm(uoc, 6) to determine number of times course must be taken.
+// e.g. 2 UOC course must be taken 3 times, 3 UOC course must be take 2 times
+const getNumTerms = (uoc) => {
+  let num1 = uoc;
+  let num2 = 6;
```
gtangelo: Can we define 6 as a constant var and use that instead for num2 and the return?

gtangelo (commented):
Code looks good apart from those comments.

Few things I wanted to ask:
- Could you test the actions from the context menu on your end? I think deleting a multi term course may crash the app.
- Dragging the second or third instance for a multi term course seems a bit glitchy? I think its because you're setting that as the start index and populating the rest? Would it be better to drag all instance of that course individually rather than forcing the movement for all instance? That or maybe also just dragging it by year instead?
- also, marks, i think its using the same one? Any plans to encapsulate them for their own instance (or if there is a need?)

Peedee2002 (approved):
Damn thats a lot of code ooft

Looks ok from BE side.

frontend/src/pages/CourseSelector/CourseDescription/AddToPlannerButton.jsx
sixohtoo: Replying here because I can't figure out how to reply directly. 

> Would it be better to drag all instance of that course individually rather than forcing the movement for all instance?

I considered doing this but multiterm courses need to be done consecutively (i.e. you can't do ENG2600 in 2020T1, 2020T3 and 2021T1). They are allowed to wraparound to the next year, though. I think it would mislead the user if they could break up a multiterm course, even if there was a warning saying they shouldn't. 

> also, marks, i think its using the same one? Any plans to encapsulate them for their own instance (or if there is a need?)

I don't think there's a need for this. Multiterm courses get given a mark once every instance has been completed (not 2 or 3 seperate marks). Students don't receive a mark for each term so they wouldn't expect to be able to enter a different mark for each term.

frontend/src/reducers/plannerSlice.js
```diff
@@ -103,6 +174,8 @@ const plannerSlice = createSlice({
     },
     // TODO NOTE: think about if you would want to call the backend first to fetch dependant courses
     removeCourse: (state, action) => {
+      // eslint-disable-next-line no-debugger
+      debugger;
```
gtangelo: remove debugger here and eslint ignore

frontend/src/pages/TermPlanner/TermPlanner.jsx
```diff
@@ -73,7 +73,7 @@ const TermPlanner = () => {
   const plannerPic = useRef();
 
   const handleOnDragStart = (courseItem) => {
-    const course = courseItem.draggableId;
+    const course = courseItem.draggableId.slice(0, 8);
```
gtangelo, line 89: Maybe add a comment here that draggableId is in the format of course code + term so its clear that it is slicing the course code

frontend/src/pages/TermPlanner/TermPlanner.jsx
```diff
@@ -82,18 +82,34 @@ const TermPlanner = () => {
...
     setIsDragging(false);
 
-    const { destination, source, draggableId } = result;
+    const { destination, source, draggableId: draggableIdUnique } = result;
+    const draggableId = draggableIdUnique.slice(0, 8);
 
     if (!destination) return; // drag outside container
 
     if (destination.droppableId !== "unplanned") {
+      if (checkIsMultiterm(draggableId, planner.courses)) {
```
gtangelo: nitpick: can we combine the both the if conditions? i.e. `if (checkIsMultiterm(...) && !checkMultitermInBounds(...))`

frontend/src/pages/TermPlanner/TermPlanner.jsx
```diff
@@ -82,18 +82,34 @@ const TermPlanner = () => {
...
+        // check if multiterm course extends below bottom row of term planner
+        if (!checkMultitermInBounds({
+          destRow: destination.droppableId.match(/[0-9]{4}/)[0] - planner.startYear,
+          destTerm: destination.droppableId.match(/T[0-3]/)[0],
+          course: planner.courses[draggableId],
+          isSummerTerm: planner.isSummerEnabled,
+          numYears: planner.numYears,
+        })) {
+          return;
```
gtangelo: Maybe it is worth adding an antd notification popup if a user drags a course outside of bounds? Since it may be a bit confusing on why a course cant be dragged to the last row last column?

gtangelo (approved):
lgtm for code! other than the comments :)

Few things:

I believe locking and unlocking term boxes does not affect multi term courses? Well tbh not entirely sure what's a good way to handle it if you want to go with the automatic dragging approach. Just pointing this out in case if you have any ideas 😅 

Regarding your comment on dragging multi term courses, thats fair enough. However, a note I want to add, when dragging the second instance of a multi term course, would it be better to populate the dest term box from that second instance course? 

i.e For example, lets say a multi term course is planned for 2020T2 2020T3 2021T1. 
- Dragging the course planned for 2020T3 to 2021T3, that course should be planned for 2021T2 2021T3 2022T1. Since we're dragging the second instance of the course, the first and third instance should be in the same position as the original src term box
- Another example: dragging the course 2021T1 to 2021T3, the couse should be planned for 2021T1 2021T2 2021T3.

Right now, I believe it starts populating the multi term courses on the dest term box and onwards?

Would love to hear your thoughts if you want to do this or not! :D 

Either way, I'm happy with the changes.

frontend/src/reducers/plannerSlice.js
sixohtoo: Still can't figure out how to reply properly :(

> I believe locking and unlocking term boxes does not affect multi term courses?

That's a good point. I reckon the best thing to do would be to check if any instance of a multiterm course is in a locked term, then none of them could be moved. It wouldn't make any sense to move 2/3 of the instances while one of them remains behind. I've tried to do this but can't figure it out, so I'll make its own PR for it once I figure out how check if the term is locked.

> would it be better to populate the dest term box from that second instance course?

Great idea! Just implemented this.

backend/data/processors/courses_processing.py
```diff
@@ -157,6 +158,17 @@ def process_enrolment_rules(processed: dict, course: dict):
     processed["raw_requirements"] = re.sub(
         "<br/><br/>", "", course["enrolment_rules"])
 
+def process_multi_term(processed: dict, formatted: dict) -> None:
+    is_multiterm = False
+
+    for attribute in formatted["attributes"]:
```
Peedee2002: this should be an `any()` statement

backend/data/processors/courses_processing.py
```diff
@@ -157,6 +158,17 @@ def process_enrolment_rules(processed: dict, course: dict):
...
+        if attribute["type"] == "multi-term_course":
+            is_multiterm = True
+
+    if int(formatted["UOC"]) % 6 == 0:
+        is_multiterm = False
+    
+    processed["is_multiterm"] = is_multiterm
```
Peedee2002: oneliner:
```python
    processed["is_multiterm"] = any(attribute["type"] == "multi-term_course" for attribute in formatted["attributes"]) and int(formatted["UOC"]) % 6 != 0
```

backend/data/processors/courses_processing.py
```diff
@@ -157,6 +158,17 @@ def process_enrolment_rules(processed: dict, course: dict):
...
         "<br/><br/>", "", course["enrolment_rules"])
 
+def process_multi_term(processed: dict, formatted: dict) -> None:
+    is_multiterm = False
+
+    for attribute in formatted["attributes"]:
+        if attribute["type"] == "multi-term_course":
+            is_multiterm = True
+
+    if int(formatted["UOC"]) % 6 == 0:
```
imagine-hussain: is this `% 6` required here? Are there any courses that are `6 uoc` and multi-term? If so, do we want to exclude them?

Not sure if `0 uoc` courses are in the scope of this card but, if there's multi-term courses that cost `0 uoc`, wouldn't this make their `is_multiterm` false which (i dont think) is what we want

imagine-hussain (approved):
BE LGTM;
Just had a question about the `uoc % 6` but not if that's actually an issue in this context.


Great work!

backend/data/processors/courses_processing.py
Peedee2002: @imagine-hussain the only 0uoc courses are the cringe commerce courses so its ok

sixohtoo: > is this % 6 required here? Are there any courses that are 6 uoc and multi-term? If so, do we want to exclude them?

UNSW has 2 types of multi-term courses. 

There are courses like `ENGG2600` or `ENGG3600` where you have to do the same course multiple times to get credit for it. There are other courses like `ACCT3202 (Industry placement 2)` and `ACCT3203 (Industry placement 3)` which are  a multi-group course, in which you need to complete several different courses to get credit for all of them).

The handbook doesn't actually differentiate between them. The multi-term_course attribute simply says
```
This course is taught across multiple terms or is a member of a group of courses taken across multiple terms. In either case a final grade is withheld until the final course is completed.
```

From looking at the courses, it seems that the UOC of all 'multi-group courses' is divisible by 6 (e.g. `ACCT3202` and `ACCT3203` both have 6 UOC), whereas multi-term courses all have UOCs that aren't divisible by 6 (e.g. 2, 3, 10, etc).

I've excluded multi-group courses from being counted as 'multiterm' because the logic I've applied to multiterm courses like `ENGG2600` isn't applicable for them.

>  if there's multi-term courses that cost 0 uoc, wouldn't this make their is_multiterm false
I'm not fully up to date with 0 UOC courses, but they don't seem to be in coursesFormattedRaw.json, which is what the parser works on. 

If there were some multiterm courses that had 0 UOC, then parts of my logic would need to be re-written because my algorithm would assume that they need to be repeated infinite times.

imagine-hussain: Checks out

PR #556, opened 2022-06-26
Uzaif/cf 243

Uzaif-Sheikh:
Added arrow down btn to scroll to the progression checker courses.
ezgif com-gif-maker (2)


frontend/src/pages/ProgressionChecker/Dashboard/Dashboard.jsx
```diff
@@ -41,6 +57,15 @@ const Dashboard = ({ isLoading, degree }) => {
                 <DegreeCard key={index} concentration={concentration} />
               ))}
           </div>
+          { arrowButton && (
```
gtangelo: lets just display the button always :)

frontend/src/pages/ProgressionChecker/Dashboard/Dashboard.jsx
```diff
@@ -17,6 +19,20 @@ const Dashboard = ({ isLoading, degree }) => {
     config: { tension: 80, friction: 60 },
   });
 
+  window.addEventListener("scroll", () => {
+    if (window.scrollY > 75) {
+      setarrowButton(false);
+    }
+  });
+
+  const clickArrow = () => {
```
gtangelo, line 21: could we use `react-scroller` instead? (similar to the degree wizard) where instead of defining a top for scroll, we just scroll to an `id`? You can define an id value in the main page

gtangelo (changes requested):

frontend/src/pages/ProgressionChecker/Dashboard/Dashboard.jsx
```diff
@@ -1,15 +1,15 @@
...
 import { animated, useSpring } from "@react-spring/web";
-import { Typography } from "antd";
+import { Button, Typography } from "antd";
 import LiquidProgressChart from "components/LiquidProgressChart";
 import DegreeCard from "../DegreeCard";
 import SkeletonDashboard from "./SkeletonDashboard";
 import "./index.less";
 
-const Dashboard = ({ isLoading, degree }) => {
+const Dashboard = ({ isLoading, degree, clickArrow }) => {
```
gtangelo: Is there a reason why `clickArrow` is passed down as a prop rather than being defined within the `Dashboard` component?

gtangelo (approved):
lgtm!

- [x] define `clickArrow` function inside `Dashboard` component

frontend/src/pages/ProgressionChecker/Dashboard/Dashboard.jsx
Uzaif-Sheikh: The reason I passed the ```clickArrow``` function as a prop is I have defined the function in the ```ProgressionChecker.jsx``` to scroll to the ```<Divider id="divider" />``` which is in ```ProgressionChecker.jsx```.

PR #557, opened 2022-06-26
Uzaif/cf 247

Uzaif-Sheikh:
Used styled component for PlannerCart.


frontend/src/components/PlannerCart/CartCourseCard.jsx
```diff
@@ -7,6 +7,7 @@ import {
 } from "antd";
 import { addTab } from "reducers/courseTabsSlice";
 import { removeCourse } from "reducers/plannerSlice";
+// import S from "./styles";
 import "./index.less";
```
gtangelo: Could we replace this to styled component aswell? We would want to completly remove `index.less`

frontend/src/components/PlannerCart/styles.js
```diff
@@ -0,0 +1,70 @@
+import { Button as Btn } from "antd";
+import styled from "styled-components";
+
+const plannerCartRoot = styled.div`
```
gtangelo: Could we capitalise the first letter for the styled components? i.e. `const PlannerCartRoot`

frontend/src/components/PlannerCart/styles.js
```diff
@@ -0,0 +1,70 @@
...
+margin: 10px;
```
gtangelo: identation

frontend/src/components/PlannerCart/styles.js
```diff
@@ -0,0 +1,70 @@
...
+`;
+
+const plannerCartCard = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  padding: 10px;
+  border-radius: 10px;
+
+  :hover {
```
gtangelo: nitpick but could we use `&:hover {` instead?

frontend/src/components/PlannerCart/PlannerCart.jsx
```diff
@@ -48,7 +49,7 @@ const PlannerCart = () => {
         />
       </Tooltip>
       {openMenu && (
-        <div className="planner-cart-menu">
+        <S.plannerCartMenu style={{ backgroundColor: theme === "light" ? "#ffffff" : "#434343" }}>
```
gtangelo: Could we use the `theme` prop within the styled component file instead? Rather than hardcoding the background color here?

gtangelo (changes requested):

frontend/src/config/theme.js
```diff
@@ -34,7 +34,10 @@ export const lightTheme = {
   ...lightBaseColors,
   body: "#ffffff",
   text: "#323739",
-
+  plannerCartCardHover: "#f4f4f4",
```
gtangelo: Nitpick, include this styling within `plannerCartCard` key. So something like the below would be better

```
plannerCartCard: {
    backgroundColorHover: "#f4f4f4",
}
```
This is because we can add future variables under `plannerCartCart` key later on if we need it.

frontend/src/components/PlannerCart/CartCourseCard.jsx
```diff
@@ -31,34 +33,36 @@ const CartCourseCard = ({ code, title, showAlert }) => {
...
-              size="small"
-              shape="circle"
-              icon={<DeleteOutlined />}
-              loading={loading}
-            />
-          </Tooltip>
-        </Popconfirm>
-      </div>
-    </div>
+    <ThemeProvider theme={theme === "light" ? lightTheme : darkTheme}>
```
gtangelo: You don't need to define `ThemeProvider` here? It should be already used in the top level component. Remove `ThemeProvider`

frontend/src/components/PlannerCart/PlannerCart.jsx
```diff
@@ -38,74 +41,77 @@ const PlannerCart = () => {
...
-                selector
-              </Text>
-              <Button
-                type="secondary"
-                shape="round"
-                className="planner-cart-link-to-cs"
-                onClick={() => {
-                  navigate("/course-selector");
-                }}
+    <ThemeProvider theme={theme === "light" ? lightTheme : darkTheme}>
```
gtangelo: Remove `ThemeProvider`. Its already define in the top level component

gtangelo (approved):
lgtm! few things to fix before merging:

- [x] remove `ThemeProvider` component as it is defined already in another top component
- [x] fix `plannerCartCard` theme stylings

PR #560, opened 2022-06-27
Xing/tp/cf 259 when grabbing a course the tern planner shifts by a few pixels

xincczn:
fix(tp): Added coloured border to term boxes upon pick up


gtangelo (approved):
lgtm!

PR #561, opened 2022-06-27
Xing/pc/cf 240 notification icon tooltip should be clicked and redirect the user to the term planner

xincczn:
feat(pc): Redirection via tooltip of non planned courses


gtangelo (approved):
lmaooo this imcluded the deseket button but lgtm

PR #565, opened 2022-06-29
Circles 336 gened endpoint

sonar3d:


joooooooooooooooooooooooooooooooooooosh (approved):

backend/server/routers/programs.py
```diff
@@ -211,4 +212,21 @@ def get_structure(
...
+        200: {
+            "description": "Returns all geneds available to a given to the given code",
+        },
+    },
+)
+
+def getGenEds(
+    programCode: str, search = None):
+    all_geneds = data_helpers.read_data("data/scrapers/genedPureRaw.json")
+    return {"general education courses" : all_geneds[programCode]}
```
gtangelo: Im assuming this is the key if the payload for this endpoint? Really nitpick but can we name this to 'courses' instead? (if that's better? Haha) 😅

backend/server/routers/programs.py
```diff
@@ -211,4 +212,21 @@ def get_structure(
...
+
+@router.get(
+    "/getGenEds/{programCode}",
+    responses={
+        400: {
+            "description": "The given program code could not be found in the database",
+        },
+        200: {
+            "description": "Returns all geneds available to a given to the given code",
+        },
```
Peedee2002, line 245: Would you mind giving a more concrete example here?

backend/server/routers/programs.py
```diff
@@ -211,4 +212,21 @@ def get_structure(
...
+    },
+)
+
+def getGenEds(
+    programCode: str, search = None):
```
Peedee2002: Search doesn't seem to do anything? Is that intended?

backend/server/routers/programs.py
```diff
@@ -211,4 +212,21 @@ def get_structure(
...
 
     
-    return {"structure": structure}
\ No newline at end of file
+    return {"structure": structure}
+
+@router.get(
+    "/getGenEds/{programCode}",
+    responses={
+        400: {
```
Peedee2002, line 222: There should be a response model here. You can use Courses if you take Gabriel's advice.

backend/server/routers/programs.py
sonar3d: ah yes, only didn't cause the course list is so long

backend/server/routers/programs.py
```diff
@@ -211,4 +213,472 @@ def get_structure(
...
+                            "COMP1521",
+                            "COMP1531",
+                            "COMP1911",
+                            "COMP2511",
+                            "COMP2521",
+                            "COMP3511",
+                            "COMP6441",
+                            "COMP6443",
+                            "COMP6445",
+                            "COMP6447",
```
Peedee2002: Ok shit lol that's massive - have a few as an example and cut it off lol

Peedee2002 (approved):
Just that one thing and we should be good

Peedee2002 (approved):
Lgtm

PR #566, opened 2022-06-29
refactor(degree-wizard): Moved less stylesheets over to styled components and adjusted dark mode for page

gtangelo:


frontend/src/config/theme.js
```diff
@@ -58,4 +58,8 @@ export const darkTheme = {
   courseSidebar: {
     menuSubColor: "#262626", // gray-10
   },
+
+  specialsationStep: {
+    background: "#fff",
```
Darianlmj, line 63: Noticed that you're using hex values here but rgb values in other files. Might want to keep everything as the same type for consistency =)

Darianlmj (approved):
looks good to me =)

PR #568, opened 2022-07-01
Cf 272 component deal with subgroup of spec that has a rule and create course list component

leonardo-fan:
- modal for subgroups with gened/rule substring type
- changed grid view to use styled components and removed less
- fixed bug for tableview where courses don't show

https://user-images.githubusercontent.com/90736577/176834699-2b0abeea-0911-4594-8643-04079cfefbcf.mov


frontend/src/pages/ProgressionChecker/GridView/CourseBadge/CourseButton/styles.jsx
```diff
@@ -0,0 +1,26 @@
+import { Button } from "antd";
+import styled from "styled-components";
+
+const InTPCourseButton = styled(Button)`
```
gtangelo: Probably better to rename this to 'CourseButton'?

frontend/src/pages/ProgressionChecker/GridView/CourseBadge/CourseButton/styles.jsx
```diff
@@ -0,0 +1,26 @@
...
+  border-radius: 10px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 271px;
+  height: 80px;
+  white-space: normal;
+`;
+
+const NotInTPCourseButton = styled(InTPCourseButton)`
```
gtangelo: I think this can be combined to the styled component above. Rather than creating two seperate styled components, perhaps u can include the css and toggle between them using styled components props? 

i.e.
```
const CourseButton = styled(Button)`
  border-radius: 10px;
  display: flex;
  align-items: center;
  justify-content: center;
  width: 271px;
  height: 80px;
  white-space: normal;

  ${({planned}) => planned && `
  /* css here */
  `}
`;
```

frontend/src/pages/ProgressionChecker/GridView/CourseBadge/CourseButton/styles.jsx
```diff
@@ -0,0 +1,26 @@
...
+  background: #FFF; 
+  && {
```
gtangelo: Why is this &&?

frontend/src/pages/ProgressionChecker/GridView/GridModal/styles.jsx
```diff
@@ -0,0 +1,47 @@
...
+const { Title } = Typography;
+
+const ModalHeader = styled.div`
+  padding-top: 10px;
+  padding-left: 10px;
+  padding-right: 10px;
+`;
+
+const ModalTitle = styled(Title)`
+  && {
```
gtangelo: what does && do here?

frontend/src/pages/ProgressionChecker/GridView/GridModal/styles.jsx
```diff
@@ -0,0 +1,47 @@
...
+const Instruction = styled.p`
+  margin-top: 10px;
+  margin-bottom: 0px;
+`;
+
+const CourseModal = styled(Modal)`
+  top: 0px;
+  margin: 30px;
+  float: right;
+  .ant-modal-content, .ant-modal-header {
```
gtangelo: after looking at the modal, think its better to remove the border-radius styling 😅

frontend/src/pages/ProgressionChecker/GridView/GridModal/styles.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+const Instruction = styled.p`
+  margin-top: 10px;
+  margin-bottom: 0px;
+`;
+
+const CourseModal = styled(Modal)`
+  top: 0px;
+  margin: 30px;
+  float: right;
```
gtangelo, line 25: Does antd modal do not have a prop to align it to the right (rather than overwriting the css)?

gtangelo (approved):
lgtm other than my comments made! Suprised that you made everything into styled components. Thanks haha 😅 

Though one UI thing, could we create/use this component for the course list (with the add/remove to planner functionality)?
image

I'm happy for this to be merged without the change above tho (since we want to release the progression checker soon and probs u want to start on the prereq tree). If you want to do this change later on, create a ticket for this and put it in the backlog :)

frontend/src/pages/ProgressionChecker/GridView/GridModal/styles.jsx
leonardo-fan: only for centre alignment, otherwise they say to change the style

frontend/src/pages/ProgressionChecker/GridView/GridModal/styles.jsx
leonardo-fan: ran into some trouble from antd so i read this https://github.com/styled-components/styled-components/issues/1816

seems like !important does the trick anyway - so will change it :)

frontend/src/pages/ProgressionChecker/GridView/CourseBadge/CourseButton/styles.jsx
leonardo-fan: same as below

frontend/src/pages/ProgressionChecker/GridView/CourseBadge/CourseButton/styles.jsx
leonardo-fan: oo nice

PR #569, opened 2022-07-01
Chore/run app python3 support

imagine-hussain:
- use load_env and os.getenv for getting backend env variables in `run_app` - now order of the env doesn't matter
- add `PYTHON_VERSION` as a viable envirionment variable (by default, `python`) to allow devs to specify what python version they want to use to run their app.
- Add clarification about python env variable to backend readme

### DIFF
--- a/run_app.py
+++ b/run_app.py
@@ -1,11 +1,13 @@
 # pylint: disable=cyclic-import
 """ run all of circles in one terminal """
 import logging
-from subprocess import Popen, check_call
-from dotenv import load_dotenv
-import threading
 import os
 import sys
+import threading
+from subprocess import Popen, check_call
+
+from dotenv import load_dotenv
+
 
 class LogPipe(threading.Thread):
     """ boilerplate abstraction for redirecting the logs of a process """
--- a/run_app.py
+++ b/run_app.py
@@ -2,6 +2,7 @@
 """ run all of circles in one terminal """
 import logging
 from subprocess import Popen, check_call
+from dotenv import load_dotenv
 import threading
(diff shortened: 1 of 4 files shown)


run_app.py
```diff
@@ -40,14 +43,15 @@ def write(self, message):
...
-    data[1] = data[1].replace('\n', '')
-    return (data[0], data[1])
+    """
+        reads backend.env for mongodb username and password and python
+        version.
+    """
+    load_dotenv("./env/backend.env")
+    username = os.getenv("MONGODB_USERNAME")
+    password = os.getenv("MONGODB_USERNAME")
+    python = os.getenv("PYTHON_VERSION") or "python"
```
Peedee2002, line 53: Wait i didn't know python did this
~~im gonna abuse this feature to hell now~~

run_app.py
```diff
@@ -1,10 +1,13 @@
...
 """ run all of circles in one terminal """
 import logging
-from subprocess import Popen, check_call
-import threading
 import os
 import sys
+import threading
+from subprocess import Popen, check_call
+
+from dotenv import load_dotenv
```
Peedee2002, line 9: Where is this used?

imagine-hussain: the threading and subprocess was already there; the `load_dotenv` is used to load up variables from `backend.env` as actual environment variables. very nice for not having to load up the file urself and write ur own parsing

Peedee2002: Nevermind oops

Peedee2002 (approved):

PR #582, opened 2022-07-08
feat(frontend): Removed craco and used Vite

gtangelo:
Changes made
- removed craco and used vite instead
- improved load times when starting it up coz of vite 😎


kellieyau (approved):
is the dark mode in the .less file deleted because we're changing to style components?

Tho integration looks good and so does clean up

frontend/package.json
```diff
@@ -6,26 +6,23 @@
...
     "@react-spring/web": "9.4.5",
     "@reduxjs/toolkit": "1.8.3",
     "@tippyjs/react": "4.2.6",
+    "@vitejs/plugin-react-refresh": "^1.3.6",
     "antd": "4.21.5",
     "axios": "0.27.2",
-    "craco-less": "1.20.0",
     "framer-motion": "6.4.2",
     "moment": "2.29.3",
     "react": "17.0.2",
```
Peedee2002, line 17: We should attempt to update this as well as the other dependencies we've blocked

gtangelo: I realised this when removing cracro but the issue with updating to react 18 is that `react-beautiful-dnd` (our drag and drop library) only officially supports upto react 17. And it seems based on their github that its not currently being maintained by atlassian at the moment 😢. We can try using legacy peer deps but that seems such a workaround and may have some unknown bugs due to the version differences

gtangelo:
> is the dark mode in the .less file deleted because we're changing to style components?

Yep exactly haha

frontend/package.json
Peedee2002: Smh @SpanishPear plz fix

(Jk)

spanishpear: [This](https://github.com/atlassian/react-beautiful-dnd/issues/2399) issue suggest that it does work in react 18, just not in strict mode StrictMode

Xhale1: Hello! I just released a fork of `react-beautiful-dnd` which should fully support react 18 and react strict mode.

Just replace `react-beautiful-dnd` with `@hello-pangea/dnd` and you should be good to go 👍

PR #586, opened 2022-07-09
feat: tested lots of specialisations (especially engineering) and fixed them

Peedee2002:


frontend/package.json
```diff
@@ -9,11 +9,11 @@
     "@react-spring/web": "9.4.5",
```
Peedee2002, line 9: I am salty that we still have vulnerabilities in our repo (i think from ant-design/charts)

@gtangelo is it worth fixing?

frontend/src/pages/CourseSelector/CourseBanner/CourseBanner.jsx
```diff
@@ -14,15 +14,15 @@ const CourseBanner = () => {
     programCode, programName,
   } = useSelector((state) => state.degree);
 
-  const { isLockedEnabled } = useSelector((state) => state.courses);
+  const { showLockedCourses } = useSelector((state) => state.settings);
```
gtangelo, line 17: cheers for picking this up :D

frontend/src/pages/DegreeWizard/DegreeStep/DegreeStep.jsx
```diff
@@ -42,14 +42,10 @@ const DegreeStep = ({ incrementStep }) => {
...
-      const fullDegreeName = Object.keys(allDegrees).map((code) => `${code} ${allDegrees[code]}`);
-      setOptions(
-        fullDegreeName.filter((degree) => degree.toLowerCase().includes(newInput.toLowerCase())),
-      );
-    } else {
-      setOptions([]);
-    }
+    const fullDegreeName = Object.keys(allDegrees).map((code) => `${code} ${allDegrees[code]}`);
+    setOptions(
+      fullDegreeName.filter((degree) => degree.toLowerCase().includes(newInput.toLowerCase())).splice(0, 8),
```
gtangelo, line 47: Think I already know the answer but why is there an added `splice(0,8)` here? Is this here to restrict the options once we add more degrees?

frontend/package.json
gtangelo: Hmm right now it shouldn't be a big issue right? If it worst down to it, we could probably remove `ant-design/charts` since there's only one component that relies on it as of this moment.

gtangelo (approved):

frontend/src/pages/DegreeWizard/DegreeStep/DegreeStep.jsx
Peedee2002: @gtangelo yup

PR #591, opened 2022-07-13
feat(progression-checker): Add disclaimer and minor improvements with UOC and showing courses

gtangelo:
Changes made:
- Add disclaimer for accuracy
- Made uoc count for dashboard reflect total courses in planner rather than the group containers
- Add collapsible component to groups
- Show courses in planner in full mode for hasLotsOfCourses

image
image


leonardo-fan (approved):
lgtm :)

PR #612, opened 2022-07-16
feat(route)/circles-351/get: `programs/getStructureCourseList/{programCode/{spec}/`

imagine-hussain:
##  Current Changes

- FE currently does the effort of taking the course structure and picking out all the courses from within the specialisations and other containers
- This responsibility has been moved to the BE so that there is just a single call for the courseList in a structure given a program name and specialisations in the same way `getStructure` works.
- retval of this call is less bytes than getstructure -> faster

## Upcoming

This is the first step to moving the computation of the graph to the BE. Next step is to move the following line to be BE responsibility - effectively making this endpoint redundant too.

```js
const res = await Promise.all(
   courseList.map(
    (c) => axios.get(`/courses/getPathFrom/${c}`).catch((e) => e)
  )
);
```


frontend/src/pages/GraphicalSelector/GraphicalSelector.jsx
```diff
@@ -76,17 +76,14 @@ const GraphicalSelector = () => {
...
-    const courseList = (
-      Object.values(structure)
-        .flatMap((specialisation) => Object.values(specialisation)
-          .filter((spec) => typeof spec === "object" && spec.courses && !spec.type.includes("rule"))
-          .flatMap((spec) => Object.keys(spec.courses)))
-        .filter((v, i, a) => a.indexOf(v) === i) // TODO: hack to make courseList unique
-    );
+
+    // TODO: Move this to the backend too
+    // should be a universal /programs/getGraphEdges/{programCode}/{specs}
```
gtangelo, line 112: I'm a bit confused with the approach here (not neccesarly disagreeing it tho). Were you planning on updating the `/programs/getStructureCourseList/` endpoint to become `/programs/getGraphEdges/`? So, `/programs/getGraphEdges/` would handle both the logic of:

```
const courseList = (
      Object.values(structure)
        .flatMap((specialisation) => Object.values(specialisation)
          .filter((spec) => typeof spec === "object" && spec.courses && !spec.type.includes("rule"))
          .flatMap((spec) => Object.keys(spec.courses)))
        .filter((v, i, a) => a.indexOf(v) === i) // TODO: hack to make courseList unique
    );
```
and
```
const res = await Promise.all(courseList.map((c) => axios.get(`/courses/getPathFrom/${c}`).catch((e) => e)));
```
?

I was initially thinking without this line that the approach you're approach taken would the BE return a course list from the endpoint. And have another endpoint just linking the edges, given a course list? But now I'm confuse haha.

imagine-hussain: In my mind, the eventual goal was to have a route
```
/programs/getGraph/{programCode}/{spec}
```
This would
- construct the course structure
- flatten the course structure to get a list of courses
- find the edges corresponding to the course list
 
```py
return {
  "courses": List[str],
  "edges": List[Tuple(str, str)]
}
```

My rationale was that the FE would need the `course` list anyways since otherwise courses with no edges get skipped but, I wanted to avoid the FE needing to send back a course list to get the edges.

The most optimal thing I could think of was to have everything sent back in a single call.

backend/server/routers/programs.py
```diff
@@ -195,28 +195,27 @@ def add_specialisation(structure: dict, code: str):
...
-    structure['General'] = {}
-    structure['Rules'] = {}
-    with suppress(KeyError):
-        for container in programsResult['components']['non_spec_data']:
-            add_subgroup_container(structure, "General", container, [])
-    apply_manual_fixes(structure, programCode)
-
+    structure = add_specialisations({}, spec)
+    uoc = add_program_code_details(structure, programCode)
+    course_list_from_structure(structure)
```
Peedee2002: we must still apply manual fixes here!

Peedee2002 (changes requested):
small fix needed here

backend/server/routers/programs.py
imagine-hussain: The add program code details function applies manual fixes

backend/server/routers/programs.py
```diff
@@ -240,17 +239,87 @@ def get_structure(
...
+        """ Recursively search for courses in a structure """
+        print(structure.keys())
+        if not isinstance(structure, (list, dict)):
+            return
+        # For a list, recurse on all its object
+        if not isinstance(structure, dict):
+            return
+        for k, v in structure.items():
+            with suppress(KeyError):
+                if not isinstance(v, dict) or "rule" in v["type"]:
```
gtangelo, line 303: Could we also exclude gened courses (if it doesn't already)?

imagine-hussain: the constructor for the `structure` doesn't attach geneds so this route won't return geneds

imagine-hussain: > Could we also exclude gened courses (if it doesn't already)?

the constructor for the structure doesn't attach geneds so this route won't return geneds

If in a future use-case, geneds need to be excluded before we get a course-list, imo its preferable to have a function to remove the geneds before they get sent to this function

Peedee2002 (approved):
unfortunately three is a bit of pain to go, but it looks good from a code perspective. Just need to get getStructure working 😅

PR #613, opened 2022-07-16
feat(backend): Add gened courses support to getStructure endpoint

gtangelo:
FE wants to show gen ed courses on the progression checker, and the simple way to do this is to add the gen ed courses as part of the `getStructure` endpoint.

Changes made:
- Gen ed scrapper now returns course title alongside course code
- gen ed courses gets appended to the `getStructure` endpoint
- suppress gen eds from showing up on the course selector sidebar on the FE

### DIFF
--- a/backend/server/routers/model.py
+++ b/backend/server/routers/model.py
@@ -103,7 +103,7 @@ class Courses(BaseModel):
     courses: list[str]
 
 class GenEdCourses(BaseModel):
-    courses: dict
+    courses: dict[str, str] = {}
 
 class CoursesPath (BaseModel):
     original: str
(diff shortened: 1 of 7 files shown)


backend/server/routers/model.py
```diff
@@ -99,9 +99,11 @@ class PlannerData(BaseModel):
     mostRecentPastTerm: dict
 
 
-class Courses (BaseModel):
+class Courses(BaseModel):
     courses: list[str]
 
+class GenEdCourses(BaseModel):
+    courses: dict[str, str] = {}
```
imagine-hussain: Is this model here to just differentiate between courses being returned as just a list of coursecodes vs dict with code and title?
If so, it might be good to decouple this from geneds specifically and just have a general type where the `dict` option is actually `Courses` and the list option is something like `CourseList` or `CourseCodes`.

imagine-hussain (approved):
I'm a little conflicted on the naming for the different models but... I'm also not 100% sure on what would work better.

Otherwise, lgtm! feel free to merge

PR #621, opened 2022-07-20
fix(courses): fix silly mistake

Peedee2002:


Peedee2002:
@gtangelo this fixes the 500 error you were getting 🤒

imagine-hussain (approved):

PR #629, opened 2022-07-23
Circles 337 add static typecheck

Peedee2002:
Cool, this is now pretty cool! I am convinced that this is a nice addition. 

The only problems that lie here are:
1. my annotations for models.py in data might be wrong! please @hdphuong let me know if i am stupid.
2. i messed with how getStructure returns, since i added a "content" field to seperate it from "name" This nessesitated light changes in the FE, which is why i added @gtangelo


backend/algorithms/tests/test_conditions.py
```diff
@@ -87,51 +87,6 @@ def test_composite_condition_course():
     })
     assert (and_or_cond.validate(user))[0]
 
-
-def test_composite_condition_course():
```
Peedee2002, line 91: random duplicate test?

backend/algorithms/tests/test_conditions.py
```diff
@@ -87,51 +87,6 @@ def test_composite_condition_course():
...
-    '''AND/OR conditions with only course requirements'''
-    user = User(USERS["user3"])
-
```
joooooooooooooooooooooooooooooooooooosh, line 94: Were all these tests removed because of typing issues or something else?

joooooooooooooooooooooooooooooooooooosh:
It seems fine to me - can you elaborate on when/why it weighs us down?

Peedee2002:
> It seems fine to me - can you elaborate on when/why it weighs us down?

@joooooooooooooooooooooooooooooooooooosh some of the error checking is quite redundant, and in some cases just wrong (eg some of the stuff where i said #type: ignore)

backend/server/routers/programs.py
```diff
@@ -78,7 +78,7 @@ def add_subgroup_container(structure: dict, type: str, container: dict, exceptio
     title = container.get("title")
     if container.get("type") == "gened":
         title = "General Education"
-    if container.get("type") is not None and "rule" in container.get("type"):
+    if container.get("type") is not None and "rule" in container.get("type"): # type: ignore
```
imagine-hussain: Why does this need an ignore.
i.e - why does it think this type is and what do we actaully want it to be

Peedee2002: Yea this is ignoring the case where it is None. For some reason, mypy doesn't understand short circuit

imagine-hussain: hmmm, sounds like a consequence of dictionary w/out strongly defined types and classes (OOP and its consequence :( )
To avoid the `#type: ignore`, we could possibly just do this with a `supress(KeyError)`

```py
type = "whatever_default"
with suppress(KeyError):
  if "rule" in container.get("type"):
    type = "Rule"
```

But just leaving this as is, is fine too tbh

spanishpear: N.B: I haven't read the codebase except the highlighted section

but, if you're using mypy (or any typechecker) you should almost never have #ignores - unless there is literally a bug in the type system, or no reasonable workaround.

Have you investigated having `container` being a `TypedDict` https://mypy.readthedocs.io/en/stable/more_types.html#typeddict

spanishpear: Aside: walrus is fun here if you want to minimise the number of indexes you do
```suggestion
if (type := container.get("type")) is not None and "rule" in type:
```

Peedee2002: ooo this was a very nice rabbithole, which helps heaps with accessing the database. I typedicted the database results now, so should be cool 👀

frontend/src/pages/ProgressionChecker/ProgressionChecker.jsx
```diff
@@ -109,12 +109,11 @@ const ProgressionChecker = () => {
     };
 
     // Example subgroup: Core Courses, Computing Electives
-    Object.keys(structure[group]).forEach((subgroup) => {
+    Object.keys(structure[group].content).forEach((subgroup) => {
       // Do not include if field is not an object i.e. 'name' field
```
gtangelo: this comment should not be relevant anymore, if so can we remove it?

frontend/src/pages/CourseSelector/CourseSidebar/CourseSidebar.jsx
```diff
@@ -45,12 +45,11 @@ const CourseSidebar = ({ structure, showLockedCourses }) => {
       newCoursesUnits[group] = {};
 
       // Example subgroup: Core Courses, Computing Electives
-      Object.keys(structure[group]).forEach((subgroup) => {
+      Object.keys(structure[group].content).forEach((subgroup) => {
         // Do not include if field is not an object i.e. 'name' field
```
gtangelo: we no longer need this check right?

backend/server/routers/model.py
```diff
@@ -39,8 +40,19 @@ class CourseDetails(BaseModel):
...
+    courses: NotRequired[dict[str, str | list[str]]]
+    type: NotRequired[str]
+
+class Container(TypedDict):
+    name: str
+    content: dict[str, ContainerContent]
+
+StructureType = dict[str, Container]
+
 class Structure(BaseModel):  # this is a copout - we should avoid this
```
imagine-hussain: can remove the `this is a copout` comment now?

gtangelo (approved):
lgtm for fe changes

backend/server/routers/model.py
Peedee2002: hmm, i thought i did

Peedee2002: hmm ok gh is wierd, after i pushed again my diff was corrected lol

imagine-hussain (approved):

backend/data/processors/models.py
```diff
@@ -0,0 +1,71 @@
...
+
+class SpecData(TypedDict):
+    is_optional: bool
+    specs: dict[str, str]
+    notes: str
+
+class SpecsData(TypedDict):
+    honours: Optional[dict[str, SpecData]]
+    minors: Optional[dict[str, SpecData]]
+    majors: Optional[dict[str, SpecData]]
```
hdphuong: Shouldn't these be NotRequired?

backend/data/processors/models.py
```diff
@@ -0,0 +1,71 @@
...
+    spec_data: SpecsData
+
+class Program(TypedDict):
+    title: str
+    code: str
+    duration: Optional[int]
+    UOC: int
+    faculty: str
+    overview: str
+    Structure_summary: str
```
hdphuong: `structure_summary` (with small s) here

backend/data/processors/models.py
```diff
@@ -0,0 +1,71 @@
...
+    level: int
+    description: str
+    study_level: Literal["Undergraduate"] | Literal["Postgraduate"]
+    school: str
+    faculty: str
+    campus: Literal["Sydney"] | Literal["UNSW Canberra"] | Literal["Paddington"] | Literal[""]
+    equivalents: dict[str, Literal[1]]
+    exclusions: dict[str, Literal[1]]
+    terms: list[Literal["T1"] | Literal["T2"] | Literal["T3"] | Literal["T0"]]
+    gen_ed: Literal[1] | Literal[0]
```
hdphuong: probs a wontfix but can this be `bool` for consistency?

backend/server/routers/model.py
```diff
@@ -28,19 +28,28 @@ class CourseDetails(BaseModel):
     study_level: str
     school: Optional[str]
     campus: str
-    equivalents: dict
+    equivalents: dict[str, str]
```
hdphuong, line 31: should be `dict[str, Literal[1]]`

hdphuong (approved):
lgtm! just some trivial nitpicking 😃

backend/server/routers/model.py
Peedee2002: I think not - it is the name and code of the course
this is checked at runtime by pydantic soooo... should be ok if the tests pass lol

PR #633, opened 2022-07-26
feat(graphical-selector): toggle on/off nodes and edges for unlocked

kellieyau:


frontend/src/pages/GraphicalSelector/GraphicalSelector.jsx
```diff
@@ -148,13 +153,63 @@ const GraphicalSelector = () => {
...
 
+  const isUnlockedEdge = (courses, node1, node2) => (
+    (Object.values(courses).indexOf(node1) > -1) && (Object.values(courses).indexOf(node2) > -1)
+  );
+  const isUnlockedNode = (courses, node) => (Object.values(courses).indexOf(node) > -1);
+
+  const showUnlockedGraphOnly = (courses) => {
+    const nodes = graph.getNodes();
+    const edges = graph.getEdges();
+    nodes.forEach((n) => (isUnlockedNode(courses, Object.values(n)[0].id) ? n.show() : n.hide()));
```
gtangelo: What is at `Object.values(n)[0]`? May not be that readable to index at 0 so could be a better way to access this variable?

gtangelo (changes requested):
lgtm aside from that comment! Though I'm not sure why there are some of uzaif's changes and the modal implementation ahahah. Could we clean this up? 😅

frontend/src/pages/GraphicalSelector/GraphicalSelector.jsx
kellieyau: oh so basically i have to get the values of each node which contains all their info, and id is onlt located in 0 as such:
image
by doing Obejct.values(n) we get:
image
not sure of any way around that

kellieyau: Ideally I would try to access ._cfg / [_cfg] but that doesnt work

Darianlmj: Oh yh I get what you mean. I think you can access it via `['_cfg']` instead over `[_cfg]`. If that doesn't work, try:

```
nodes.forEach((n) => {
  const { _cfg } = n;
  return isUnlockedNode(courses, _cfg.id) ? n.show() : n.hide()
});
```
Though it is not a one liner hahaha.

I'm pretty sure the above two should work but lmk if it doesn't 😄. Would prefer doing this way since I'm not totally sure if Object.values returns values in order i.e. at index 0, could return `false` over the object for some browsers (side objects are not ordered I think?).

Written by Gabriel from Darian's account since I don't have my laptop and mobile GitHub does not work well for PR comments xD

kellieyau: haha yep all g! should be using that instead

gtangelo (approved):

PR #639, opened 2022-07-29
Elliot/circles 352

sixohtoo:
Rewritten condition objects str method to return jsonified version of prereq tree.

Below is an example of a create_condition argument converted into json

```["(", "(", "COMP1511", "||", "COMP1521", ")", "&&", "(", "MATH1141", "&&", "MATH1151", ")", "&&", "(", "COMP2041", "&&", "COMP1531", ")", ")"]```

```
{
	'children': [{
			'children': [{
					'children': ['CourseCondition(COMP1511)']
				},
				{
					'children': ['CourseCondition(COMP1521)']
				}
			],
			'id': 'subtree.0',
			'logic': '||'
		},
		{
			'children': [{
					'children': ['CourseCondition(MATH1141)']
				},
				{
					'children': ['CourseCondition(MATH1151)']
				}
			],
			'id': 'subtree.1',
			'logic': '&&'
		},
		{
			'children': [{
					'children': ['CourseCondition(COMP2041)']
				},
				{
					'children': ['CourseCondition(COMP1531)']
				}
			],
			'id': 'subtree.2',
			'logic': '&&'
		}
	],
	'id': 'root',
	'logic': '&&'
}
````

### CODE REVIEW

File: backend/algorithms/objects/conditions.py [hunk 1]
```diff
@@ -126,9 +135,12 @@ def beneficial(self, user: User, course: dict[str, Tuple[int, int | None]]) -> b
             return False
         return any(c in course.keys() for c in self.courses)
 
-
     def __str__(self) -> str:
-        return f"CoreqCoursesCondition(courses={self.courses}, logic={self.logic})"
+        logic = "and" if self.logic == Logic.AND else "or"
+        return json.dumps({
+            'logic': logic,
+            'children': [f'CoreqCoursesCondition({course})' for course in self.courses],
```
leonardo-fan: could we also use something similar to line 88 here? i.e. ```{ 'id': course }```

leonardo-fan (commented):
lgtm!! know we had a discussion just before this - but just had an extra comment hehe

backend/algorithms/objects/conditions.py
sixohtoo: So children is an array of dictionaries with an id field? Or change children to 'id' which is an array of course code strings.

leonardo-fan: Yeah an array of dictionaries with an id field would be good

backend/algorithms/objects/conditions.py
```diff
@@ -305,7 +330,9 @@ def is_path_to(self, course: str) -> bool:
         return False
 
     def __str__(self) -> str:
-        return f"SpecialisationCondition: {self.specialisation}"
+        return json.dumps({
+            'Specialisation': self.specialisation,
```
Peedee2002: wierd inconsistancy in case?

Peedee2002 (approved):
besides that case, looks good to me :)

PR #642, opened 2022-08-12
Hotfix/gened rule key error

gtangelo:
Added back courses related to geneds and subsequently rules to the get structure api because python key errors goes brr

Before:
Screen Shot 2022-08-12 at 10 36 57 pm

After:
Screen Shot 2022-08-12 at 10 43 48 pm


imagine-hussain (approved):

Peedee2002 (approved):
I'm big clown, oops

PR #643, opened 2022-08-13
perf(cache): optimise graph edges

imagine-hussain:
# graph go vroom vrooom

## Changes

- Caching the graph and then loading the entire thing in memory on the server.
- Building graph should now take ~2-3 orders of magnitude less time because just formatting a lookup to pre-computed list rather looping over every course in the handbook, for every course in the selector.

## Total Performance Diff
- Bottleneck on the graph call is now probably the creation of the `courseList` from `structure`.
- Performance ChangeLog
  - Old (before any refactor): 9 Seconds inital, ~5-6 seconds on following renders
  - Move to BE: still ~7 sec initial but ~4-5 sec on consecutive refreshes
  - Now:
     - sub-second load times for graph component :pog
    

https://user-images.githubusercontent.com/93496985/184472090-b7db5aeb-a07b-4517-8782-0a8efd94d63a.mov


backend/data/processors/cache_graph.py
```diff
@@ -0,0 +1,70 @@
+"""
+Constructs the course graph using the conditions objects and jumps it as a json file.
+
+From `/backend`, run this as:
+    `python3 -m data.processors.cache_graph`
+
+Do NOT run this from it's current location.
+
+TODO: This should be moved as a command that can be run by `run_processors.py`
```
Peedee2002: Doing this is very easy, add a function as an entry into the `run` dictionary

backend/data/processors/load_conditions.py
```diff
@@ -16,21 +19,30 @@ def cache_conditions_pkl_file():
...
-    all_conditions_tokens_file = "./data/final_data/conditionsTokens.json"
-    pickle_file = "./data/final_data/conditions.pkl"
+    # Dump the dictionary as a pickle file
+    with open(CONDITIONS_PICKLE_FILE, "wb") as f_out:
+        pickle.dump(all_objects, f_out, pickle.HIGHEST_PROTOCOL)
 
-    with open(all_conditions_tokens_file, "r", encoding="utf8") as f:
-        all_conditions_tokens = json.load(f)
+def construct_conditions_objects() -> Dict[str, CompositeCondition | None]:
+    """
```
Peedee2002, line 15: Define this function before the one above?
I know that it doesn't actually matter, but it's confusing to have your first function in your file dump stufd

backend/server/routers/courses.py
```diff
@@ -3,10 +3,10 @@
 """
 from contextlib import suppress
 import re
-from typing import Dict, List, Mapping, Set, Tuple
+from typing import Any, Dict, List, Mapping, Set, Tuple
```
Peedee2002: You aren't using Any

backend/server/routers/courses.py
```diff
@@ -397,15 +399,10 @@ def get_path_from(course: str) -> dict[str, str | list[str]]:
     fetches courses which can be used to satisfy 'course'
     eg 2521 -> 1511
     """
-    course_condition = CONDITIONS.get(course)
-    if not course_condition:
-        raise HTTPException(400, f"no course by name {course}")
+    out: List[str] = list(INCOMING_ADJACENCY.get(course, []))
```
Peedee2002, line 402: This is not a list[str] yet. It might be None since that's the return type of .get().
That's probably why mypy is screaming at you. 
If the course is not found, you will get a list(None) which I think is a type error (?)

Peedee2002: You should add a HTTP exception in the case it is None.

Peedee2002 (changes requested):
A few small changes to make. I didn't check the pipeline yet so I only guessed at the mypy error. I'll read it and give you another comment on what it might be if I am wrong.

Peedee2002:
OOOOH you need to delete backend/__init__.py. that will help mypy not die. You only need __init__.py for modules. The top level directory is obviously not a module.
You should download a mypy extension for vim so you get feedback for this stuff when working.

backend/server/routers/courses.py
Peedee2002: Actually now that I think about it, nvm this makes sense, I'm wrong.

imagine-hussain:
> OOOOH you need to delete backend/**init**.py. that will help mypy not die. You only need **init**.py for modules. The top level directory is obviously not a module. You should download a mypy extension for vim so you get feedback for this stuff when working.

OHHHH i think this would be it. I funny moment where i think some other mypy error at the very start of the PR told me to add a `__init__.py` here, probably because I had some really messy filepath stuff being imported and then this lingered around

backend/data/processors/load_conditions.py
imagine-hussain: 👍

Peedee2002 (approved):
ive added the code for running it from runprocessors.py

imagine-hussain:
> ive added the code for running it from runprocessors.py

ty 🥺

PR #650, opened 2022-08-15
chore(frontend): reduce initial load bundle size

Peedee2002:
i changed all the imports to directly reach for their dependancy, which reduces build time. Then, i also lazy-loaded the graph libraries to reduce package size to 3/4ths of what it was.


lhjt:
This is good, but we are still at ~900KB initial JS load. We should aim to get around ~500KB maximum hopefully at some point; should take a look at https://rollupjs.org/guide/en/#outputmanualchunk

Peedee2002:
> This is good, but we are still at ~900KB initial JS load. We should aim to get around ~500KB maximum hopefully at some point; should take a look at https://rollupjs.org/guide/en/#outputmanualchunk

Hmmm now that I think about it, liquid chart and radial chart can probs be lazy loaded too

lhjt:
image
This can probably extracted out into its own module and lazy loaded as well (only used to export).

frontend/src/components/EditMarkModal/EditMarkModal.tsx
```diff
@@ -1,10 +1,10 @@
 import type { Dispatch, SetStateAction } from 'react';
 import React, { useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
-import {
-  Button, Input, message,
-  Modal,
-} from 'antd';
+import Button from 'antd/lib/button';
```
gtangelo, line 4: I know importing the component directly is our only option to minimise build sizes but it still weird to import it like this rippp

lhjt: Actually - I'm not sure if this is the case since tree shaking _should_ be a thing. I'll have to test it with the previous way of importing and check out if the tree shaking is actually in effect.

lhjt:
Superseded by #673.

PR #671, opened 2022-08-16
chore(frontend): reduce initial load bundle size

Peedee2002:
Done messing with this, i promise. 
Its basically using suspense to lazy-load some libraries. We should be more sensitive of bundle sizes going forward.
the bundle size is now under 500kb.
a possible todo here is to make some proper skeletons for suspense, but it doesnt really matter too much because usually the backend network call is significantly slower.

### DIFF
--- a/frontend/.eslintignore
+++ b/frontend/.eslintignore
@@ -1,0 +1,2 @@
+vite.config.ts
+vitest.config.ts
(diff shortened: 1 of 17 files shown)


lhjt (approved):
~~Looks good; we'll need to ensure we have some better fallbacks later but it should be okay for the time being.~~

PR #674, opened 2022-08-17
Economics program

sixohtoo:


Peedee2002:
you also need to add it to the programs.py list, and it is possible that there are some courses in economics that are unsupported. More importantly though, you need to add the double degrees, and make a manual fix (in manual_fixes.py) to support the changes that double degrees make.

Peedee2002:
also, you need to rerun the processors.

Peedee2002:
the final way to check that you were successful in this is that the whole workflow should be functional on the FE. You should test it manually for now, and but we will be working on automating this.

PR #677, opened 2022-08-17
Feat/circles 362/terms available route

imagine-hussain:
## Route
- Add routes `courses/termsOffered/{course}` and `courses/legacyTermsOffered/{year}/{course}`
- Both will return list of courses that are offered. Used same as `/getCourse` and the corresponding legacy route.

## Bug (maybe)
- Some instances of `/courses/getCourse` were failing because the returned object didn't have a `multiterm` attribute as required by the response model. Moved this as an optional setting for now since if it's not `true` as it shouldn't affect the courses that are *actually* multi-term right now. Not sure if just a matter of re-running processors to update course info or, adding in a clause to the processors to add in that multi-term aspect. Worth looking at in future card.

### DIFF
--- a/backend/server/routers/courses.py
+++ b/backend/server/routers/courses.py
@@ -446,8 +446,26 @@ def courses_unlocked_when_taken(userData: UserData, courseToBeTaken: str) -> Dic
     }
 
 
-@router.get("/termsOffered/{course}", response_model=TermsList)
-def terms_offered(course: str) -> List[str]:
+@router.get(
+    "/termsOffered/{course}",
+    response_model=TermsList,
+    responses={
+        400: {
+            "description": "The given course code could not be found in the database",
+        },
+        200: {
+            "description": "Returns the terms a course is offered in",
+            "content": {
+                "application/json": {
+                    "example": {
+                        "terms": ["21T2", "21T2", "21T3"],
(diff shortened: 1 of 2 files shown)


backend/server/routers/courses.py
```diff
@@ -444,6 +445,70 @@ def courses_unlocked_when_taken(userData: UserData, courseToBeTaken: str) -> Dic
...
+                "application/json": {
+                    "example": {
+                        "terms": ["21T2", "21T2", "21T3"],
+                    },
+                },
+            },
+        },
+    }
+)
+def terms_offered(course: str) -> Dict[str, List[str]]:
```
Peedee2002, line 468: We need to ask for the years they want, and give them the terms **for every year accurately**.
For example, 2041 was offered in 22t1. We need to help the frontend know this.

An example would be:
TermsOffered(course, sYear, lYear):
  Return {
     2020: [t1, T2],
     2021: [T3]
  }

Peedee2002 (changes requested):

backend/server/routers/courses.py
imagine-hussain: realised I misread the spec the moment I opened this :/

PR #686, opened 2022-08-22
feat: added and display notes for rules in the program structure

gtangelo:
Retuned new field `notes` for a subgroup from the BE getStructure endpoint (spefically for rules) for the progression checker. Otherwise, there isnt enough information to tell the user if it is just the rule title that is being displayed.

Screen Shot 2022-08-22 at 11 44 50 am


Peedee2002:
We are actually looking to enforce these rules at some point. This may need some UI support.

Peedee2002 (approved):

Peedee2002:
oop @sixohtoo looks like your second card doesnt need to be made lol

PR #689, opened 2022-08-22
feat/circles 362/terms available route

imagine-hussain:
## Route
- Add routes `courses/legacyTermsOffered/{course}/{years}`
- returns a two items
  - `terms` a dict indexed by keys, where values is a list of terms that the course ran in, in that given year
  - `years` a list of tuples where the first entry is the `year` and the second entry is the exception that caused that year to fail returning a result
- note that `years` is expected to be a list of years, connected by `+`. e.g `"2021+2022`

## tests (i'm as surprised at their existence as u are)
- doesn't test LIVE_YEAR
- (relatively) safe to test COMP1511 from legacy years as a sanity check
- also checks that the `fails` are being returned properly when the course or years are cringe

## abt multi-term....
- i caught it *occasionally* resulting in a 500 error because the model fails to validate as it's not a field that exists in some of the data base so `model.py` get's very sad
- quick fix is just to make the thingy `Optional` since null values are falsey anyways


imagine-hussain:
if you're seeing this and the tests are still failing, it is my tests that are wrong, not my code

Peedee2002 (approved):

backend/server/routers/courses.py
```diff
@@ -445,6 +446,59 @@ def courses_unlocked_when_taken(userData: UserData, courseToBeTaken: str) -> Dic
...
+    responses={
+        400: {
+            "description": "The given course code could not be found in the database",
+        },
+        200: {
+            "description": "Returns the terms a course is offered in",
+            "content": {
+                "application/json": {
+                    "example": {
+                        "terms": ["21T2", "21T2", "21T3"],
```
hdphuong, line 462: should this be `terms` and `fails`?

backend/server/routers/model.py
```diff
@@ -37,7 +37,7 @@ class CourseDetails(BaseModel):
     gen_ed: bool
     is_legacy: bool
     is_accurate: bool
-    is_multiterm: bool
+    is_multiterm: Optional[bool]
```
hdphuong, line 40: `NotRequired` may be better here?

hdphuong (commented):
LGTM! Just some nitpicking :D

backend/server/routers/courses.py
imagine-hussain: yes it should

backend/server/routers/model.py
imagine-hussain: NotRequired is actually not allowed in this context

PR #699, opened 2022-08-25
Circles 342 why course locked

sonar3d:
is a mess :'((, will clean tomorrow just want to get a start on the reviews


Peedee2002:
Screenshots plz 😁

edit: 
Screen Shot 2022-08-27 at 3 30 02 pm
Screen Shot 2022-08-27 at 3 29 54 pm

Peedee2002:
Also, mypy is angry with you, you should install the mypy extension on your ide and solve the error

backend/algorithms/objects/conditions.py
```diff
@@ -364,7 +388,8 @@ def __init__(self, exclusion: str):
         self.exclusion = exclusion
 
     def validate(self, user: User) -> tuple[bool, list[str]]:
-        return not user.in_program(self.exclusion), []
+        excluded = not user.in_program(self.exclusion)
+        return (excluded, []) if excluded else excluded, ["This course cannot be taken in your program"]
```
imagine-hussain: it may be wise to wrap brackets around the part after the else as well because otherwise it *may* come across as `return (excluded .... if ... else ... excluded), ["this course..."]`

Doing this will actually also fix ur `mypy` error on this line because you're returning a bool here as is needed but because of the above thingy, it thinks that the whole thing on the left of the last comma as one entry in the tuple

backend/algorithms/objects/conditions.py
```diff
@@ -401,9 +426,16 @@ def validate(self, user: User) -> tuple[bool, list[str]]:
         validations = [cond.validate(user) for cond in self.conditions]
         # unzips a zipped list - https://www.geeksforgeeks.org/python-unzip-a-list-of-tuples/
         unlocked, warnings = list(zip(*validations))
-        satisfied = all(unlocked) if self.logic == Logic.AND else any(unlocked)
+        if self.logic == Logic.AND:
+            satisfied = all(unlocked) 
+            return (satisfied, []) if satisfied else (satisfied, ['('+' AND '.join(sum(warnings, []))+')'])  # warnings are flattened
+            #return (satisfied, []) if satisfied else (satisfied, [f"Requires completion of all of the following courses: " + ', '.join(sum(warnings, []))])
```
imagine-hussain: #

run_app.py
```diff
@@ -49,7 +49,7 @@ def get_backend_env():
     """
     load_dotenv("./env/backend.env")
     username = os.getenv("MONGODB_USERNAME")
-    password = os.getenv("MONGODB_USERNAME")
+    password = os.getenv("MONGODB_PASSWORD")
```
imagine-hussain, line 52: mb 😔

backend/algorithms/objects/conditions.py
```diff
@@ -239,7 +249,20 @@ def _validate_course(category: Category):
                 unlocked, warnings = list(zip(*validations))
                 satisfied = all(unlocked) if category.logic == Logic.AND else any(unlocked)
                 return satisfied, warnings
-            elif isinstance(category, ClassCategory):
+
+
+            course = course.class_name
```
imagine-hussain: in this scope, for `course = course.class_name`, where is the `course` object on the RHS of the equals coming from?
afaict it may not be initialised at this point

imagine-hussain (changes requested):
ternary thingy could cause some errs down the line so good to be explicit here

also not sure abt the `course` thing

run_app.py
```diff
@@ -100,4 +100,4 @@ def main():
         logging.shutdown()
 
 if __name__ == '__main__':
-    main()
+    main()
```
Peedee2002: add newline ❤️

backend/algorithms/objects/conditions.py
```diff
@@ -239,7 +249,20 @@ def _validate_course(category: Category):
...
+                return False, []
+
+            user_grade = user.get_grade(course)
+            if user_grade is None:
+                return True, [self.get_warning()]
+            if user_grade < self.grade:
+                return False, [f"Your grade {user_grade} in course {course} does not meet the grade requirements (minimum {self.grade}) for this course"]
+            return True, []
+
+            '''elif isinstance(category, ClassCategory):
```
Peedee2002: delete stuff if you dont want it, dont comment out.

Peedee2002: in the worst case, we can recover it using git

backend/algorithms/objects/conditions.py
```diff
@@ -81,7 +81,8 @@ def is_path_to(self, course: str) -> bool:
         return self.course == course
 
     def validate(self, user: User) -> tuple[bool, list[str]]:
-        return user.has_taken_course(self.course), []
+        #return (True, []) if user.has_taken_course(self.course) else (False, [f"Prerequisite not met: {self.course}"])
+        return (True, []) if user.has_taken_course(self.course) else (False, [self.course])
```
Peedee2002, line 86: again, dont comment out, delete.

backend/algorithms/objects/conditions.py
sonar3d: yep will do, this was me being confused by the git conflict and not remembering what was going on here

backend/algorithms/objects/conditions.py
sonar3d: tbh i don't think I changed much in the GradeCondition section except adding a string to one of the returns so will look through this now

imagine-hussain:
splitting the string by replacing fullstops to have newlines after them might make it easier to read

backend/algorithms/objects/conditions.py
```diff
@@ -400,10 +417,18 @@ def validate(self, user: User) -> tuple[bool, list[str]]:
...
         # unzips a zipped list - https://www.geeksforgeeks.org/python-unzip-a-list-of-tuples/
-        unlocked, warnings = list(zip(*validations))
-        satisfied = all(unlocked) if self.logic == Logic.AND else any(unlocked)
-
-        return satisfied, sum(warnings, [])  # warnings are flattened
+        wam_warning = []
+        unlocked, all_warnings = list(zip(*validations))
+        for unlocked_cond, warning in validations:
+            if unlocked_cond and len(warning) > 0:
+                wam_warning.append(warning)
```
Peedee2002: this should be a list comprehension:
```python
wam_warning = sum(warning for unlocked_cond, warning in validations if unlocked_cond, [])
```
this also gets rid of weird issues below with types.

Peedee2002: (my code may be buggy, take it as a suggestion not as a command lol)

backend/algorithms/tests/test_conditions.py
```diff
@@ -218,19 +219,22 @@ def test_grade_condition():
...
     math1131_70_user_unlocked = math1131_70.validate(user)
     assert math1131_70_user_unlocked[0]
     assert len(math1131_70_user_unlocked[1]) == 1
     assert "Requires 70 mark in MATH1131. Your mark has not been recorded"
 
     # Has not taken the course. Should be false
     assert not (comp1521_70.validate(user))[0]
-    assert len((comp1521_70.validate(user))[1]) == 0
+    assert len((comp1521_70.validate(user))[1]) == 1
+    assert '((Need 70 in COMP1521 for this course))' in ((comp1521_70.validate(user))[1])[0]
```
Peedee2002, line 237: wierd bracketing? in both the code and the string?

backend/server/tests/planner/test_validate_term_planner.py
```diff
@@ -180,16 +180,14 @@ def test_validateTermPlanner_past_term_suppress_warnings():
             "is_accurate": True,
             "unlocked": False,
             "handbook_note": "",
-            "warnings": [],
+            "warnings": ['((COMP1531 AND (COMP2521 OR COMP1927)))'],
```
Peedee2002, line 183: whats with the extra brackets at top level?

Peedee2002 (changes requested):
few small changes needed :)

imagine-hussain:
image
Is it possible to shorten this to not repeat the `Level 2 Courses && Math Courses` and instead use (insert funky regex here) to make it something like `Level 2 Math Courses`

imagine-hussain:
> image
> 
> Is it possible to shorten this to not repeat the `Level 2 Courses && Math Courses` and instead use (insert funky regex here) to make it something like `Level 2 Math Courses`

new change lgtm 👍

backend/algorithms/objects/conditions.py
```diff
@@ -109,20 +111,29 @@ def set_logic(self, logic: Logic):
...
+                )
+                warning = []
+                for course in self.courses:
+                    if not (user.has_taken_course(course)
+                    or user.is_taking_course(course)) :
+                        warning.append(course)
+                return met, ([] if met else ['(Corequisites: ' + ' AND '.join(warning) + ')'])
             case Logic.OR:
-                return any(
+                met = any(
```
imagine-hussain: really minor nitpick but this name could perhaps be a bit more descriptive;
Currently, not sure if it's conditions met or courses w/ conditions met

backend/server/tests/planner/test_validate_term_planner.py
sonar3d: ah added them for all composite condition calls

backend/algorithms/objects/conditions.py
```diff
@@ -197,18 +210,20 @@ def validate(self, user: User) -> tuple[bool, list[str]]:
...
+        return True, [warning]
 
-    def get_warning(self, applicable_wam: Optional[float]) -> str | None:
+    def get_warning(self, applicable_wam: Optional[float]) -> str:
         """ Returns an appropriate warning message or None if not needed """
+        category = re.sub(r"courses && ([A-Z]{4}) courses", r"\1 courses", str(self.category))
+        wam_warning = f"Requires {self.wam} WAM in {category}. "
         if applicable_wam is None:
-            return f"Requires {self.wam} WAM in {self.category}. Your WAM in {self.category} has not been recorded"
+            return wam_warning + f"Your WAM in {category} has not been recorded"
```
imagine-hussain: since wam_warning is that string anyways, could just make this a f-string perhaps?

backend/server/tests/planner/test_validate_term_planner.py
Peedee2002: do these show up on the FE?

sonar3d: ye :/ not sure where to remove extra brackets

Peedee2002: Hmm that's wierd. If we can't figure out why it's being sent, we should at least strip them in the FE.

backend/algorithms/objects/conditions.py
```diff
@@ -239,16 +254,17 @@ def _validate_course(category: Category):
...
 
                 user_grade = user.get_grade(course)
                 if user_grade is None:
                     return True, [self.get_warning()]
                 if user_grade < self.grade:
-                    return False, []
+                    return False, [f"Your grade {user_grade} in course {course} does not meet the grade requirements (minimum {self.grade}) for this course"]
                 return True, []
             else:
                 return True, ["We have failed to parse this correctly"]
```
imagine-hussain, line 262: *theoretically* it would be funny if u made all of this into a one-line ternary
(probably don't do this) (but if you do, don't say it was my idea)

imagine-hussain (approved):
lgtm

backend/server/tests/planner/test_validate_term_planner.py
sonar3d: the only place I have brackets added is around any AND or OR s but im not checking if its just the composite conditions inside, could try do that or just get rid of the extra brackets at the end somewhere

Peedee2002: hmm actually @sonar3d i like this idea of checking if the brackets are needed or not by checking if the composite condition has more than one child.

backend/algorithms/objects/conditions.py
```diff
@@ -109,20 +111,29 @@ def set_logic(self, logic: Logic):
...
             case Logic.AND:
-                return all(
+                met = all(
                     user.has_taken_course(course)
                     or user.is_taking_course(course)
                     for course in self.courses
-                ), []
+                )
+                warning = []
+                for course in self.courses:
```
Peedee2002: this should also be a list comp :)
```
warning = [course for course in self.courses if not user.has_taken_course(course) and not user.is_taking_course(course)]
```

Peedee2002: or better yet, we can use this to replace `met`, so we dont run the same calculation twice.
if the warnings list is empty, we know its true, else false.

backend/algorithms/objects/conditions.py
```diff
@@ -434,3 +460,10 @@ def __str__(self, id='start') -> str:
                 data['children'].append(json.loads(str(cond)))
         return json.dumps(data)
 
+def flatten_list_list(nested_list: List[List[str]]) -> List[str]:
```
Peedee2002: i dont think you need to use this function anymore if you use my list comprehension above.

Peedee2002 (changes requested):

Peedee2002:
sorry i know its a lot of review comments 😅

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
Peedee2002: this file should be deleted

backend/server/tests/planner/test_validate_term_planner.py
imagine-hussain: image
I added the bracket cleaner to the FE. It removes any extra parens that encompass the whole thing like `(smth smth smth )` but, leaves in those that are part of a composite condition like `(smth) AND (smth)`

sonar3d: > 

will have a look at the children thing

sonar3d:
> sorry i know its a lot of review comments 😅

better than not enough reviews!

backend/server/tests/planner/test_validate_term_planner.py
sonar3d: update had a look at children thing but is messyyy any suggestions welcome

backend/algorithms/objects/categories.py
```diff
@@ -49,6 +51,12 @@ def set_logic(self, logic: Logic):
         """ AND or OR """
         self.logic = logic
 
+    def is_child(self, child: bool):
+        self.child = child
+    
+    def set_isWrapAnd(self, isWrap: bool):
```
imagine-hussain: i think pylint will complain abt this not being snake case

sonar3d: ah yes will add to my list of fixing for tomorrow '.'

imagine-hussain (commented):
image

For something like this, is it possible to differentiate the actual condition, from the feedback we're providing?

The things that come to mind are either
- have feedback be a different colour
- could wrap feedback in `{}` or smth else, but idk if that would clutter the string too much
- return the feedback as a seperate string that can get rendered afterwards, so that it looks like

```
48 UOC required in all courses AND (COMP1917 OR COMP2521)
You have done 0 uoc.
```

backend/algorithms/objects/categories.py
```diff
@@ -40,6 +40,8 @@ class CompositeCategory(Category):
     def __init__(self, logic: Logic = Logic.AND):
         self.categories: list[Category] = []
         self.logic = logic
+        self.child = False
+        self.isWrapAnd = False
```
imagine-hussain: i think now that there's quite a few attributes in this class, its a good time to write a docstring for what the attributes do since I can already feel myself start to forget the diff between `child` and `isWrapAnd`

backend/algorithms/objects/categories.py
```diff
@@ -49,6 +51,12 @@ def set_logic(self, logic: Logic):
         """ AND or OR """
         self.logic = logic
 
+    def is_child(self, child: bool):
```
imagine-hussain: this is nitpicking but, i think this could be better as `set_is_child` as i feel `is_child` feels like it denotes a function that checks whether the given object is a child or not.

Also, for something like this, is it possible for the caller to just modify the attribute directly?

```py
cond = Condition()
cond.is_child = True
```

It may not be allowed due to some property we're inheriting but, i think this would help simplify things

imagine-hussain (changes requested):

backend/algorithms/objects/conditions.py
```diff
@@ -390,20 +401,56 @@ def set_logic(self, logic: Logic):
...
         """
+
+        # set is_child for composite conditions
+        count = 0
         if not self.conditions:
             return True, []
+        for cond in self.conditions:
+            if isinstance(cond, CompositeCondition): 
+                if self.isWrapAnd: 
+                    if count > 0: cond.is_child(True)
```
Peedee2002: ok sorry i really hate this code lol

Im happy to shunt this to the FE.

Peedee2002: sorry I am indecisive about this 🧎‍♂️

Peedee2002:
I think we can get this in (?)

sonar3d:
> I think we can get this in (?)

yep, think i need to pull/check some tests after the last revert/commit then ill hit enable auto merge

backend/server/tests/planner/test_validate_term_planner.py
sonar3d: update #1000 this be scrapped cause code ew backend way - thanks Hussain for frontend help

imagine-hussain (approved):

Peedee2002 (approved):

PR #703, opened 2022-08-26
hotfix/legacy data

imagine-hussain:
- rerun processors to fix `getLegacyCourse` data
- `terms` still comes up as `[]` for some courses but that may be a processors bug (TBA)


imagine-hussain:
I've tested the route against *most* of the core cs courses and lgtm but feel free to run locally and see if anything currently still broke

Peedee2002 (approved):

PR #709, opened 2022-08-27
Mae/circles 346 local storage

hdphuong:
- create a save button to migrate local storage to BE
- I'm not sure why the json files are modified 😳


Peedee2002:
i think the json files are modified because we reran the processors. Can you rerun them too then? I think you got some wild merge issues.
Also, no tests are running on this because its not going to dev 👀 I guess we will have to be ok with that.
EDIT: wait nvm this has the updated code i think

Peedee2002 (approved):

gtangelo:
Can we update the data migration branch to dev? So the changes are more manageable to look at haha 😅

hdphuong:
Yup just merged in dev. It's more manageable to review now :D

backend/server/example_input/example_local_storage_data.json
```diff
@@ -0,0 +1,53 @@
...
+        "mostRecentPastTerm": {
+            "Y": 1,
+            "T": 0
+        },
+        "programName": "Computer Science",
+        "unplanned": ["COMP6447"],
+        "numYears": 3,
+        "isSummerEnabled": false
+    }
+}
```
Peedee2002: add newline 🥺

backend/server/routers/planner.py
```diff
@@ -76,3 +76,8 @@ async def validate_term_planner(plannerData: PlannerData):
         user.year += 1
 
     return {"courses_state": coursesState}
+
+@router.post("/saveLocalStorage/")
+def save_local_storage(localStorage: LocalStorage):
+    #TODO: replace this with a real implementation
+    print(localStorage)
```
Peedee2002: i think this should also take in a token.

Peedee2002: if we take a token, we can make an actual implementation of this. I have the database set up, so you can add to it by using a token as a key, with an objectId is a value for that token to be mapped to. look at databases.py.

frontend/src/pages/TermPlanner/OptionsHeader/OptionsHeader.tsx
```diff
@@ -3,10 +3,12 @@ import React from 'react';
 import { FaRegCalendarTimes } from 'react-icons/fa';
```
Peedee2002, line 3: why are there there these FE changes?

Peedee2002: oop im a dummy

frontend/src/components/RadialBarChart/RadialBarChart.tsx
```diff
@@ -1,32 +0,0 @@
-import React, { Suspense } from 'react';
```
Peedee2002, line 1: why is this file deleted?

Peedee2002 (changes requested):
I think there are a few wierd things that happened when you were merging...
Have a look at the diff to see what has changed and see if you meant to make those changes. if not, revert them.

frontend/src/components/RadialBarChart/RadialBarChart.tsx
gtangelo: deleted this component file because it wasn't used at all - think it showed up here since the data migration branch was not merged to the latest dev?

Peedee2002: wierd, i thought it was updated. I think we may have borked something then. No worries, we will just accept those changes then.

frontend/src/pages/TermPlanner/OptionsHeader/OptionsHeader.tsx
```diff
@@ -71,6 +73,22 @@ const OptionsHeader = ({ plannerRef }: Props) => {
             </Tooltip>
           </div>
         </Tippy>
+        <Tippy
```
gtangelo, line 75: When clicking the save button, it show any indication that it is being saved to the backend? Also, is there a reason to use tippy when we are not displaying a popup?

Think this could be improved by either of these approaches:
- when clicking the save button, it brings a popup to confirm to save to the backend? (similar to export planner/unschedule all courses buttons) - right now the popup is empty
- when clicking the saving icon, it creates a spinner animation once the save request has been processed and displays a tick icon to indicate it has been save for a few seconds

frontend/src/utils/prepareLocalStorageData.ts
```diff
@@ -0,0 +1,62 @@
...
+const prepareLocalStorageData = (
+  planner: PlannerSliceState,
+  degree: DegreeSliceState,
+): LocalStorageData => {
+  const {
+    years, startYear, courses, unplanned, numYears, isSummerEnabled,
+  } = planner;
+  const { programCode, programName, specs } = degree;
+
+  const plan: YearPlan[] = [];
```
gtangelo, line 33: we are basically copyiong the logic when creating the plan from the other prepareCoursesForValidation util function right? Do you expect the two endpoints require the same schema for the plan? If so, might be better to abstract it out into another util function like `createYearPlane` that creates the year plan and just use that util function in both places?

frontend/src/utils/saveLocalStorageData.ts
```diff
@@ -0,0 +1,15 @@
+import axios from 'axios';
+import { DegreeSliceState } from 'reducers/degreeSlice';
+import { PlannerSliceState } from 'reducers/plannerSlice';
+import prepareLocalStorageData from './prepareLocalStorageData';
+
+const saveLocalStorageData = (): void => {
```
gtangelo: nitpict: can we rename this util function to `migrateLocalStorageData`? Might be a better name imo since `saveLocalStorageData` sounds like its saving data rather than migrating it?

gtangelo (changes requested):

gtangelo:
Something to consider, before merging the final data migration changes in to dev, we may want to remove the use of `redux-persist`. As such, it may not be likely that the way to save the user data is through the term planner button. Instead, it might be better as a popup? (similar to how when u try to reset your degree wizard, it shows a warning popup if you want to reset it). Though it's something to worry bout later on I guess.

backend/server/tests/planner/test_save_local_storage.py
```diff
@@ -0,0 +1,20 @@
...
+def test_saveLocalStorage_empty():
+    x = requests.post(
+        'http://127.0.0.1:8000/planner/saveLocalStorage', json=DATA["empty_year"])
+    assert x.status_code == 200
+
+
+def test_saveLocalStorage_simple():
+    x = requests.post(
+        'http://127.0.0.1:8000/planner/saveLocalStorage', json=DATA["simple_year"])
+    assert x.status_code == 200
```
Peedee2002, line 19: Can we also check the database?

I know that's not a black box test, so it might be better to expand this test later. Tell me what you think is better.

Peedee2002 (approved):
Just a comment on testing, but looks good!

backend/server/tests/planner/test_save_local_storage.py
hdphuong: I agree. I added something to it but it's not working...

PR #712, opened 2022-08-30
debug(cf-326): Trees with only one node will not be shown

kellieyau:
Those with only one node are hidden e.g gen5013:

image

Those that unlock or have prereques and unlock still act the same:
image
image


gtangelo (changes requested):
Lets provide a fallback text (maybe something like `No prerequisite visualisation is needed for this course` instead rather than showing an empty container?

frontend/src/components/PrerequisiteTree/PrerequisiteTree.tsx
```diff
@@ -113,15 +113,17 @@ const PrerequisiteTree = ({ courseCode }: Props) => {
 
       // render graph
       if (!graph) {
-        generateTreeGraph(graphData);
+        if (graphData.children.length !== 0) {
+          generateTreeGraph(graphData);
+        } else {
+          <p>No prerequisite visualisation is needed for this course</p>;
```
gtangelo: Since this is in the useEffect, it is not returning this html and thus not showing it. You would need to conditionally render this around line ~132
```
{loading && <Spinner text="Loading tree..." /> }
{!loading && <TODO> && <p> No prerequisite visualisation is needed for this course </p> }
```
Replace `<TODO>` from above with the logic to check if the graph has any prereqs and unlocks nodes. You can't use`graphData` unlike in line 116 right now since that is only the scope of the useeffect call. I think you can check via the `graph` state variable (i.e. get num nodes of graph, if its only one, then its only has no unlocks/prereqs nodes)?

kellieyau: awesome should be good to go with the most updated branch, in future how do I use console log or its equivalent to chek the graph?

gtangelo: Oh you can just console log the graph getters and setters, since the graph var is just an instance of a class. Typescript can just tell you what methods there are to extract the graph internal representation

gtangelo: Also, feel free to revert this change since you dont need this logic anymore?

frontend/src/components/PrerequisiteTree/PrerequisiteTree.tsx
```diff
@@ -113,21 +113,25 @@ const PrerequisiteTree = ({ courseCode }: Props) => {
...
       setLoading(false);
     };
-
     if (courseCode) setupGraph(courseCode);
   }, [courseCode, degree, dispatch, graph, planner]);
 
   return (
     <S.PrereqTreeContainer ref={ref} height={calcHeight(coursesRequires, courseUnlocks)}>
       {loading && <Spinner text="Loading tree..." />}
+      {!loading && (!graph?.getNodes)
```
gtangelo: nitpick: do we need the brackets over `!graph?.getNodes`? Might aswell remove it to be consistent with style

gtangelo (changes requested):

frontend/src/components/PrerequisiteTree/PrerequisiteTree.tsx
kellieyau: true

kellieyau: when you mean revert i still need the second if condition because that determines if there are any children:

`if (!graph && graphData.children.length !== 0) {
        generateTreeGraph(graphData);
      } else {
        // NOTE: This is for hot reloading in development as new graph will instantiate every time
        updateTreeGraph(graphData);
      }`

gtangelo: Oh wait I just realised it should be `{!loading && graph?.getNodes() <= 1 && ...` rather than `{!loading && (!graph?.getNodes) ...`. Forgot the brackets haha since getNodes is a function/method.

For the revert, I think it doesn't matter (or atlease I dont think). The graph should generate with 1 children (the course node (i.e. if the tree belongs to comp1511, it will show the comp1511 node only if there are no children) and hence the `graph.getNodes()` will return only 1 which you can use to compare.

gtangelo: Oooo wait nvm, dont mind what I said, i Think I get what you mean. Dont need to revert it if it works on your end. This should just be fine then? `if (!graph && graphData.children.length !== 0) { generateTreeGraph(graphData); } else { // NOTE: This is for hot reloading in development as new graph will instantiate every time updateTreeGraph(graphData); }`

gtangelo (approved):

frontend/src/components/PrerequisiteTree/PrerequisiteTree.tsx
kellieyau: oh ok sweet!

PR #715, opened 2022-08-30
feat(Graphical Selector): Zoom in and out of graph

kellieyau:
Note: 
zoom occurs towards the top right of the graphical selector

https://user-images.githubusercontent.com/97064025/187418018-0eaf3880-7d95-4ff7-aeaa-817e4c92ca54.mov


frontend/src/pages/ProgressionChecker/ProgressionChecker.tsx
```diff
@@ -247,6 +247,7 @@ const ProgressionChecker = () => {
                     <TableView
                       uoc={structure[group].content[subgroup].UOC}
                       subgroupTitle={subgroup}
+                      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
gtangelo: reason why we are disabling eslint check on these lines?

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
```diff
@@ -200,6 +203,19 @@ const GraphicalSelector = () => {
     }
   };
 
+  const zoomIn = () => {
+    if (!graph) return;
+    const zoom = graph.getZoom();
+    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
gtangelo: prefer not disabling eslint line here. You can avoid it by using if statement instead!

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
```diff
@@ -200,6 +203,19 @@ const GraphicalSelector = () => {
...
+    zoom >= 1 ? graph.zoomTo(zoom + 0.5) : graph.zoomTo(zoom + 0.1);
```
gtangelo: what is the `>= 1` check for?

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
```diff
@@ -200,6 +203,19 @@ const GraphicalSelector = () => {
...
+  };
+  const zoomOut = () => {
+    if (!graph) return;
+    const zoom = graph.getZoom();
+    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+    zoom >= 1 ? graph.zoomTo(zoom - 0.5) : graph.zoomTo(zoom - 0.1);
```
gtangelo: would be better to define `0.5` as a constant above and reuse that variable?

gtangelo (changes requested):

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
kellieyau: so basically when the zoom becomes < 1 it zooms really far out in proportion to if the zoom is 2, 3, 4 etc. Hence at that point we need to reduce the zoom amount by 0.1. Its based on the builtin zoom feature

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
kellieyau: yep zoomRatio is the new variable

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
kellieyau: igtm

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
```diff
@@ -200,6 +203,29 @@ const GraphicalSelector = () => {
...
+      zoomRatio = 0.1;
+    }
+    graph.zoomTo(zoom + zoomRatio);
+  };
+  const zoomOut = () => {
+    if (!graph) return;
+    let zoomRatio;
+    const zoom = graph.getZoom();
+    if (zoom >= 1) {
+      zoomRatio = 0.5;
```
gtangelo: > yep zoomRatio is the new variable

Oh like I mean can we define `0.5` and `0.5` as const? i.e. example, fyi not sure the context of the var so rename these vars as suited
```
// place these var declaration at the top of the file (just after the imports)
// consider rename the variable to a better context. Not entirely sure what 0.5 and 0.1 means in this case.
const ZOOM_RATIO_IN = 0.5;
const ZOOM_RATIO_OUT = 0.1;
// might also make 1 as the zoom limit? again rename var - im just gussing what 1 means
const ZOOM_LIMIT = 1;
...
// within the zoom out function
if (zoom >= ZOOM_LIMIT) zoomRatio = ZOOM_RATIO_IN;
else zoomRatio = ZOOM_RATIO_OUT;
```

frontend/src/pages/ProgressionChecker/ProgressionChecker.tsx
```diff
@@ -257,6 +258,7 @@ const ProgressionChecker = () => {
                     <GridView
                       uoc={structure[group].content[subgroup].UOC}
                       subgroupTitle={subgroup}
+                      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
gtangelo: can we remove this elsint disable line (same goes at line ~250)?  Not sure why this was added in the git diff?

gtangelo (changes requested):

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
kellieyau: yep so ive just adjusted that essentially what occurs is that 0.5 and 0.1 should be used because when the zoom is <1 it snaps into really far out/in when the zoom is in that range. Hence values from 0.2 - 1 are really sensitive compared to the zoom values of 2-5. I've left it as ZOOM_LIMIT but could be a better name like turning point or end point... idk

frontend/src/pages/ProgressionChecker/ProgressionChecker.tsx
kellieyau: yep should be done

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
```diff
@@ -200,6 +207,25 @@ const GraphicalSelector = () => {
     }
   };
 
+  function getZoomRatio() {
+    if (!graph) return 0;
```
gtangelo, line 211: `return;` over `return 0;`

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
```diff
@@ -200,6 +207,25 @@ const GraphicalSelector = () => {
...
+    const zoom = graph.getZoom();
+    if (zoom >= ZOOM_LIMIT) return HIGHER_ZOOM_RATIO;
```
gtangelo: use ternary return `return graph.getZoom() >= ZOOM_LIMIT ? HIGHER_ZOOM_RATIO : ...`

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
```diff
@@ -200,6 +207,25 @@ const GraphicalSelector = () => {
     }
   };
 
+  function getZoomRatio() {
```
gtangelo: prefer to use arrow function here

`const getZoomRatio = () => { ... } `

gtangelo (changes requested):
forgive me 🥺

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
kellieyau: sweet!

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
kellieyau: there are two problems with that:
1. the return type must be an integer as the function returns must be a consistent type
2. Since we add it to the funcs zoomIn and zoomOut it would return an undefined value and we cannot not add that to zoom `graph.zoomTo(zoom + getZoomRatio());`

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
kellieyau: yep! should be fixed with the update :D

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
gtangelo: Oh right me dum dum xd. I legit didn't see it was returning a number before whoops

gtangelo: Thanks for the clarifications :D

gtangelo (approved):

gtangelo (approved):

frontend/src/pages/GraphicalSelector/GraphicalSelector.tsx
kellieyau: haha all g thx to for the merge :D

PR #744, opened 2022-09-07
feature(conditions): add support for core courses

Peedee2002:
The number of changes looks scary, but that is because i also had to run processors on this branch.
The code changes arent that large.

There is some impotous to get this branch in btw, since it also fixes the Engineering majors bug oops


backend/algorithms/objects/user.py
```diff
@@ -73,7 +78,8 @@ def has_taken_specific_course(self, course):
 
     def has_taken_course(self, course: str):
         """ Determines if the user has taken this course """
-        return course in self.courses or any(c in self.courses for c in (CACHED_EQUIVALENTS.get(course) or []))
+        return any(c in self.courses and (self.courses[c][1] or 50) >= 50 for c in chain([course], (CACHED_EQUIVALENTS.get(course) or [])))
```
Peedee2002, line 81: spicy, i know

backend/server/routers/utility.py
```diff
@@ -17,3 +17,17 @@ def map_suppressed_errors(func: Callable, errors_log: List[Any], *args, **kwargs
     except Exception as e:
         errors_log.append((*args, e))
     return None
+
+def get_core_courses(program: str, specialisations: list[str]):
```
imagine-hussain, line 21: bruh

backend/server/routers/programs.py
```diff
@@ -391,7 +405,7 @@ def add_program_code_details(structure: dict[str, StructureContainer], programCo
     structure['Rules'] = {"name": "General Program Rules", "content": {}}
     return (structure, programsResult["UOC"])
 
-def add_geneds_to_structure(structure: dict[str, StructureContainer], programCode: str) -> dict[str, StructureContainer]:
+def add_geneds_to_structure(structure: dict[str, StructureContainer], programCode: str, spec: Optional[str]) -> dict[str, StructureContainer]:
```
imagine-hussain: what does adding `spec` here do? afaict, it's not being used

backend/algorithms/objects/conditions.py
```diff
@@ -276,6 +277,30 @@ def __str__(self) -> str:
...
+    def set_category(self, category_classobj: Category):
+        """ Set own category to the one given """
+        self.category = category_classobj
+
+    def is_path_to(self, course: str) -> bool:
+        return self.category.match_definition(course)
+
+    def validate(self, user: User) -> tuple[bool, list[str]]:
+        res = user.completed_core(self.category)
+        return res, [] if res else [f'you have not completed your {self.category} cores']
```
imagine-hussain: adding a pair of parenthesis after the comma might be a good idea to prevent people from misunderstanding the order of operations and accidentally breaking smth in future

backend/algorithms/objects/conditions.py
```diff
@@ -276,6 +277,30 @@ def __str__(self) -> str:
         })
 
 
+class CoresCondition(Condition):
```
imagine-hussain, line 280: iirc (damn its been a while), the category was typically smth along the lines of 

> You must finish Level 1 Econ Cores?

It may be worth making explicit in the docs that `category` defines what subset of the core courses u need to do

backend/algorithms/objects/helper.py
```diff
@@ -3,13 +3,15 @@
...
 
 
 class Logic(Enum):
     """ Logic Keywords """
-    AND = auto()
-    OR = auto()
+    AND = "&&"
+    OR = "||"
+    def __eq__(self, other):
+        return self.__class__ is other.__class__ and other.value == self.value
```
imagine-hussain: not saying this is wrong or needs to be changed, but out of curiosity, is there a reason to use `__class__` with `is` instead of `type(self) == type(other)`

backend/algorithms/tests/test_user_has_taken.py
```diff
@@ -0,0 +1,80 @@
...
+    user.add_courses({
+        "COMP1511": (6, 100)
+    })
+    assert user.has_taken_course("COMP1511") is True
+
+
+def test_user_has_taken_user_1():
+    user = User(USERS["user1"])
+    assert user.has_taken_course("COMP1511")
+    assert user.has_taken_course("COMP1511")
```
imagine-hussain: is this line repeated on purpose?

backend/algorithms/tests/test_user_has_taken.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def test_user_has_taken_empty_user():
+    """
+    The user object being passed in for construction is empty.
+    Note that this is differentiated from the case where the user
+    has no courses.
+    """
+    user = User()
+    assert user.has_taken_course("COMP1069") is False
```
imagine-hussain, line 23: i think i remember when these were getting written the rationale was

- Using `assert not user.has_taken_course`=
    - Pros: very pythonic
- Using `assert user.has_take_course is False`
    - mmm lines have very pretty aligned aesthetic even when you assert for the true case

imagine-hussain (changes requested):
*basically* lgtm

bit unsure abt adding `spec` but the rest is 👍

backend/server/routers/utility.py
Peedee2002: I mean importing it would be `pain` because of  cycle errors

Peedee2002: But fine ill make it a dynamic import lol

backend/algorithms/objects/helper.py
Peedee2002: oh oops, i think this was me being dumb. Ill remove this equality since i am not using it

Peedee2002: (dont copy paste off of stack overflow kids)

backend/algorithms/tests/test_user_has_taken.py
Peedee2002: lmao nice. There is the semantic difference that has_taken_course must also be a boolean, but we have mypy for that now 😎

backend/server/routers/utility.py
imagine-hussain: oh, the import location is fine
i was saying bruh cos epic one liner

imagine-hussain (approved):
👍

backend/server/routers/utility.py
Peedee2002: Surely indentation fixed it :tm:

PR #764, opened 2022-09-17
Mae/circles 376 enrolment data

hdphuong:
- ssh to cse server and collect data with `members`
- hash zIDs before storing to a file in final_data
- a raw zIDs file is also created at `backend/data/scrapers/enrolmentDataRaw.json` which you can use to double-check the accuracy of data collected. This file is included in `.gitignore` so it won't be public.


backend/data/scrapers/enrolment_scraper.py
```diff
@@ -0,0 +1,74 @@
...
+        ssh = paramiko.SSHClient()
+        ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+        ssh.connect('login.cse.unsw.edu.au', username=username, password=password)
+        stdin, stdout, stderr = ssh.exec_command('ls', timeout=1)
+    except paramiko.ssh_exception.AuthenticationException:
+        print('Authentication failed')
+        exit(0)
+    except paramiko.ssh_exception.SSHException:
+        print('SSH failed')
+        exit(0)
```
Peedee2002: Probably return 1 here

backend/data/scrapers/enrolment_scraper.py
```diff
@@ -0,0 +1,74 @@
...
+        if not course_code.startswith("COMP"):
+            continue
+        print(course_code)
+        try:
+            waiting.wait(lambda: stdin.channel.send_ready(), timeout_seconds=13)
+            stdin.channel.sendall(f'members {course_code}_Student\n'.encode())
+
+            waiting.wait(lambda: stdout.channel.recv_ready(), timeout_seconds=13)
+            result = stdout.channel.recv(MAX_BYTES).decode('utf-8')
+            enrolments = list(map(lambda s: s.split(' ')[0], result.split('\n')))
```
Peedee2002: I think you should use list comprehension here instead of lambda and map (it hurts me too, I used to abuse the hell out of this notation)

Peedee2002: There are a few more that use this notation, all of these can be turned into list comprehensions. I know they are equivalent, but you have to admit that python makes list comprehensions much easier to read :sunglasses:

backend/runprocessors.py
```diff
@@ -99,6 +126,9 @@ def run_manual_fixes():
         "program": cache_program_mappings,
         "graph": cache_graph
     },
+    "enrolment": {
+        "scrape": run_scrape_enrolment_data,
```
Peedee2002, line 130: Have a more explicit error here instructing the user to add username and password if not given.

Peedee2002 (commented):
Small changes to make to make your script a bit more user friendly, but otherwise awesome :pog:

backend/data/scrapers/enrolment_scraper.py
hdphuong: Sure but what difference would this make?

Peedee2002: If we decide to run this stuff in pipelines, exiting with 1 will indicate an error has happened in the pipeline. I eventually want to add scrapers to the pipeline so this would be a nasty bug to find later lol

hdphuong: ah okok

backend/data/scrapers/enrolment_scraper.py
```diff
@@ -0,0 +1,78 @@
...
+        ssh = paramiko.SSHClient()
+        ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+        ssh.connect('login.cse.unsw.edu.au', username=username, password=password)
+        stdin, stdout, stderr = ssh.exec_command('ls', timeout=1)
+    except paramiko.ssh_exception.AuthenticationException:
+        print('Authentication failed')
+        return 1
+    except paramiko.ssh_exception.SSHException:
+        print('SSH failed')
+        return 1
```
Peedee2002: Oh no this isn't what I meant, I just meant `exit(1)` instead. We want the process to die if this fails

backend/data/scrapers/enrolment_scraper.py
```diff
@@ -0,0 +1,78 @@
...
+        try:
+            waiting.wait(lambda: stdin.channel.send_ready(), timeout_seconds=13)
+            stdin.channel.sendall(f'members {course_code}_Student\n'.encode())
+
+            waiting.wait(lambda: stdout.channel.recv_ready(), timeout_seconds=13)
+            result = stdout.channel.recv(MAX_BYTES).decode('utf-8')
+
+            enrolments = []
+            for res in result.split('\n'):
+                if re.match(r'[0-9]{7}', res):
```
Peedee2002: Oh, here I meant:
```py
enrolments = [res.split(' ')[0] for res in result.split('\n') if re.match(r'[0-9]{7}', res)]
```
Instead of the map. (That syntax^ is called "list comprehension", because you write what you want to be inside the list, instead of making the list by hand in the way you have it here)

Peedee2002 (commented):
Sorry, I think I should have given more code examples in my last review

backend/data/scrapers/enrolment_scraper.py
hdphuong: oops mb should have clarified

Peedee2002 (approved):

PR #771, opened 2022-09-20
fix: rerun enrolment scraper to get termly data

hdphuong:
- get termly enrolment data 
- I ignored mypy in this PR, will update it with another PR (also need to add threading since too much time is spent on waiting)


Peedee2002 (approved):
Sure, I trust you know what you're doing, but you should definitely revisit those mypy errors later

hdphuong:
Tyty will do sorry my brain not working now 😅

PR #778, opened 2022-09-24
tests(backend): legacy courses not returning for years before 2021 [CIRCLES-373]

imagine-hussain:
- tests for `getLegacyCourse` route


backend/server/tests/courses/test_get_legacy_course.py
```diff
@@ -0,0 +1,122 @@
+"""
+Tests the `courses/getLegacyCourse` route to ensure that legacy
+data is working.
+TODO: Update this once LIVE_YEAR is updated to 2023.
```
Peedee2002: We can remove this TODO: now

Peedee2002 (approved):

PR #779, opened 2022-09-24
chore(ci): add test reporting to ci

lhjt:
This PR adds the test result comment for all PRs.

### DIFF
--- a/.github/workflows/ci_BE.yml
+++ b/.github/workflows/ci_BE.yml
@@ -5,20 +5,19 @@ on:
     branches: [dev, main]
 
 jobs:
-  build:
+  test:
     name: Backend CI Testing
-
     runs-on: ubuntu-latest
-
+    permissions:
+      checks: write
+      pull-requests: write
     steps:
       - name: Checking out repository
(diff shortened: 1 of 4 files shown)


Peedee2002:
what does this do 👀

lhjt:
> what does this do 👀

Once/If I get it to work, you'll see 😎

Peedee2002:
Yo @lhjt there is a way to exclude files from coverage right?

lhjt:
> Yo @lhjt there is a way to exclude files from coverage right?

Yes, [there is](https://pytest-cov.readthedocs.io/en/latest/config.html?highlight=exclude#:~:text=using%20a%20.coveragerc%20file%20with%20the%20omit%20option%20set). Feel free to add in a `.coveragerc` file to this branch and push it and we can check it out.

lhjt:
I'm unsatisfied with the format that the code coverage PR comment uses so I have removed it for the time being. Otherwise this is ready for review.

Peedee2002 (approved):
Very cool

PR #787, opened 2022-09-26
chore(migration): implemented toggleSummerTerm and fixed bad issues with the workflow

Peedee2002:
So i am mostly opening up this PR early as an example, and will discuss tomorrow.


backend/server/database.py
```diff
@@ -112,7 +110,7 @@ def create_dynamic_db():
                             "isSummerEnabled": {
                                 'bsonType': 'bool'
                             },
-                            "plan": {
+                            "years": {
```
hdphuong, line 113: the frontend has this as `plan`, I think we should have this as `plan` for consistency.

Peedee2002: Hmm as far as I can tell, it's `years`

(Have a look at plannerSlice.ts)

hdphuong: Gabriel told me to get data from the root local storage instead of redux slices?  https://compclub.atlassian.net/jira/software/projects/CIRCLES/boards/5?selectedIssue=CIRCLES-346 

iirc last time i printed it out it has a different data structure

Peedee2002: Hmm I also printed it out, that's how I noticed the differences 

Can you double check on your end?

hdphuong: oops my bad, yup, it's `years` sorry for that xD

backend/server/routers/user.py
```diff
@@ -1,19 +1,62 @@
...
 router = APIRouter(
     prefix="/user",
     tags=["user"],
 )
 
+# DANGEROUS!!!!!!!
+# NEVER ALLOW TO RUN IN PROD
+# PLEASE DONT BE STUPID
+if getenv("DANGEROUS_ALLOW_DELETE_DB_REQUEST") == "true":
+    @router.post("/drop")
```
lhjt: It would be a better idea to have your tests directly execute MongoDB queries rather than relying on an endpoint like this 😄 

I will re-review once changes that were discussed on Discord are made (env vars for testing)

lhjt (changes requested):

backend/server/database.py
```diff
@@ -200,7 +198,7 @@ def overwrite_all():
     overwrite_archives()
 
 def optionally_create_new_data():
-    # TODO: DELETE NEXT LINE IN PROD
+    # TODO: DELETE NEXT LINES IN PROD
```
lhjt, line 201: Is this still relevant?

backend/server/tests/user/utility.py
```diff
@@ -0,0 +1,13 @@
+import os
+from dotenv import load_dotenv
+load_dotenv("../env/backend.env")
+
+os.environ["MONGODB_SERVICE_HOSTNAME"] = "localhost"
```
lhjt, line 5: Should this be set conditionally instead of being overwritten?

backend/server/database.py
Peedee2002: Yes, I want that for now, this gets overwritten every time you relaunch your app, but this is not desirable in prod

backend/server/tests/user/utility.py
Peedee2002: No, because for testing, you will always be hitting a localhost db. If you aren't, there are bigger problems. 

You may be launching your stuff in docker, so this variable will be `mongodb`. but when you are testing, you are launching a process from your local machine, not through docker. So, the database will be at `localhost`, not `mongodb`. So I need to overwrite it for those using docker to have this test run smoothly.

lhjt: Yeah alright, that makes sense; if I run into issues with this running in CI I'll raise it then but otherwise understandable (if my understanding is correct for the docker scenario)

lhjt (approved):

PR #795, opened 2022-09-30
feat(backend) Expand `ProgramRestriction` classes [CIRCLES-364]

imagine-hussain:
Contains the following classes:
- `CompositeRestriction`
- `CategoryRestriction`
- `CategoryUOCRestriction`
- `CourseRestriction`

I was hesitant to write these initially because these are unused and won't be built since the pre-processing / tokenising for these types of conditions hasn't happened yet. But
- I was bored and this seemed like less work than the processors
- It should be fine to merge these in since their structure is pretty uniform wrt to the other Restrictions and should be agnostic of what the processors are doing anyways.


Peedee2002 (approved):
Nice

PR #811, opened 2022-10-04
fix(Course Menu): Locked courses are only shown when toggle is off

kellieyau:
image
image


Peedee2002 (approved):
Oop

gtangelo (changes requested):
What I meant in the ticket was that if a course in your planner card/planner is still locked, it should show it on the course menu sidebar even when the 'Show all courses' is toggled off

gtangelo:
For example, if I only have `COMP2511` in my planner when doin a comp sci major, when the toggle is off, it the course menu should display COMP1511, COMP2511, MATH1131, MATH1081.

Right now, it showes COMP1511, MATH1131, MATH1081.

We want to consider locked courses in your planner card as 'unlocked' per say when the toggle is off (with the lock icon alongside it)

kellieyau:
> Member

OOOOOO yep that makes mroe sense, should be corrected in the update :D
image

gtangelo:
Closing the PR coz I think this was done already? Don't know why there's git diff though haha. But the fix seems to be on prod already soooo..

PR #838, opened 2022-10-16
chore(ci): use seperate test result workflows 

sachk:
just a quick test to see if this makes the frontend unit tests run


lhjt (approved):
Yes for testing

PR #845, opened 2022-10-17
feat(backend): provide most common followup courses to a given course [CIRCLES-377]

Basialo:
I finally did it!! 
((let me know what's wrong with it....))

This is a router that uses enrolment data to calculate the most common courses to be taken in succession. We currently only have 2 terms of data, so it's limited to T2->T3 atm.


Peedee2002:
ah fun, mypy is failing you lol - you need to download the mypy plugin for your favourite IDE, or use `mypy backend` to check for mypy errors.

backend/server/routers/followups.py
```diff
@@ -0,0 +1,86 @@
...
+def get_followups(origin_course: str, origin_term: str) -> Dict[str, str]:
+    # origin_term is the term that the original course was/would be taken in
+ 
+    # we only have enrolment data from T2(2022) -> T3(2022) at this stage
+    origin_term = "T2" # remove this line after we get more data
+    next_term = get_next_term(origin_term)
+
+    # get all comp courses
+    all_comp_courses = []
+    for course in coursesCOL.find():
```
Peedee2002: better expressed as:
```
all_comp_courses = [course[code] for course in coursesCOL.find()]
```

backend/server/routers/followups.py
```diff
@@ -0,0 +1,86 @@
...
+        if course["code"].startswith("COMP"):
+            all_comp_courses.append(course["code"])
+    
+    # get enrolment data
+    enrolment_data_file = open("./data/final_data/enrolmentData.json")
```
Peedee2002: better expressed as:
```
with open("./data/final_data/enrolmentData.json") as enrolment_data_file:
    enrolment_data = json.load(enrolment_data_file) 
```

backend/server/routers/followups.py
```diff
@@ -0,0 +1,86 @@
...
+    enrolment_data = json.load(enrolment_data_file)
+    enrolment_data_file.close()
+    origin_course_members = enrolment_data[origin_course][origin_term]
+    
+    # calculate followups
+    followups = {}
+    for course in all_comp_courses:
```
Peedee2002, line 77: very pretty implementation 😍

backend/server/routers/followups.py
```diff
@@ -0,0 +1,86 @@
...
+        #  people who took the origin_course in origin_term 
+        #  people who took this course in the following term.
+        num_followups = len(set(enrolment_data[course][next_term]) & set(origin_course_members))
+        
+        if num_followups != 0:
+            followups.update({ course: { next_term: num_followups } })
+
+    top_followups = sorted(followups.items(), reverse=True, key=lambda course: course[1][next_term])
+ 
+    return {"originCourse": origin_course,
```
Peedee2002: fix indentation here

backend/server/routers/followups.py
```diff
@@ -0,0 +1,86 @@
...
+                                "T3": 45,
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+)
+def get_followups(origin_course: str, origin_term: str) -> Dict[str, str]:
```
Peedee2002: mypy is angry at you because of this line - you set the return type to be `Dict[str, str]`, but `followups`'s value is actually a dictionary, not a string.

Peedee2002 (changes requested):
few small changes to make, but otherwise cool

backend/server/routers/followups.py
Basialo: Whoops ok I changed the return but forgot to change the declaration, will fix this. (and get mypy)

Peedee2002:
Oh also, sorry to do this to you - tests plz 🙏

backend/server/routers/followups.py
```diff
@@ -0,0 +1,90 @@
...
+ 
+    # we only have enrolment data from T2(2022) -> T3(2022) at this stage
+    origin_term = "T2" # remove this line after we get more data
+    next_term = get_next_term(origin_term)
+
+    # get all comp courses
+    all_comp_courses = filter(lambda course: course.startswith("COMP"), 
+                              [course["code"] for course in coursesCOL.find()])
+    
+    print(all_comp_courses)
```
Peedee2002: Debugging statement
```suggestion
```

backend/server/routers/followups.py
```diff
@@ -0,0 +1,90 @@
...
+def get_next_term(term: str) -> str:
+    # 'S' -> 'T1'
+    # 'T1' -> 'T2'
+    # 'T2' -> 'T3'
+    # 'T3' -> 'S'
+    if term[0] == 'T':
+        if int(term[-1]) < 3:
+            return 'T' + str(int(term[-1]) + 1)
+        return 'S'
+    return 'T1'
```
imagine-hussain: I think the multiple returns here add more branches in the logic that is required;

Since we want the next element of an ordered set, we can hardcode the term strings as
`["S", "T1", "T2", "T3"]`

Then we can just grab the index of the str given and return the next element (modded by the list to allow cycling)

backend/server/routers/followups.py
```diff
@@ -0,0 +1,90 @@
...
+                                "T3": 45,
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+)
+def get_followups(origin_course: str, origin_term: str) -> dict[str, str | dict[str, dict[str, int]]]:
```
imagine-hussain, line 51: For complex return types like this, we should probably start using `typeddicts` but, idk if that's smth we wanna start now or, leave it for a later refactor.

@Peedee2002 thoughts?

backend/server/routers/followups.py
```diff
@@ -0,0 +1,90 @@
...
+    next_term = get_next_term(origin_term)
+
+    # get all comp courses
+    all_comp_courses = filter(lambda course: course.startswith("COMP"), 
+                              [course["code"] for course in coursesCOL.find()])
+    
+    print(all_comp_courses)
+    
+    # get enrolment data
+    with open("./data/final_data/enrolmentData.json") as enrolment_data_file:
```
imagine-hussain:
```suggestion
    with open("./data/final_data/enrolmentData.json", "r", encoding="utf8") as enrolment_data_file:
```

i've had issues w/ python trying to open these files w/ some other encodings before, this just helps keep the consistency w/ the rest of the codebase and, avoid silent bugs like that

imagine-hussain (commented):
very good stuff

just a few minor points i've commented on but otherwise lgtm

backend/server/routers/followups.py
Peedee2002: I'm not too fussed, because this is only used in one place. TypedDicts are more important for complex data types used everywhere

backend/server/routers/followups.py
```diff
@@ -0,0 +1,90 @@
...
+    top_followups = sorted(followups.items(), reverse=True, key=lambda course: course[1][next_term])
+ 
+    return {
+        "originCourse": origin_course,
+        "originTerm": origin_term,
+        "followups": dict(top_followups),
+    }
+
+
+
```
Peedee2002: clean whitespace
```suggestion

```

Peedee2002 (approved):

PR #846, opened 2022-10-18
Circles 349/migrate rest of planner slice

Peedee2002:
got test and implementation for the slices


Peedee2002:
@sixohtoo I choose you!

backend/server/routers/user.py
```diff
@@ -28,12 +29,25 @@ def set_user(token: str, item: Storage, overwrite: bool = False):
 @router.post("/saveLocalStorage/")
 def save_local_storage(localStorage: LocalStorage, token: str = DUMMY_TOKEN):
     # TODO: turn giving no token into an error
-    # THINK: this is wierd security wise, because we are setting it to what we are given no ma
+    planned: list[str] = sum((sum(year.values(), []) for year in localStorage.planner['years']), [])
```
sixohtoo, line 32: Can't vouch for what this line does as I don't understand it, but otherwise everything looks good to me.

sixohtoo (approved):

backend/server/routers/user.py
imagine-hussain: Suppose years looks something like:

```json
 "years": [
                {
                    "T0": [],
                    "T1": [
                        "MATH1241"
                    ],
                    "T2": [
                        "MATH1141"
                    ],
                    "T3": []
                },
                {
                    "T0": [],
                    "T1": [],
                    "T2": [],
                    "T3": []
                },
                {
                    "T0": [],
                    "T1": [],
                    "T2": [],
                    "T3": []
                }
            ],
```

Then, in `year in localstorage['years']` each `year` object is that dict with keys as terms and each `value` is the list of courses in that year.

So `year.values()` is an iterable containing lists of courses.
Well, if we `sum` over `year.values()` then that will add all those lists together.
Then, if we sum over that for all years, we collect all of those lists of courses, into one giant list of courses.


Tldr:

> Gets a list containing every course that is planned

backend/server/tests/user/test_update_start_year.py
```diff
@@ -0,0 +1,48 @@
...
+    user_after = requests.get(f'http://127.0.0.1:8000/user/data/{DUMMY_TOKEN}').json()
+    assert len(user_before['planner']['years']) == len(user_after['planner']['years'])
+    assert user_after['planner']['years'][0] == {"T0": [], "T1": [], "T2": [], "T3": []}
+    assert user_after['planner']['years'][1] == { "T0": [], "T1": [ "COMP1511", "MATH1141", "MATH1081"], "T2": [ "COMP1521", "COMP1531", "COMP2521"], "T3": []}
+    assert user_after['planner']['years'][2] == {"T0": [], "T1": [], "T2": [], "T3": []}
+    
+
+def test_updateCourseMark_back_time_2018():
+    clear()
+    requests.post(
```
imagine-hussain, line 26: we love to see it

imagine-hussain (approved):
can't speak for any of the docker stuff but, lgtm

PR #855, opened 2022-10-21
feat(backend): CIRCLES-382 API for autoplanning

hdphuong:
I hope this is what we want xD 
Status 200 success:
Screen Shot 2022-10-21 at 4 51 15 pm

Status 400 can't plan:
Screen Shot 2022-10-21 at 4 50 47 pm


backend/server/routers/planner.py
```diff
@@ -51,3 +55,23 @@ def validate_term_planner(plannerData: PlannerData):
...
+        raise HTTPException(status_code=400, detail=f"Error: {e}")
+    
+    for year in range(programTime.startTime[0], programTime.endTime[0] + 1):
+        result["plan"].append([])
+        for term in range(0, 4):
+            result["plan"][year - programTime.startTime[0]].append({})
+            for course in autoplanned:
+                print(course)
+                if course[1][0] == year and course[1][1] == term:
+                    result["plan"][year - programTime.startTime[0]][term][course[0]] = [get_course(course[0]).uoc , None]
```
Peedee2002: @gtangelo would you like the output in this format? Or would you like it differently?

backend/server/routers/utility.py
```diff
@@ -32,3 +37,13 @@ def get_core_courses(program: str, specialisations: list[str]):
             if "Major" in spec_name or "Honours" in spec_name)
          , [])
 
+
+def get_course(code: str) -> Course:
```
Peedee2002: I think we can collect this data already using `getTermsOffered` and `getCourse` from our own apis

backend/server/routers/planner.py
gtangelo: Was there any reason on why we would want to return the UOC and mark alonside the course keys? (I'm assuming the second key is for the marks but feel free to correct me). Can't think anything that we might need that for on the top of my head rn.

Either way, dont think the format matters too much since we can just change it on the FE side anyways.

Though would be good to have it same to the format as `years` field in the `plannerSlice` i.e. something like this?
```
"plan": [
  {"T0":[],"T1":["COMP1511","COMP2511"],"T2":["COMP1521"],"T3":["COMP1531"]},
  {"T0":[],"T1":["COMP3900"],"T2":["COMP2521"],"T3":["COMP3121"]},
  {"T0":[],"T1":["MATH1231"],"T2":["MATH1081"],"T3":["COMP4920"]},
  {"T0":[],"T1":["COMP3131"],"T2":[],"T3":["COMP3311"]}
]
```

gtangelo:
Also, seems a bit weird on why `courseCodes` has a nested key layer `courses` when it can be compressed to a single layer? lol Not a biggie tho

Also, what is the difference between `courseCodes.courses` vs `userData.courses` 👀 

Screen Shot 2022-10-21 at 11 26 06 pm

backend/server/routers/planner.py
hdphuong: Yup, I agree that's better, will modify this :D 
The only reason was to re-use PlannerData in model.py. But can create a new model that conforms to the years field in PlannerSlice.

backend/server/routers/utility.py
hdphuong: Yup `getTermsOffered` would be helpful.
But not sure which `getCourse` you mean as `get_course` in `courses.py` returns a Dict with stuff like `is_legacy`, `handbook_note`, which is not helpful to convert to `Course` i think.

hdphuong:
Umm `courseCodes` is already declared in `model.py`, not sure where it's originally used for but we can use a normal list there if that's easier. Also,`courseCodes.courses` are courses to plan and user.courses are courses that user has taken. Pls correct me if that's wrong @Peedee2002

backend/server/routers/utility.py
Peedee2002: Oh yea you already got the uoc some other way nvm. Then getTermsOffered is what we need to call to make this accurate. We also need the FE to give us their mark.

Peedee2002:
> Umm `courseCodes` is already declared in `model.py`, not sure where it's originally used for but we can use a normal list there if that's easier. Also,`courseCodes.courses` are courses to plan and user.courses are courses that user has taken. Pls correct me if that's wrong @Peedee2002

yup, thats correct. We don't currently actually correctly deal with lock-on of courses, but will implement that soon.

Peedee2002 (approved):

PR #857, opened 2022-10-22
fix: max-height for Collapsible component

gtangelo:


frontend/src/App.tsx
```diff
@@ -26,8 +26,8 @@ const App = () => {
   return (
     <ThemeProvider theme={theme === 'light' ? lightTheme : darkTheme}>
       <GlobalStyles />
-      <Suspense fallback={<PageLoading />}>
-        <ErrorBoundary>
```
Peedee2002, line 30: NOOOO dont tell me this was the bug

Peedee2002 (approved):

frontend/src/App.tsx
Peedee2002: https://github.com/csesoc/Circles/commit/384c237ca51d3a812030a91dca4773a2ff4c1b87
😳

Peedee2002: lmao its official that every *single* commit to dev jared made that day introduced a bug

frontend/src/App.tsx
```diff
@@ -39,8 +39,8 @@ const App = () => {
               <Route path="*" element={<Page404 />} />
```
Peedee2002, line 39: if we are looking at this, one thing that annoys me is that 404 is rendered if it loads first, which is irritating. can we improve this routing to not have that happen?

frontend/src/App.tsx
gtangelo: lmaooo i actually dont know if this is the bug or not. but like i dont think it makes sense for the error boundary to be lazy loaded coz its suppose to catch errors. will probs have to monitor this still tbh

frontend/src/App.tsx
gtangelo: yep will fix this eventually haha

Peedee2002: https://stackoverflow.com/questions/54625911/404-page-in-react-renders-in-every-page I think this might help?

PR #860, opened 2022-10-22
feature: cf-359 - Return button for Error Screen

ollibowers:
Added a cute button on `ErrorBoundary` so we don't always have to reset data. 
image

Refactored and reused the `ResetModal` component from the degree-wizard to achieve this. 
**Note**: had to remove the `open` state from it to make it simpler, (and because it wasn't recreating all the time?)
image


frontend/src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -57,9 +71,14 @@ class ErrorBoundary extends React.Component<Props, State> {
               to inform us how the error occurred! Please also include brief description on the
               steps that led up to the error and a copy of the error messages seen below.
```
Peedee2002, line 72: Add info here to explain the difference between the two buttons. Probably something along the lines of "if you are seeing this page often or are seeing circles behave in some way you don't expect, try to reset your data. Otherwise you can return to circles to try again.

ollibowers: Good idea

frontend/src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -31,18 +33,30 @@ class ErrorBoundary extends React.Component<Props, State> {
...
 
   render() {
-    const { hasError, error, errorInfo } = this.state;
+    const { hasError, error, errorInfo, modalOpen } = this.state;
     const { children } = this.props;
 
-    const handleClick = () => {
-      localStorage.clear();
+    const gotoDegrees = () => {
       window.location.href = '/degree-wizard';
```
gtangelo, line 40: nitpick: is it possible/could we use `useNavigate` for the redirection?

frontend/src/components/ResetModal/ResetModal.test.tsx
```diff
@@ -31,20 +31,20 @@ describe('ResetModal', () => {
   });
 
   it('should render', () => {
-    renderWithProviders(<ResetModal />);
+    renderWithProviders(<ResetModal open={false} />);
```
gtangelo: probs can make the `open` prop optional so we dont have to be forced to set it to `false` and instead just don't include the prop at all

frontend/src/components/ResetModal/ResetModal.test.tsx
```diff
@@ -63,12 +63,14 @@ describe('ResetModal', () => {
     ]);
   });
 
-  it('should redirect back to course selector if cancel button is clicked', async () => {
-    const dummyNavigate = vi.fn();
-    useNavigateMock.mockReturnValue(dummyNavigate);
+  // TODO: Disabled since it now takes a onCancel to be reusable
```
gtangelo: Since we are changing the interface of `ResetModal`,  rather than commenting this out, could we change the test? We can have a test case where we check if it calls the `onCancel` callback passed when clicking the cancel button.

Also, if you want, can add another test to see if `onOk` callback passed is called when clicking the confirm button. 

Though let me know if you want me to this bit or not if you're confused on writing FE tests. I can write the tests if you want :D

frontend/src/components/ResetModal/ResetModal.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import { Modal } from 'antd';
+import { useAppDispatch } from 'hooks';
+import { resetCourses } from 'reducers/coursesSlice';
+import { resetTabs } from 'reducers/courseTabsSlice';
+import { resetDegree } from 'reducers/degreeSlice';
+import { resetPlanner } from 'reducers/plannerSlice';
+
+type Props = {
+  open: boolean;
```
gtangelo: could make this optional

gtangelo (changes requested):
lgtm!! Just see the comments :DD

frontend/src/components/ErrorBoundary/ErrorBoundary.tsx
ollibowers: I tried, but since the error boundary is outside the router, it is not possible i dont think?

frontend/src/components/ResetModal/ResetModal.test.tsx
ollibowers: sure, my bad!

frontend/src/components/ResetModal/ResetModal.test.tsx
ollibowers: goooood idea!!!

ollibowers: oki done <3

gtangelo (approved):

PR #861, opened 2022-10-22
fix(backend): circles-388 rerun processors for prev year data

hdphuong:
- autoplanning works for 2019 & 2020 now
- added year to `runprocessors.py` and fixed a bug in `courses_processing.py`
- threw exception in case a course is discontinued


backend/runprocessors.py
```diff
@@ -44,6 +44,12 @@
     help="program, specialisation, course, condition, data-fix, enrolment",
 )
 
+parser.add_argument(
+    "--year",
```
Peedee2002: Not needed, there is processors_archive.py. either use that or delete that file since it will now be redundant

backend/runprocessors.py
```diff
@@ -112,9 +118,9 @@ def run_scrape_enrolment_data():
         "process": customise_spn_data,
     },
     "course": {
-        "scrape": scrape_course_data,
-        "format": format_course_data,
-        "process": process_course_data,
+        "scrape": scrape_course_data(args.year),
```
Peedee2002: Hmm doesn't this get called immediately? I think you would want to make this:

```
lambda : scrape_course_data(args.year)
```

backend/runprocessors.py
hdphuong: Wait we have that? Why didn't u tell me that earlier 😭

backend/server/tests/courses/test_get_legacy_course.py
```diff
@@ -97,6 +97,7 @@ def test_legacy_math3361():
     However, it still has legacy handbook entries in the years that it doesn't run.
     """
     res2019 = get_legacy_course_wrapper(2019, "MATH3361")
+    print(res2019.json())
```
Peedee2002:
```suggestion
```

Peedee2002 (approved):

PR #862, opened 2022-10-22
refactor: no more axios imports!

ollibowers:
I dislike having to remember routes and their types. 
So I aggregated all the API related types and functions, and created a single file containing every endpoint. 
It is all fully type annotated and cross checked with the current backend models. 

When converting over the types, I spent a lot of time going one by one, checking that I wasn't breaking stuff, so hopefully I didn't 🥺. 

Now calls like 
```ts
const payload = JSON.stringify(prepareUserPayload(degree, planner);
axios.post<CoursesUnlockedWhenTaken>(`/courses/coursesUnlockedWhenTaken/${courseCode}`, payload));
```
instead look like 
```ts
API.courses.unlockedWhenTaken(courseCode, prepareUserPayload(degree, planner));
```

Benefits:
- single point of failure for the types
- single point to change types and routes when backend makes changes
- no more forgor 💀 
- no need to cross check api reference, as it is all documented in the file too
- could be easy to write tests?
- ✨ pretty ✨ 

Todo:
- [ ] cross check some types with backend (including the possible `str | str[]`)
- [ ] mayb? replace the try-catch repetition everywhere with a unwrap function
- [ ] mayb? add route documentation to the endpoint types so easy to figure out which you want
- [ ] i had more but i forgot

side note, thoughts on renaming some of the routes? a get request to `/courses/getCourse/{code}` seems to have some redundancy.


olllllli:
woa!! lets get this merged!!!!!

Peedee2002:
I'm sorry to disappoint you when this all gets nuked next year when we refactor the API for data migration, but good work lol

ollibowers:
> I'm sorry to disappoint you when this all gets nuked next year when we refactor the API for data migration, but good work lol

thats ok <3 itll be easy to rebuild

frontend/.prettierignore
```diff
@@ -0,0 +1,2 @@
+/src/utils/api/api.ts
```
gtangelo, line 1: any reason why these are ignored?

frontend/src/utils/api/api.ts
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable */
+
+import axios from 'axios';
+import { APIEndpoints } from './types/endpoints';
+import composeEndpoint from './composeEndpoint';
+
+const API: APIEndpoints = {
```
gtangelo, line 7: Do we need to define a type for this Object? It would be better if we can just infer the type of this object so that in future, we can easily add a new route without touching a type declaration

frontend/src/utils/api/api.ts
```diff
@@ -0,0 +1,45 @@
...
+  planner: {
+    validate: (plannerData) => axios.post('/planner/validateTermPlanner', plannerData),
+  },
+
+  courses: {
+    jsonified: (courseCode) => axios.get(`/courses/jsonified/${courseCode}`),
```
gtangelo, line 13: can we rename the keys for the different endpoints?

Like for example, if the endpoint was `/courses/getCourse/`, the object should be `courses: { getCourse: ..., ... }`

So that its more obvious which api call we are calling i.e. `API.courses.getCourse` over `API.courses.course`

frontend/src/utils/api/api.ts
```diff
@@ -0,0 +1,45 @@
...
+    types: (programCode: string) => axios.get(`/specialisations/getSpecialisationTypes/${programCode}`),
+    specialisations: (programCode: string, type: string) => axios.get(`/specialisations/getSpecialisations/${programCode}/${type}`),
+  },
+
+  default: {
+    liveYear: () => axios.get(`/live_year`),
+  },
+};
+
+export default API;
```
gtangelo, line 45: rather than defining a single object that houses all of the API calls, be preferred if we do something like this
```
export const coursesApi = {
  jsonified: ...
}

export const programsApi = {
...
}

export { 

```

So we dont need to continually chain objects
i.e. `coursesApi.jsonified()` over calling `API.courses.jsonified()`

gtangelo (changes requested):
Some brief comments, will review how it goes once the PR feedback has been resolved. Though I'm not too fused if you want to decline this PR or continue with it 😅

gtangelo:
Closing this PR as discussed with @ollibowers previously.

PR #864, opened 2022-10-23
feat(backend): remove core courses from geneds [CIRCLES-378]

imagine-hussain:
- ensure there is no intersection between the courselist from `/programs/getStructure` and `programs/getGenEds`

- tests will ensure this;

- Added checks to make sure that `KeyErrors` on the scraped geneds doesn't lead to `500` errors but rather, `400` that have actual meaning


backend/server/tests/programs/test_get_programs.py
```diff
@@ -9,3 +9,15 @@ def test_sanity():
...
         assert re.match(r"[0-9]{4}", key)
         assert type(value) is str
+
+def test_sanity_all():
+    x = requests.get('http://127.0.0.1:8000/programs/getAllPrograms')
+    assert x.status_code == 200
+    programs = x.json()['programs']
+
+    # Not actual number, just a rough lower bound
+    assert programs.length > 100
```
Peedee2002: Too much JavaScript

```suggestion
    assert len(programs) > 100
```

backend/server/routers/programs.py
```diff
@@ -303,10 +325,43 @@ def get_structure_course_list(
...
+    ) -> Dict[str, Dict[str, str]]:
+    """
+    fetches gen eds from file and removes excluded courses
+    WARNING:
+        The `excluded_courses` is an optional parameter. This does
+        NOT mean you are free to ignore it. Default behaviour is to generate
+        a list of courses to exclude based on the `getStructure` call
+        for the respective programCode.
+        Be very careful leaving this field empty as `None` unless you are sure
+        you want to auto-generate excluded_courses. This can lead to unexpected
```
Peedee2002: This is confusing behaviour, why do we want to keep this the default?

imagine-hussain: oop, this it outdated docs;
the new behaviour doesn't bootstrap `excluded_course` from `getStructure`, it just assumes that u want an empty list as ur default.

see:

```py
excluded_courses = excluded_courses if excluded_courses is not None else []
```

This was needed before because getStructure also builds geneds but, that's been changed now that getstructure is able to take an `ignore` param so i'll just delete these docs

backend/server/routers/programs.py
```diff
@@ -119,7 +133,7 @@ def add_geneds_courses(programCode: str, structure: dict[str, StructureContainer
     item = structure["General"]["content"]["General Education"]
     item["courses"] = {}
     if container.get("courses") is None:
-        gen_ed_courses = list(set(get_gen_eds(programCode)["courses"].keys()) - set(sum(
+        gen_ed_courses = list(set(get_gen_eds(programCode, [])["courses"].keys()) - set(sum(
```
Peedee2002: This [] is not needed

backend/server/routers/programs.py
```diff
@@ -129,7 +143,9 @@ def add_geneds_courses(programCode: str, structure: dict[str, StructureContainer
             for spec_name, spec in structure.items()
             if "Major" in spec_name or "Honours" in spec_name)
         , [])))
-        geneds = get_gen_eds(programCode)
+        # clist = course_list_from_structure(structure)
```
Peedee2002: Delete these lines

Peedee2002 (approved):

PR #872, opened 2022-10-29
cf-348: Locked courses in your planner should be shown in the course menu when the locked courses toggle is off

kellieyau:
Thought I closed this, 
image


Peedee2002:
What does the toggle do now?

gtangelo (approved):

PR #875, opened 2022-10-31
feat: legacy offering support for term planner

### LINKED ISSUE #831: course is available but says it is not
gens1111 is available 22t3, says unavailable

ollibowers:
### WOOOOO!!!!!!!!!!!! 🥳 🥳 🥳 🥳 🥳 🥳 🥳 🥳 

# Added support (warnings and green boxes) for past offerings of courses in the term planner

https://user-images.githubusercontent.com/80164276/198997419-e649b7a6-79bf-48be-8598-f5969655cd20.mp4

### What was changed / Notes:

- Added a `legacyOfferings` field to the `PlannerCourse` in the `planner.courses` slice
- This key is undefined by default, and is not set when courses are initially added to planner (since it comes from a different route)
- When the term planner page loads, it checks all courses in the planner slice, and fetches this data for any that are not already stored, or don't have the correct years required
- Note: when the backend returns null for a year's offerings, it will replace this with the most recent non null offerings (so it assumes the future), or no offerings if there isn't any (can't assume the past)


Peedee2002:
Poggers poggers poggers poggers

frontend/src/pages/TermPlanner/TermBox/TermBox.tsx
```diff
@@ -17,12 +19,12 @@ const Droppable = React.lazy(() =>
 type Props = {
   name: string;
   coursesList: string[];
-  termsOffered: string[];
-  dragging: boolean;
+  draggingCourse?: string | undefined;
```
gtangelo: dont need to make this have `| undefined` attached since the optional param makes is also be undefined when using it as a prop?

spanishpear:
THIS IS VERY POG OLLI I have been complaining about this since day 1 (sorry @Peedee2002 )

spanishpear:
THIS IS VERY POG OLLI I have been complaining about this since day 1 (sorry @Peedee2002 )

frontend/src/reducers/plannerSlice.ts
```diff
@@ -265,6 +265,23 @@ const plannerSlice = createSlice({
...
+    ) => {
+      const { code, offerings } = action.payload;
+
+      if (state.courses[code]) {
+        if (state.courses[code].legacyOfferings === undefined) {
+          state.courses[code].legacyOfferings = {};
+        }
+
+        Object.entries(offerings).forEach(([year, terms]) => {
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
gtangelo, line 280: avoid doing eslint disable next line?

frontend/src/pages/TermPlanner/TermPlanner.tsx
```diff
@@ -34,11 +36,40 @@ const TermPlanner = () => {
...
+      const data = await getAllCourseOfferings(outdated, years);
+      Object.entries(data).forEach(([code, offerings]) => {
+        dispatch(updateLegacyOfferings({ code, offerings }));
+      });
+    };
+
+    updateOfferingsData();
+    // hacky solution to useEffect deps, because it cant compare arrays well enough :c
+    // also only want it to update on new/removed courses, not altered inner state
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
gtangelo, line 76: Is it possible to avoid disabling this eslint error? If it triggers useEffect often, try using useCallback/useMemo. Lmk if it isnt

frontend/src/pages/TermPlanner/TermPlanner.tsx
```diff
@@ -34,11 +36,40 @@ const TermPlanner = () => {
   const planner = useSelector((state: RootState) => state.planner);
   const degree = useSelector((state: RootState) => state.degree);
 
-  const [termsOffered, setTermsOffered] = useState<Term[]>([]);
-  const [isDragging, setIsDragging] = useState(false);
+  const [draggingCourse, setDraggingCourse] = useState<string | undefined>();
```
gtangelo: can we set for the initial state to be an empty string or null rather than undefined?

frontend/src/pages/TermPlanner/TermPlanner.tsx
```diff
@@ -217,15 +246,14 @@ const TermPlanner = () => {
...
-                            dragging={isDragging}
+                            draggingCourse={draggingCourse}
                           />
                         );
                       })}
                     </React.Fragment>
                   );
                 })}
-                <UnplannedColumn dragging={isDragging} />
+                <UnplannedColumn dragging={draggingCourse !== undefined} />
```
gtangelo: could do `!draggingCourse`?

frontend/src/pages/TermPlanner/TermPlanner.tsx
```diff
@@ -34,11 +36,40 @@ const TermPlanner = () => {
...
+      for (let i = 0; i < planner.numYears; i++) {
+        years.push((planner.startYear + i).toString());
+      }
+
+      // get the outdated legacy offerings
+      const outdated = Object.entries(planner.courses)
+        .filter(([, course]) => {
+          const oldOfferings = course.legacyOfferings;
+          // no legacy offering data, or doesn't include the wanted years
+          return !oldOfferings || years.filter((y) => !(y in oldOfferings)).length > 0;
```
gtangelo, line 62: Ooo i like this approach! Doesn't need us to do a migration script for existing data already! :D

gtangelo (changes requested):
everything lgtm!! Liked the approach taken to support previous years offerings of courses. Left some style comments mainly. Happy to be merged in once those are addressed 🥳

frontend/src/utils/getAllCourseOfferings.ts
```diff
@@ -0,0 +1,59 @@
...
+        const res = await axios.get<APIResult>(`/courses/termsOffered/${course}/${yearsStr}`);
+        if (res.status === 200) {
+          let mostRecentYear: Term[] = [];
+          offers[course] = {};
+          Object.entries(res.data.terms).forEach(([year, terms]) => {
+            offers[course][year] = terms ?? mostRecentYear;
+            if (terms !== null) {
+              mostRecentYear = terms;
+            }
+          });
```
imagine-hussain: The route has been updated to use the `LIVE_YEAR` as a fallback  - which already isn't perfect but, it's something to keep in mind. the difference between this and just using the be's implementation is that this will also predict the future for courses that aren't being offered in this term. I feel that perhaps, we should add a future card to detect if a course has been offered inconsistently and just throw up a tooltip / warning for those courses.

Thoughts?

imagine-hussain (approved):
Very noice

frontend/src/pages/TermPlanner/TermPlanner.tsx
ollibowers: <3

frontend/src/utils/getAllCourseOfferings.ts
ollibowers: i like

frontend/src/reducers/plannerSlice.ts
ollibowers: I think it has to be there or I get warnings? I think the warnings are wrong tho because I initialize that value 2 lines above. Thoughts?

gtangelo (approved):

PR #880, opened 2022-11-04
feat(Allow multiterm courses to be dragged simultaneously)

Uzaif-Sheikh:
Added Badge to show the instance when dragging a multi-term course.
ezgif com-gif-maker (8)


frontend/src/pages/TermPlanner/UnplannedColumn/UnplannedColumn.tsx
```diff
@@ -31,9 +32,29 @@ const UnplannedColumn = ({ dragging }: Props) => {
               droppable={dragging}
               isSmall={isSmall}
             >
-              {unplanned.map((course, courseIndex) => (
-                <DraggableCourse code={course} index={courseIndex} key={course} term="" />
```
gtangelo, line 35: could simply add `isDragged={planner.courses[course].isMultiterm}` to the prop rather than needing a if statement?

frontend/src/pages/TermPlanner/DraggableCourse/DraggableCourse.tsx
```diff
@@ -117,32 +120,43 @@ const DraggableCourse = ({ code, index, term }: Props) => {
...
-                      {/*
-                      Marks can be strings (i.e. HD, CR) or a number (i.e. 90, 85).
-                      Mark can be 0.
-                    */}
-                      {typeof mark === 'string' || typeof mark === 'number' ? mark : 'N/A'}
-                    </Text>
-                  </div>
-                )}
-              </S.CourseLabel>
+              <Badge
```
gtangelo: can we override it via styled components? Also, I notice some weird issue with the `line-height`. Could you fix this (the issue with the line height comes from the css in `Badge` since we are wrapping it within the course label. Either override the `line-height` css property of do an absolute position with the badge (similar to the uoc badge)

frontend/src/pages/TermPlanner/DraggableCourse/DraggableCourse.tsx
```diff
@@ -3,28 +3,31 @@ import { useContextMenu } from 'react-contexify';
...
 import useMediaQuery from 'hooks/useMediaQuery';
 import ContextMenu from '../ContextMenu';
+import { getNumTerms } from '../utils';
 import S from './styles';
 
 type Props = {
   code: string;
   index: number;
   term: string;
+  isDragged: boolean;
```
gtangelo: can we rename this to `showMultiCourseBadge` or smth like that rather than `isDragging` for better semantics?

frontend/src/pages/TermPlanner/TermBox/TermBox.tsx
```diff
@@ -82,9 +83,29 @@ const TermBox = ({ name, coursesList, termsOffered, dragging }: Props) => {
               ref={provided.innerRef}
               {...provided.droppableProps}
             >
-              {coursesList.map((code, index) => (
-                <DraggableCourse key={`${code}${term}`} code={code} index={index} term={term} />
-              ))}
+              {coursesList.map((code, index) => {
+                if (multiCourseDrag !== '' && multiCourseDrag === code) {
```
gtangelo: can remove this if statement and just pass in `isDragged={multiCourseDrag === code`

frontend/src/pages/TermPlanner/TermPlanner.tsx
```diff
@@ -212,13 +220,26 @@ const TermPlanner = () => {
...
+                          );
+                        }
                         return (
                           <TermBox
                             key={key}
                             name={key}
                             coursesList={year[term as Term]}
                             termsOffered={termsOffered}
                             dragging={isDragging}
+                            multiCourseDrag=""
```
gtangelo: could just remove the if statement and simply pass in as `multiCourseDrag={checkYearMultiTerm && checkYearMultiTerm === key}`?

frontend/src/pages/TermPlanner/TermPlanner.tsx
```diff
@@ -36,6 +36,8 @@ const TermPlanner = () => {
 
   const [termsOffered, setTermsOffered] = useState<Term[]>([]);
   const [isDragging, setIsDragging] = useState(false);
+  const [checkYearMultiTerm, setCheckYearMultiTerm] = useState('');
```
gtangelo, line 39: do we need to have the `checkYearMultiTerm` state? We can simply reuse the `multiCourse` state as a check to show the badge instead? Both state use cases seems to be the same application?

frontend/src/pages/TermPlanner/TermBox/TermBox.tsx
```diff
@@ -19,9 +19,10 @@ type Props = {
   coursesList: string[];
   termsOffered: string[];
   dragging: boolean;
+  multiCourseDrag: string;
```
gtangelo: we can just call this smth like `showMultiCourseBadge` and make this a boolean instead rather than the need to compare it as a course code string?

gtangelo (changes requested):
just some styled comments. but otherwise lgtm!! :DD

frontend/src/pages/TermPlanner/TermPlanner.tsx
Uzaif-Sheikh: So, I can't remove the `checkYearmultiTerm` since `checkYearmultiTerm` helps to check which draggable course in the term box we are dragging and then adds a badge to that course. If I remove `checkYearmultiTerm` it will add a badge to all the courses in that term box.

gtangelo (approved):

PR #893, opened 2022-11-12
CIRCLES-386

hdphuong:
- Lock courses in by "forcing" term offerings
- I TRIED USING ORTOOLS (yes 5 hours well spent) but couldn't figure out how to do that in a nice way given the data structures we have so i went back to my original solution 😭

### DIFF
--- a/backend/server/routers/planner.py
+++ b/backend/server/routers/planner.py
@@ -5,10 +5,7 @@ from typing import Tuple, List, Dict
 from fastapi import APIRouter, HTTPException
 from algorithms.validate_term_planner import validate_terms
 from algorithms.autoplanning import autoplan
-from algorithms.objects.user import User
-from algorithms.objects.course import Course
-from server.routers.model import (ValidCoursesState, PlannerData, ValidPlannerData, 
-                                CourseCodes, UserData, ProgramTime)
+from server.routers.model import (ValidCoursesState, PlannerData, ValidPlannerData, ProgramTime)
 from server.routers.courses import get_course
 from server.routers.utility import get_course_object, extract_user_from_planner_data
 
--- a/backend/server/routers/planner.py
+++ b/backend/server/routers/planner.py
@@ -10,7 +10,7 @@ from algorithms.objects.course import Course
 from server.routers.model import (ValidCoursesState, PlannerData, ValidPlannerData, 
                                 CourseCodes, UserData, ProgramTime)
 from server.routers.courses import get_course
-from server.routers.utility import get_course_object
+from server.routers.utility import get_course_object, extract_user_from_planner_data
 
 def fix_planner_data(plannerData: PlannerData) -> ValidPlannerData:
     """ fixes the planner data to add missing UOC info """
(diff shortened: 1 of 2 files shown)


backend/server/routers/planner.py
```diff
@@ -91,11 +88,24 @@ def validate_term_planner(plannerData: PlannerData):
...
+def autoplanning(courseCodes: List, plannerData: PlannerData, programTime: ProgramTime) -> Dict:
+    user = extract_user_from_planner_data(plannerData)
+    autoplanned = []
     try:
+        # Convert list of course codes to list of Course objects 
         courses = [get_course_object(courseCode, programTime) for courseCode in courseCodes]
+
+        # Convert plannerData to list of Course object and FORCE the term the course is taken in
+        # e.g. if the user has taken COMP1511 in 2022 T1, then the autoplanning will have COMP1511 in 2022 T1 
+        # regardless of the actual COMP1511 term offerings
```
Peedee2002, line 100: wait actually i forgot about this problem... i think my solution would have forced a course placed in the wrong term to always be infeasable 😅

backend/server/routers/utility.py
```diff
@@ -48,18 +48,32 @@ def get_course_object(code: str, progTime: ProgramTime) -> Course:
...
+    """
+    Extracts a user object from the plannerData
+    """
+    user = User()
+    user.program = plannerData.program
+    user.specialisations = plannerData.specialisations
+    for year_index, year in enumerate(list(plannerData.plan)):
+        for term in year:
+            user.add_courses(term) 
+    return user
```
Peedee2002, line 79: extra whitespace 🔫

Peedee2002 (commented):
few small nitpicks

backend/server/routers/planner.py
hdphuong: ooo what's your solution? 👀

Peedee2002: basically figure out what the index the user set that would correspond to the number in the model, then hardcode it
```python
i = course.getIndexPlaced() # assume this produces like 0 if they placed it in summer term first year or 4 if its t3 first year
model.Add(map_variable_from_course(course ,variables)== i)
```
... then let the rest of the algo run normally. This lets the model see that the variable is there, but doesnt give it the flexibility to move it and is instead forced to work around it.

backend/server/routers/planner.py
```diff
@@ -91,11 +88,24 @@ def validate_term_planner(plannerData: PlannerData):
         }
     }
 )
-def autoplanning(courseCodes: List, userData: UserData, programTime: ProgramTime) -> Dict:
-    user = User(dict(userData))
-
+def autoplanning(courseCodes: List, plannerData: PlannerData, programTime: ProgramTime) -> Dict:
```
Peedee2002, line 91: also, please add tests :)

Peedee2002: also i just noticed - shouldnt this be a POST? I know we are not actually changing anything, but we probably want to let the FE use the body to make a request.

backend/server/routers/planner.py
hdphuong: looks neat, let me steal that real quick xD

jks i think we should keep the current one since it allows past terms to be weird, also i'm not a fan of having routers data model imported to algorithms code and we need that O(n^3) conversion loop to get courses marks anyway. Oops forgot to do so xD

hdphuong:
sorry 🙏💀

PR #907, opened 2022-11-20
Feat(Landing Page cf-382): Footer

kellieyau:
image

image


frontend/src/pages/LandingPage/KeyFeaturesSection.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import drag from '../../assets/LandingPage/KeyFeatures/dragImg.png';
+import flag from '../../assets/LandingPage/KeyFeatures/flag.png';
+import padlock from '../../assets/LandingPage/KeyFeatures/openPadlock.png';
```
zax-xyz: thoughts on using absolute imports here (`'assets/...'`) so you don't have to find your way back to the root?

frontend/src/pages/LandingPage/styles.ts
```diff
@@ -0,0 +1,282 @@
...
+  gap: 15%;
+  flex-wrap: wrap;
+  margin: 15% 5%;
+`;
+
+const InteractiveViewTextContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  width: 370px;
+  viewBox="0 0 723 505";
```
zax-xyz: should this be here?

frontend/src/pages/LandingPage/styles.ts
```diff
@@ -0,0 +1,282 @@
...
+  margin-bottom: 0;
+  width: 100%;
+  background: -webkit-linear-gradient(rgba(145, 84, 222, 0) 160px, rgba(145, 84, 222, 1) 10px);
+  @media all and (max-width: 450px) {
+    background: -webkit-linear-gradient(rgba(145, 84, 222, 0) 5%, rgba(145, 84, 222, 1) 10px);
+  }
+`;
+
+const FooterContentContainer = styled.div`
+  position: 'absolute';
```
zax-xyz: sussy quotes

zax-xyz (commented):
ඞ

frontend/src/pages/LandingPage/KeyFeaturesSection.tsx
kellieyau: Oh very true

frontend/src/pages/LandingPage/styles.ts
kellieyau: as in should it be in styles.ts or should it be in its own styles like in a folder for footer?
If the latter ive moved them to separate files

frontend/src/pages/LandingPage/styles.ts
kellieyau: very sussy indeed

frontend/src/pages/LandingPage/styles.ts
zax-xyz: as in the `viewBox="0 0 723 505"` (i don't think this is valid css?)

kellieyau: Oh is there a reason y it works? 

(I can just put it with the svg)

zax-xyz: 🤔 are you sure it works? it doesn't seem to do anything for me (and you seem to have since removed it anyway)

kellieyau: Oh lol nvm then... Yeh I think I added it to the svg anyway so yeh. Defs not valid CSS, thanks for the pick up!

gtangelo (approved):

PR #910, opened 2022-11-21
refactor(backend) stop accessing ALL_COURSES as global

imagine-hussain:
doing a `if _ is None` check before accessing a global is cringe


Peedee2002 (approved):
I hate this still, but better lol

imagine-hussain:
> I hate this still, but better lol

agreed. there really should be some sort of actual startup process (soon:tm: ?)
also, rust has broken my brain. the fact that these globals aren't immutable haunts me

PR #911, opened 2022-11-21
feat(backend): add filter route for courses

imagine-hussain:
- add route to allow filtering courses
- Not used (for now) - *may* be useful if adding filters to the search
- tbh, this is mostly just for my own degree planning :)

- in future: add support for `OR` conditions

### DIFF
--- a/backend/server/routers/courses.py
+++ b/backend/server/routers/courses.py
@@ -173,7 +173,7 @@ def get_course(courseCode: str) -> Dict:
     return result
 
 @router.get(
-    "filterCourses/{filters}",
+    "/filterCourses/{filters}",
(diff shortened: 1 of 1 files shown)


imagine-hussain:
also i may have pulled #910 changes first, which is why u see those commits here aswell

Peedee2002 (changes requested):
I dont think this should be implemented in the BE. The FE is more than capable of doing this itself, and this is better implemented as an actual `.filter` in the FE with some small UI tweaks.

The best implementation of this feature would be to use first class functions, which is obviously not easy over a network (without having log4j happen again).

imagine-hussain:
> 

fair enough

>

PR #943, opened 2022-12-18
chore: update codeowners to 2023

gtangelo:


Peedee2002 (commented):
Hmm shouldn't we just give them both codeowners for both files?

gtangelo:
> Hmm shouldn't we just give them both codeowners for both files?

They can do that themselves if they want lol. This PR is more of removing us as codeowners so that we dont get notifs xd

Peedee2002 (approved):

PR #953, opened 2022-12-28
Peter/circles 381 add tests for autoplanning

Peedee2002:
ive added tests for autoplanning itself, but ideally i would also add testing for the API route since it has become significantly complex. LMK if you want me to add tests or if you are bothered to @imagine-hussain or @leonardo-fan :)

Ill probably add tests either today if i am bothered or later this week if i am not


.github/workflows/ci.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           echo "VITE_BACKEND_API_BASE_URL=http://localhost:8000/" >> env/frontend.env
           docker compose run ci-backend
       - name: Run mypy checks
-        run: python -m mypy ./backend
+        run: cd backend && python -m mypy . && cd ..
```
imagine-hussain, line 45: Is that `cd ..` needed?

Also, if the mypy has errors, this won't go up a directory and try to run everything else from inside `backend`

backend/algorithms/objects/user.py
```diff
@@ -25,8 +25,8 @@ class User:
     def __init__(self, data = None):
         # Will load the data if any was given
         self.courses: dict[str, Tuple[int, int | None]] = {}
-        self.cur_courses: list[str, Tuple[int, int]] = []  # Courses in the current term
-        self.program: str = None
+        self.cur_courses: dict[str, Tuple[int, int | None]] = {}
+        self.program: str | None = None
```
imagine-hussain: does `Optional` break anything? If not, then prefer that over union with `None`

backend/algorithms/objects/user.py
```diff
@@ -43,26 +43,26 @@ def add_courses(self, courses: dict[str, Tuple[int, int | None]]):
...
-        self.cur_courses.append(course)
+        self.cur_courses[course_code] = course
 
     def add_current_courses(self, courses: dict[str, Tuple[int, int | None]]):
         """
         Takes in a list of courses (represented as strings by course
         code) and, adds it to the list of current courses.
         """
-        self.cur_courses.extend(courses)
+        self.cur_courses = self.cur_courses | courses
```
imagine-hussain, line 58: TIL

imagine-hussain (changes requested):

.github/workflows/ci.yml
Peedee2002: Yea it's needed to switch dirs to pick up my config. Doesn't really matter that it will not run cd .. because if it fails, the pipeline aborts

imagine-hussain: ahhh, understandable

imagine-hussain (approved):
oki this lgtm, shall update more over the year

backend/algorithms/cache/cache.py
```diff
@@ -113,12 +115,16 @@ def tokeniseSchool(School):
             match_object = re.search("(?<=School\sof\s)[^\s\n\,]+", School)
         elif re.search("^(UC)", School):
             match_object = re.search("(?<=UC\s)[^\s\n\,]+", School)
+            if match_object is None:
```
imagine-hussain, line 118: the tempation to throw a walrus operator here :)

backend/server/routers/utility.py
```diff
@@ -39,27 +41,43 @@ def get_core_courses(program: str, specialisations: list[str]):
...
+    user = User()
+    user.program = plannerData.program
+    user.specialisations = plannerData.specialisations
+    for _, year in enumerate(list(plannerData.plan)):
+        for term in year:
+            cleaned_term = {}
+            for course_name, course_value in term.items():
+                cleaned_term[course_name] = (course_value[0], course_value[1]) if course_value else (COURSES[course_name]["UOC"], None)
+            user.add_courses(cleaned_term)
+    return user
```
imagine-hussain: if possible, this would ideally just be a method on `PlannerData`

lhjt (approved):

backend/server/routers/utility.py
imagine-hussain: update: not possible because cyclic imports :(

will just revert to this and then add to backlog to fix this

imagine-hussain: update 2: nvm its fixable with the import inside the method :)

leonardo-fan (approved):
lgtm from the fe side :)

PR #956, opened 2022-12-31
feat: configure landing page routing

gtangelo:
Changes made:
- made landing page as root route and fix references to load up correct page depending on context (i.e. first visit will redirect user to the landing page, existing users will be taken to course selector).
- This PR once merged, will make the landing page publicly availiable in production
- Added a 'Get Involved' section to fill whitespace between 'How it works' section and footer
image

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -35,13 +35,10 @@ const App = () => {
               <Route
                 path="/"
                 element={
-                  <Navigate
-                    to={!degree.isComplete ? '/degree-wizard' : '/course-selector'}
-                    replace
-                  />
+                  !degree.isComplete ? <LandingPage /> : <Navigate to="/course-selector" replace />
                 }
               />
-              <Route path="/landing-page" element={<LandingPage />} />
+              <Route path="/" element={<LandingPage />} />
               <Route path="/degree-wizard" element={<DegreeWizard />} />
               <Route path="/course-selector" element={<CourseSelector />} />
               {inDev && <Route path="/graphical-selector" element={<GraphicalSelector />} />}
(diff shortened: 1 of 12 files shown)


frontend/src/pages/LandingPage/LandingPage.tsx
```diff
@@ -9,8 +11,9 @@ const LandingPage = () => (
   <>
     <Hero />
     <KeyFeaturesSection />
-    <InteractiveViewSection />
+    {inDev && <InteractiveViewSection />}
```
gtangelo, line 14: making this as an in dev feature atm since it has references to graphical selector which is not ready to be released

imagine-hussain:
looks very good, will review further soon:tm:

frontend/src/App.tsx
```diff
@@ -35,13 +35,10 @@ const App = () => {
...
                 element={
-                  <Navigate
-                    to={!degree.isComplete ? '/degree-wizard' : '/course-selector'}
-                    replace
-                  />
+                  !degree.isComplete ? <LandingPage /> : <Navigate to="/course-selector" replace />
                 }
               />
-              <Route path="/landing-page" element={<LandingPage />} />
+              <Route path="/" element={<LandingPage />} />
```
leonardo-fan: can just remove this since the "/" path is already defined above (dab)

frontend/src/pages/LandingPage/LandingPage.tsx
leonardo-fan: ye betto (whip)

leonardo-fan (changes requested):
Very poggers, just one tiny comment about the routes! (nae nae)

frontend/src/App.tsx
gtangelo: oh lmao good catch 😎

leonardo-fan (approved):
LGTM! the final pr (or so you think ;))

PR #962, opened 2023-02-02
bug(backend) fix unhashable json in conditions.py

imagine-hussain:
- this causes crash because cannot push a dict inside a set since it is unhashable
- will occasionally break the jsonified route
- simple fix: remove the extraneous set of brackets


leonardo-fan (approved):
looks good to me :D

Issue #965, opened 2023-02-08
course progression disappeared


circles-slave:
course progression disappeared L + ratio skill diff

PR #971, opened 2023-02-14
fix: menu on navbar

ollibowers:
subpage menu now should use entire width given in navbar. also made sure the logo gets squished regardless of how many buttons we may add. 

<3


Peedee2002:
Pics?

imagine-hussain:
Fixes #965

ollibowers:
> Pics?

picture of it working
image

picture of the menu getting full width
image

picture of @Peedee2002 
image

leonardo-fan (approved):
looks great to me, big props for picking this up!

PR #975, opened 2023-02-20
chore: Made vscode use mypy from the Python extension

Bahnschrift:
Previously we used `"mypy.configFile": "./backend/mypy.ini",` in `.vscode/settings.json` to point the [vscode mypy extension](https://marketplace.visualstudio.com/items?itemName=matangover.mypy) at the mypy config file in the backend directory. However, the normal Python extension has since added support for mypy. This should enable that and point it at the config file.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,10 +5,13 @@
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
     "python.testing.cwd": "${workspaceFolder}/backend",
-    "mypy.configFile": "./backend/mypy.ini",
     "files.exclude": {
         "**/__pycache__": true,
         "**/.hypothesis": true,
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -5,10 +5,13 @@
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
     "python.testing.cwd": "${workspaceFolder}/backend",
-    "mypy.configFile": "./backend/mypy.ini",
```
imagine-hussain: it should be fine to leave this in for the ppl that use the mypy extension anyways

Bahnschrift: Alr lemme resurrect it quickly.

imagine-hussain (approved):
looks good

PR #977, opened 2023-02-27
chore Feb 28 sync with dev for CIRCLES-331 - data migration

imagine-hussain:


imagine-hussain:
pylint fails but... tbh thats fine for now

imagine-hussain:
Autoplanning also broken but, to be honest, expected given how many changes there have been. I'm surprised thats the *only* thing that broke and everything else was fine. I think might just allow auto-planning to be broken for now, since those routes need to be changed anyways for the new users.

PR #979, opened 2023-03-01
bug(router/pathname): revert root navigate to course selector

leonardo-fan:
due to dependencies on path name in course selector that forgot to check before oops  

fixes this 
image

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,6 +1,6 @@
 import React, { Suspense } from 'react';
 import { useSelector } from 'react-redux';
-import { BrowserRouter as Router, Route, Routes } from 'react-router-dom';
+import { BrowserRouter as Router, Navigate, Route, Routes } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


imagine-hussain (approved):
i dont understand why this change causes that break but, lgtm

leonardo-fan:
> i dont understand why this change causes that break but, lgtm

In `CourseDescriptionPanel.tsx` and `CourseAttributes.tsx` there's a reference to pathname that only shows the sidebar if the url has `course-selector`. This is because the Graphical Selector also uses the same components but doesn't want a sidebar since it takes too much space

imagine-hussain:
> > i dont understand why this change causes that break but, lgtm
> 
> In `CourseDescriptionPanel.tsx` and `CourseAttributes.tsx` there's a reference to pathname that only shows the sidebar if the url has `course-selector`. This is because the Graphical Selector also uses the same components but doesn't want a sidebar since it takes too much space

makes sense

PR #980, opened 2023-03-01
chore(docker-compose): fix dependency to run init-mongo on the database

Peedee2002:


leonardo-fan (approved):
docker go boom thank lgtm

PR #982, opened 2023-03-01
feat: Replace removeCourse slice with API

VinayakFia:


frontend/src/components/CourseCartCard/CourseCartCard.tsx
```diff
@@ -23,6 +23,18 @@ const CourseCartCard = ({ code, title }: Props) => {
     dispatch(addTab(code));
   };
 
+  const handleConfirm = async () => {
+    try {
+      const res = axios.post(
+        `/removeCourse`,
+        JSON.stringify({courseCode: code})
+      );
```
imagine-hussain: you'd need to get the token using `useSelector` and then pass that into the request

frontend/src/reducers/plannerSlice.ts
```diff
@@ -274,7 +275,7 @@ const plannerSlice = createSlice({
         const { plannedFor } = state.courses[action.payload];
 
         // remove course
-        delete state.courses[action.payload];
+        delete state.courses[action.payload];
```
imagine-hussain: nitpick: whitespace gotta go

frontend/src/components/CourseCartCard/CourseCartCard.tsx
```diff
@@ -17,6 +18,17 @@ type Props = {
...
+
+  const handleConfirm = () => {
+    try {
+      axios
+        .post(`/planner/removeCourse`, JSON.stringify({ courseCode: code }), { params: { token } });
+    } catch (e) {
+      // esling-disable-next-line no-console
+      console.error('Error at removeCourse', e);
+    }
+  };
```
leonardo-fan: async await plzz

leonardo-fan: also in instead of logging the error you can show a pop up notification saying that there was an error so the user knows that the course didn't get removed properly

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.tsx
```diff
@@ -16,12 +18,22 @@ type Props = {
...
+  const handleDelete = () => {
+    try {
+      axios.post(`/planner/removeCourse`, JSON.stringify({ courseCode: code }), {
+        params: { token }
+      });
+    } catch (e) {
+      // esling-disable-next-line no-console
+      console.error('Error at removeCourse', e);
+    }
+  };
```
leonardo-fan: async await plz, and same thing with the notification

frontend/.prettierrc
```diff
@@ -3,5 +3,6 @@
     "tabWidth": 2,
     "printWidth": 100,
     "singleQuote": true,
-    "trailingComma": "none"
+    "trailingComma": "none",
+    "endOfLine": "auto"
```
leonardo-fan: all good if we leave this as `"lf"` (or remove it, since this is the default anyway) to ensure consistency?

leonardo-fan (changes requested):
kewl kewl seems good, just the comments and can you also fix the diff with the `backend/Scripts` and `package-lock.json`?

frontend/.prettierrc
imagine-hussain: Should be kept as `"lf"` and then change own vscode settings to add a `lf` instead of `crlf`

imagine-hussain (changes requested):
Looking good. But, can you add all the `backend/Scripts` that stuff that's been added into the `.gitignore`?

Then feel free to re-request a review

frontend/src/pages/TermPlanner/ContextMenu/ContextMenu.tsx
```diff
@@ -19,19 +20,31 @@ type Props = {
...
     }
   };
+  const handleUnschedule = () => {
+    dispatch(
+      unschedule({
+        code,
+        destIndex: null
+      })
+    );
+  };
```
imagine-hussain, line 47: I don't think we'll still be needing the `dispatch` here after the react-query migration. If there was no BE work needed on this branch, then can just take this branch directly to react-query

PR #983, opened 2023-03-01
feat: Added pseudo routes for backend based on FE degree slices

ui-Auxilary:
Please check I'm on the right track
For reset, should i clear everything? Or reset to a default CS_USER

Also seems like for setProgram, the programName from FE isn't being used ^^


backend/server/routers/user.py
```diff
@@ -147,3 +147,46 @@ def update_degree_length(numYears: int, token: str = DUMMY_TOKEN):
...
+    user = get_user(token);
+    
+    user = default_cs_user();
+    user['user'] = {
+        'degree': {
+            'programCode': '',
+            'specs': [],
+            'isComplete': False,
+        },
+        'planner': planner,
```
imagine-hussain, line 198: i think this `planner` variable on the rhs is not defined

imagine-hussain: check out this definition in `model.py` for what to instantiate:

```py
class PlannerLocalStorage(TypedDict):
    mostRecentPastTerm: MostRecentPastTerm
    unplanned: list[str]
    startYear: int
    isSummerEnabled: bool
    years: list[dict[str, list[str]]]
    courses: dict[str, dict]
```

backend/server/routers/user.py
```diff
@@ -147,3 +147,46 @@ def update_degree_length(numYears: int, token: str = DUMMY_TOKEN):
...
+@router.put("/removeSpecialisation") 
+def removeSpecialisation(token: str = DUMMY_TOKEN, specialisation: str):
+    user = get_user(token);
+    specs = user['user']['degree']['specs']
+    if specialisation in specs:
+        user['user']['degree']['specs'] = specs.remove(specialisation)
+    set_user(token, user, True)
+
+@router.put("/setIsComplete") 
+def setIsComplete(token: str = DUMMY_TOKEN, isComplete: bool):
```
imagine-hussain: in python, non-default arguments cant follow default arguments. so this would need to be a bit more like
```
def setIsComplete(isComplete: bool, token: str = DUMMY_TOKEN):
```

backend/server/routers/user.py
```diff
@@ -147,3 +147,46 @@ def update_degree_length(numYears: int, token: str = DUMMY_TOKEN):
     else:
         user['planner']['years'] = user['planner']['years'][:diff]
     set_user(token, user, True)
+
+@router.put("/setProgram") 
+def setProgram(token: str = DUMMY_TOKEN, degree: dict):
```
imagine-hussain: non-default after default arg

backend/server/routers/user.py
```diff
@@ -147,3 +147,46 @@ def update_degree_length(numYears: int, token: str = DUMMY_TOKEN):
...
+    user = get_user(token);
+    user['user']['degree']['programCode'] = degree['programCode']
+    set_user(token, user, True)
+
+@router.put("/addSpecialisation") 
+def addSpecialisation(token: str = DUMMY_TOKEN, specialisation: str):
```
imagine-hussain: non-default after default arg

backend/server/routers/user.py
imagine-hussain: if we're just using the `programCode` out of the degree dict, can have this accept that as a string instead

backend/server/routers/user.py
```diff
@@ -147,3 +147,46 @@ def update_degree_length(numYears: int, token: str = DUMMY_TOKEN):
...
+    user = get_user(token);
+    user['user']['degree']['specs'].append(specialisation)
+    set_user(token, user, True)
+
+@router.put("/removeSpecialisation") 
+def removeSpecialisation(token: str = DUMMY_TOKEN, specialisation: str):
+    user = get_user(token);
+    specs = user['user']['degree']['specs']
+    if specialisation in specs:
+        user['user']['degree']['specs'] = specs.remove(specialisation)
```
imagine-hussain: `.remove` on lists is in-place and returns None so, this is sufficient as

```suggestion
        specs.remove(specialisation)
```

imagine-hussain (commented):
pretty good job so far!

there's just a few python syntax / type errors to look out for - getting `mypy` set up on your editor can really help catch these early. other than the stuff in the review, i don't see much of a problem

also for:
> For reset, should i clear everything? Or reset to a default CS_USER

clear everything

backend/server/routers/user.py
```diff
@@ -147,3 +147,56 @@ def update_degree_length(numYears: int, token: str = DUMMY_TOKEN):
...
+@router.put("/reset")
+def reset(token: str = DUMMY_TOKEN):
+    user = get_user(token)
+    planner: PlannerLocalStorage = {
+        'mostRecentPastTerm': {
+            'Y': 0,
+            'T': 0
+        },
+        'unplanned': [],
+        'isSummerEnabled': True, # shoudl default be false? who knows
```
leonardo-fan: default True sounds good to me

leonardo-fan (approved):
lgtm, just remove the comment and should be good

PR #984, opened 2023-03-01
feat: Replace unschedule reducer with API.

Bahnschrift:
Can't completely remove the reducer itself until I've also removed usages of `unscheduleAll`.
Waiting on other backend routes to be implemented before adding backend tests.


Bahnschrift:
Actually, it seems like tests for this and unscheduleAll already exist in backend/server/tests/test_moving_course.py which both pass. Maybe it would be fine to merge this and #985 as is?

backend/server/routers/planner.py
```diff
@@ -289,19 +289,14 @@ def unschedule(data: CourseCode, token: str = DUMMY_TOKEN):
...
     """
     user = get_user(token)
-    planner = user['planner']
 
-    removed: Set[str] = set()
-    # Remove every instance of the course from each year
-    for year in planner['years']:
+    # Remove every instance of the course from each year and add to unplanned
+    for year in user['planner']['years']:
         for course_list in year.values():
```
imagine-hussain, line 296: nitpick: probably just inline `year`; its not used again and is probably more readable than the nesting

```suggestion
for course_list in user['planner']['years'].values():
```

frontend/src/pages/TermPlanner/TermPlanner.tsx
```diff
@@ -167,12 +166,14 @@ const TermPlanner = () => {
...
-      dispatch(
-        unschedule({
-          destIndex,
-          code: draggableId
-        })
-      );
+      try {
+        await axios.post('/planner/unscheduleCourse', JSON.stringify({ courseCode: draggableId }), {
+          params: { token }
+        });
```
imagine-hussain, line 172: @leonardo-fan can you verify this is the correct way to make this request since i've also seen it passed straight in without the `params: ` part

imagine-hussain (approved):
same thoughts as #985 

lgtm

PR #985, opened 2023-03-04
feat: Replace unscheduleAll reducer with API

Bahnschrift:
Waiting on other backend routes to be implemented before adding backend tests.

### DIFF
--- a/backend/server/routers/planner.py
+++ b/backend/server/routers/planner.py
@@ -314,17 +314,13 @@ def unschedule_all(token: str = DUMMY_TOKEN):
         token (str, optional): The user's authentication token. Defaults to DUMMY_TOKEN.
     """
     user = get_user(token)
-    removed: Set[str] = set()
 
-    # Remove every course from each year
+    # Remove every course from each term and add it to unplanned
     for year in user['planner']['years']:
-        for term, course_list in year.items():
-            removed = removed | set(course_list)
-            year[term] = []
+        for course_list in year.values():
+            user['planner']['unplanned'].extend(course_list)
+            course_list.clear()
 
-    # Add every removed course to unplanned column
-    for course in removed:
-        user['planner']['unplanned'].append(course)
     set_user(token, user, True)
 
 
--- a/frontend/src/pages/TermPlanner/OptionsHeader/OptionsHeader.tsx
+++ b/frontend/src/pages/TermPlanner/OptionsHeader/OptionsHeader.tsx
@@ -13,9 +13,10 @@ import {
 } from '@ant-design/icons';
 import Tippy from '@tippyjs/react';
 import { Popconfirm, Switch, Tooltip } from 'antd';
+import axios from 'axios';
 import migrateLocalStorageData from 'utils/migrateLocalStorageData';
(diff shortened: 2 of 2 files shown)


frontend/src/pages/TermPlanner/OptionsHeader/OptionsHeader.tsx
```diff
@@ -106,7 +116,7 @@ const OptionsHeader = ({ plannerRef }: Props) => {
             <Popconfirm
               placement="bottomRight"
               title="Are you sure you want to unplan all your courses?"
-              onConfirm={() => dispatch(unscheduleAll())}
+              onConfirm={handleUnscheduleAll}
```
Bahnschrift, line 119: I'm fairly sure this was the only use of the `unscheduleAll` reducer, so I think it should be fine to remove it from `plannerSlice.ts`?

Bahnschrift:
Actually, it seems like tests for this and unscheduleCourse already exist in `backend/server/tests/test_moving_course.py` which both pass. Maybe it would be fine to merge this and #984 as is?

frontend/src/pages/TermPlanner/OptionsHeader/OptionsHeader.tsx
imagine-hussain: lgtm; this can easily be a one-off call

imagine-hussain: @leonardo-fan does having it as an anonymous function vs passing the function in have any consequence with the Event object being passed in incidentally later on?

imagine-hussain (approved):
not much to change here, this looks good to me;
feel free to merge.

for testing, can you just make a list of routes that are blocking you so we know what to prioritise? cheers

frontend/src/pages/TermPlanner/OptionsHeader/OptionsHeader.tsx
leonardo-fan: nope this seems all good, handleUnscheduleAll is an anon function anyway.

leonardo-fan (approved):
lgtm - re: testing do that in a separate pr when you get up to it

leonardo-fan (changes requested):
lgtm - except one thing with the backend route

backend/server/routers/planner.py
```diff
@@ -314,17 +314,13 @@ def unschedule_all(token: str = DUMMY_TOKEN):
         token (str, optional): The user's authentication token. Defaults to DUMMY_TOKEN.
     """
     user = get_user(token)
```
leonardo-fan, line 316: need a way to get the token from the request - i suggest sending this token via an authorization bearer token header

leonardo-fan: you've done it via params in the fe request - which is valid, albeit there's no way in the backend to get that token through the route. however this is not really done in practice

leonardo-fan: will do this as a later change

leonardo-fan (approved):

PR #988, opened 2023-03-07
disable adv cs

imagine-hussain:
- adv cs electives are broken - need to fix
- going to disable until they're ready


leonardo-fan (approved):
Aight bet

PR #991, opened 2023-03-08
feat(BE/FE): replace toggleSummer reducer with API [CF-412]

lhvy:
Replaces the toggleSummer plannerSlice reducer with the matching backend API (`/user/toggleSummerTerm`)

### DIFF
--- a/backend/server/routers/user.py
+++ b/backend/server/routers/user.py
@@ -118,6 +118,10 @@ def default_cs_user() -> Storage:
 def toggle_summer_term(token: str = DUMMY_TOKEN):
     user = get_user(token)
     user['planner']['isSummerEnabled'] = not user['planner']['isSummerEnabled']
+    if not user['planner']['isSummerEnabled']:
+        for year in user['planner']['years']:
+            user['planner']['unplanned'].extend(year['T0'])
+            year['T0'] = []
     set_user(token, user, True)
 
 
--- a/frontend/src/pages/TermPlanner/ImportPlannerMenu/ImportPlannerMenu.tsx
+++ b/frontend/src/pages/TermPlanner/ImportPlannerMenu/ImportPlannerMenu.tsx
@@ -7,7 +7,7 @@ import { Course } from 'types/api';
 import { JSONPlanner, Term, UnPlannedToTerm } from 'types/planner';
 import openNotification from 'utils/openNotification';
 import type { RootState } from 'config/store';
-import { moveCourse, toggleSummer } from 'reducers/plannerSlice';
+import { moveCourse } from 'reducers/plannerSlice';
 import CS from '../common/styles';
 import S from './styles';
 
@@ -108,7 +108,16 @@ const ImportPlannerMenu = () => {
             return;
           }
           if (planner.isSummerEnabled !== fileInJson.isSummerEnabled) {
-            dispatch(toggleSummer());
+            try {
+              axios.post('/user/toggleSummerTerm', {}, { params: { token } });
+            } catch {
+              openNotification({
+                type: 'error',
+                message: 'Error setting summer term',
+                description: 'An error occurred when toggling the summer term.'
+              });
+              return;
+            }
           }
           fileInJson.years.forEach((year, yearIndex) => {
             Object.entries(year).forEach(([term, termCourses]) => {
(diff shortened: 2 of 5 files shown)


backend/server/routers/user.py
```diff
@@ -99,6 +99,10 @@ def default_cs_user() -> Storage:
 def toggle_summer_term(token: str = DUMMY_TOKEN):
     user = get_user(token)
     user['planner']['isSummerEnabled'] = not user['planner']['isSummerEnabled']
+    if not user['planner']['isSummerEnabled']:
```
lhvy, line 121: There doesn't appear to be sample data for this in `example_local_storage_data.json` as no data in that file has a populated `T0`. What would you like me to do?

leonardo-fan: This looks fine to me - if you check out local storage on the FE when you have courses in summer terms, you'll be able to see they are populated as you assumed

frontend/src/pages/TermPlanner/ImportPlannerMenu/ImportPlannerMenu.tsx
```diff
@@ -79,7 +79,16 @@ const ImportPlannerMenu = () => {
...
+            try {
+              axios.put('/user/toggleSummerTerm', {}, { params: { token } });
+            } catch {
+              openNotification({
+                type: 'error',
+                message: 'Error setting summer term',
+                description: 'An error occurred when toggling the summer term.'
+              });
+              return;
+            }
```
leonardo-fan: Make function async and await the request. 

Also since this notification is for importing a file to change settings I recommend something more specific to this use case for the description e.g. `An error occurred when trying to import summer term visibility` or something along those lines.

frontend/src/pages/TermPlanner/SettingsMenu/SettingsMenu.tsx
```diff
@@ -28,7 +30,16 @@ const SettingsMenu = () => {
...
+    try {
+      axios.put('/user/toggleSummerTerm', {}, { params: { token } });
+    } catch {
+      openNotification({
+        type: 'error',
+        message: 'Error setting summer term',
+        description: 'An error occurred when toggling the summer term.'
+      });
+      return;
+    }
```
leonardo-fan: Make async function/await the request - error message looks fine to me

leonardo-fan (changes requested):
Nice job Lucas! Just a few comments

backend/server/routers/user.py
```diff
@@ -99,6 +99,10 @@ def default_cs_user() -> Storage:
 def toggle_summer_term(token: str = DUMMY_TOKEN):
```
leonardo-fan, line 118: another thing, I suggest using auth bearer token header to get the token from the request - also reflect this in the FE request too -> much more standard practice that params

frontend/src/pages/TermPlanner/SettingsMenu/SettingsMenu.tsx
lhvy: Fixed

frontend/src/pages/TermPlanner/ImportPlannerMenu/ImportPlannerMenu.tsx
lhvy: Fixed

backend/server/routers/user.py
lhvy: Fixed

frontend/src/pages/TermPlanner/SettingsMenu/SettingsMenu.tsx
```diff
@@ -37,8 +37,17 @@ const SettingsMenu = () => {
     }
   }
 
-  function handleSummerToggle() {
-    dispatch(toggleSummer());
+  async function handleSummerToggle() {
+    try {
+      await axios.put('/user/toggleSummerTerm', {}, { params: { token } });
```
imagine-hussain: make this a `post` request; `put` requests should be idempotent

imagine-hussain:
looking good; like discussed in meeting, can keep token in params for now, so should be gucci

frontend/src/pages/TermPlanner/SettingsMenu/SettingsMenu.tsx
lhvy: Oops, fixed now.

PR #998, opened 2023-03-14
feat: replace remove courses with api

VinayakFia:


imagine-hussain (changes requested):
same feedback as the other PR, looks good for a merge before migrating to react-query;

only change needed is removing that Scripts file or, adding it to `.gitignore`;
re-request when that's no longer in the git diff. otherwise lgtm!

PR #1000, opened 2023-03-15
feat: replace removeAllCourses reducer with backend api call

VinayakFia:


imagine-hussain (changes requested):
same as other branches; remove `.gitignore` and then re-req to merge

PR #1003, opened 2023-03-19
feat: CF-400: dark mode improvements

Dasyure:
@leonardo-fan this was what I was talking about in my standup from Wednesday. This draft PR contains a few commits but the ones in question involve the settingsMenu's select dropdown and datePicker, inside SettingsMenu.tsx. I added pictures on the jira ticket of the end product if you're interested. 
https://csesoc.atlassian.net/jira/software/projects/CF/boards/12/?selectedIssue=CF-400

What I tried to do:
- Initially I made a new style.ts file where I used styled components to import and edit an antd Select element, then export it back to SettingsMenu.tsx. However it didn't work as I get the error "No overload matches this call."
- I tried to use the given dropdownStyle property, as per the antd documentation. However it only partially worked.
- I also tried to override the styling within the styled component by adding styles to the classes themselves but that only half worked. 
What I ended up doing:
- I edited GlobalStyles in src/config/theme.ts and used the !important property to override the antd styles. 

Concern:
- is there a cleaner way of overriding the antd styles? I spent a while trying to figure it out but failed to. I think the problem is that some components re-render (like select), hence we have to edit GlobalStyles. 
- The main component I'm concerned about is the select dropdown. Because it means we'd have to edit GlobalStyles for every select dropdown we use, which is quite messy. I'll keep trying to find a fix though.


Dasyure:
https://ant.design/docs/react/customize-theme
One potential solution would be to upgrade to v5 for antd and play around with this. The downside of doing this is:
1. If it doesn't work perfectly then it'll be a big waste of time, since antd hasn't been very good at being customisable with past versions
2. Having to rewrite a lot of code that relies on antd v4

frontend/src/components/CourseDescriptionPanel/CourseInfoDrawers.tsx
```diff
@@ -33,83 +37,85 @@ const CourseInfoDrawers = ({
   const inPlanner = planner.courses[course.code];
 
   return (
-    <div>
+    <div className="course-info-drawers">
```
leonardo-fan, line 40: doesn't seem like this is used/defined - should remove

frontend/src/components/SidebarDrawer/SidebarDrawer.tsx
```diff
@@ -9,7 +9,7 @@ const SidebarDrawer = ({ children }: Props) => {
   const [open, setOpen] = useState<boolean | undefined>(undefined);
 
   return (
-    <S.Wrapper open={open}>
+    <S.Wrapper open={open} className="sidebar-drawer">
```
leonardo-fan, line 12: className not defined

frontend/src/pages/TermPlanner/SettingsMenu/SettingsMenu.tsx
```diff
@@ -54,7 +55,7 @@ const SettingsMenu = () => {
           unCheckedChildren={<CloseOutlined />}
         />
       </CS.PopupEntry>
-      <CS.PopupEntry>
+      <CS.PopupEntry className="settings-start-year-popup">
```
leonardo-fan, line 58: no className definition

leonardo-fan (changes requested):
Really great job with all the changes! Looks good, just some small comments regarding unused class names. 

Also had a bit of a test and there are some remaining dark mode things that I saw (refer to videos/images below). Since there are a few dependencies on this pr right now - definitely don't feel pressured to get these improvements done now. Let me know if you still want to do them, otherwise I'll put the further fixes in the backlog

Screenshot 2023-04-04 at 7 03 19 pm
Screenshot 2023-04-04 at 7 03 26 pm

https://user-images.githubusercontent.com/90736577/229743529-9c47dd10-0f77-43f8-8185-594919945e72.mov


https://user-images.githubusercontent.com/90736577/229743542-cc73aa24-54f4-4b48-a6c3-e9e7f640cd95.mov


https://user-images.githubusercontent.com/90736577/229743549-3ae3a9ef-0bbc-4fcd-95e7-71a6afd51e06.mov


https://user-images.githubusercontent.com/90736577/229743561-88e82f18-e93d-44c6-aa85-8d7b24bcd864.mov


https://user-images.githubusercontent.com/90736577/229743571-713a3329-0198-42d0-80aa-c1a3df0f4181.mov

Dasyure:
@leonardo-fan Thanks for pointing those out, I'm happy to finish off the dark mode features I forgot to cover. 

The only issue is there are a 3 other PRs which are dependent on this. Should I just wait for all 4 PRs to be merged in and then fix all the issues afterward in one go?

leonardo-fan:
> Really great job with all the changes! Looks good, just some small comments regarding unused class names.
> 
> Also had a bit of a test and there are some remaining dark mode things that I saw (refer to videos/images below). Since there are a few dependencies on this pr right now - definitely don't feel pressured to get these improvements done now. Let me know if you still want to do them, otherwise I'll put the further fixes in the backlog

refer to https://csesoc.atlassian.net/jira/software/projects/CF/boards/12/backlog?selectedIssue=CF-426 for further dark mode fixes after this PR is merged in

leonardo-fan (approved):

PR #1032, opened 2023-03-23
Style/update-typing

Bahnschrift:
Replaces usages of type hints from the typing library (List, Tuple, etc.) with builtins (see [PEP 585](https://peps.python.org/pep-0585/).

Also replaced usages of typing.Optional[T] with T | None, which is equivalent.

### DIFF
--- a/backend/server/routers/model.py
+++ b/backend/server/routers/model.py
@@ -2,7 +2,7 @@
 # pylint: disable=missing-class-docstring
 import json
 import pickle
-from typing import Literal, Optional, TypedDict
+from typing import Literal, TypedDict
 
 from algorithms.objects.conditions import CompositeCondition
 from algorithms.objects.user import User
@@ -28,7 +28,7 @@ class CourseDetails(BaseModel):
     level: int
     description: str
     study_level: str
-    school: Optional[str]
+    school: str | None
     campus: str
     equivalents: dict[str, str]
     raw_requirements: str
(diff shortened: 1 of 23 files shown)


backend/server/routers/model.py
```diff
@@ -28,7 +28,7 @@ class CourseDetails(BaseModel):
     level: int
     description: str
     study_level: str
-    school: Optional[str]
+    school: str | None
```
Bahnschrift, line 31: Does this actually need to be optional? Setting it to just `str` doesn't cause any mypy errors.

imagine-hussain: I don't remember why, but it might have been a thing with how data comes in / out from the frontend and also our database. But also I could be wrong

imagine-hussain (commented):
pep is wrong; I like the capital letters on the ones from the typing library;

I'll have to think about this one

imagine-hussain:
The `lists` vs `List`, I have mixed feelings on but, `Optional[T]` > `T | None`

Bahnschrift:
> pep is wrong; I like the capital letters on the ones from the typing library;
> 
> I'll have to think about this one

Haha fair enough. Doesn't really matter if this gets merged or not, just thought I'd make it it quickly and see.

> I have mixed feelings on but, Optional[T] > T | None

Made that a separate commit since I thought it'd be a bit more controversial. I'm not sold on it either.

imagine-hussain:
I want there to be another refactor of our types, mainly toward using more explicit types and `typeddict`s for our data instead of nested `dict` / `list`.

May revisit the idea of using native `list` annotation over `List` but until then, I'm going to close this.

Great job at taking initiative though!

PR #1033, opened 2023-03-29
feat: CF-393: Replaced setProgram dispatches with axios requests

ui-Auxilary:
Hello~
I tried combining 3 tickets into one as they featured similar changes. Not sure if that was an illegal move :((.
This was my first time modifying the frontend, so would appreciate some feedback before I go on to write a few tests (!)

I kind of just referenced the other PRs that were being fired out ^^.


Dasyure:
Looks nice, I was just reading to get familiar with how to do this myself haha
Although just one quick note, I think the indentation is a bit off? There's some uses of 4 space or 8 space indentation.

imagine-hussain (changes requested):
This is looking good!
As sandeep mentioned, there's some issues with indentation that should get linted before merging.
Once those are fixed, this should be good to merge and from here, should be a fairly straightforward port to move this to `react-query`!

imagine-hussain (approved):

PR #1034, opened 2023-03-29
feat: CF-423: additional styling on course node hover

Dasyure:
This ticket is dependent on CF-400 being merged in first. Should I merge into CF-400 instead? 

View the ticket for more details:
https://csesoc.atlassian.net/browse/CF-423

Also if this is good enough to be added in, I could also change the HelpMenu for the graphical selector to include a picture for it.

Note:
- The method for changing style on hover is slower, however it allows for greater customisation (e.g. different styles for dark mode, changing the label's color)


frontend/src/pages/GraphicalSelector/CourseGraph/CourseGraph.tsx
```diff
@@ -124,47 +204,63 @@ const CourseGraph = ({ onNodeClick, handleToggleFullscreen, fullscreen, focused
...
+
+    // Update styling for: each node, hovering state and edges
+    const repaintCanvas = async () => {
+      const nodes = graphRef.current?.getNodes();
+      nodes?.map((n) =>
+        graphRef.current?.updateItem(n, mapNodeStyle(n.getID(), plannedCourses, theme))
+      );
+
+      graphRef.current?.off('node:mouseenter');
+      graphRef.current?.off('node:mouseleave');
```
leonardo-fan, line 228: do you need these?

leonardo-fan (approved):
nice! lgtm :D

frontend/src/pages/GraphicalSelector/CourseGraph/CourseGraph.tsx
Dasyure: Yep, without these multiple event listeners get added when you try and hover over a node. The graph then gets laggy after you hover about 5-10 nodes quickly. 
Those two lines remove all listeners so you can start fresh and update the listener.

PR #1037, opened 2023-04-02
fix: CF-421: adding or removing courses now reflect in the graphical selector

Dasyure:
More details here:
https://csesoc.atlassian.net/jira/software/projects/CF/issues/CF-421

Merge in CF-356 first, before this one.

The order of merging is now CF-400, CF-423, CF-356 and now CF-421. 

EDIT: there's a tiny bug. Adding a course is reflected fine, but the newly unlocked courses aren't reflected. I might have to create another ticket to fix this, since it involves converting the reducer to axios requests. 
Edit 2: fixed


leonardo-fan (approved):
looks bombastic to me

PR #1039, opened 2023-04-08
fix: CF-425: cleaning graph loading screen

Dasyure:
More details here:
https://csesoc.atlassian.net/jira/software/projects/CF/issues/CF-425

The order of merging is now CF-400, CF-423, CF-356, CF-421 then CF-425.


leonardo-fan (approved):
noice, lgtm! :) documentation goes crazy

PR #1056, opened 2023-05-25
Use react query for term planner page

Bahnschrift:
**Very draft not working do not merge.**

This will probably also require rewriting a fair few backend routes and also a lot of the storage types.


lhvy:
Replaced by #1061

PR #1061, opened 2023-06-24
Update Term Planner to use React Query

Bahnschrift:


lhvy:
Almost there, the things that need fixing are

- Warnings
- Editing marks (still using redux)
- Most of the helpful buttons and settings (already make API calls, needs react-query)
- Testing please

Peedee2002:
This change has been merged

PR #1067, opened 2023-07-31
feat(graphical-selector): move graphical selector to use reactQuery and backend user data

ollibowers:
I wanna say this is done. 🤯 

Moved the graphical selector over to useQuery. 

To do so:
- Created a new query key 'graph' for the graph structure. This should never really get invalidated as this would require the handbook to change.
- deprecated old /courses/getAllUnlocked route.
- created new /user/unlockedCourses/{token} route.
- created functions to convert the storage user into a algorithm user for this.
- cleaned up the CourseGraph.tsx file a little bit more.
- decided to keep old /programs/graph/{code}/{specs} route since i wanted to have the degree anyway for effect dependencies, and graph structures arent user unique. It keeps it open for future ideas (that have been brought up in the past) of being able to have multiple 'tabs' of graphs on the graphicalSelector page, for different degrees and what not. 
- also fixed some typing.

i love you 💛 

p.s.
if you want to test this out, you will need to change `/addToUnplanned` to put courses into the user['courses'] dictionary as well.

### DIFF
--- a/backend/algorithms/objects/user.py
+++ b/backend/algorithms/objects/user.py
@@ -79,7 +79,10 @@ class User:
 
     def has_taken_course(self, course: str):
         """ Determines if the user has taken this course """
-        return any(c in self.courses and (self.courses[c][1] or 50) >= 50 for c in chain([course], (CACHED_EQUIVALENTS.get(course) or [])))
+        return any(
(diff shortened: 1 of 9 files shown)


ollibowers:
picture 
image

Peedee2002:
This has been merged

PR #1073, opened 2023-09-04
  Christian/feat/add login page

ui-Auxilary:
Fixed up some bugs in the Landing page ie. gap at the footer, and animation not repeating for Circle logo

- Added login page


imagine-hussain (commented):
This is looking great!

Just some minor points:

- at medium laptop width (around the 1200-1024 px mark), the horizontal margins on the container are probably a little big and make the login look smaller than we want.


image

- Since removing google, we should now be replacing that with a "continue as guest" option

imagine-hussain:
Forgot to mention, this should also have a dark-mode implementation, though that doesn't necessarily need to be done this PR since this isn't directly going to production, that would be the next thing to look at.

frontend/src/pages/Auth/Container/Container.tsx
```diff
@@ -0,0 +1,29 @@
...
+      <S.Wrapper>
+        <S.Left>
+          <S.SplashArt src={SplashArt} />
+        </S.Left>
+        <S.Right>
+          <S.Login>
+            <Link to="/">
+              <S.Back src={BackButton} />
+            </Link>
+            <h2>Login to circles</h2>
```
leonardo-fan: "Login to Circles"

leonardo-fan (changes requested):
Looks awesome! same comment as Hussain on the L & R margins. One grammar thing in comments. Also thoughts on keeping the login container centered vertically as well?

frontend/src/pages/Auth/Auth.tsx
```diff
@@ -34,97 +40,65 @@ const Auth = () => {
...
-      <div onClick={logout}>
-        <h1> LOGOUT </h1>
-      </div>
-    );
-  }
   interface Props {
-    obj: any
+    obj: any;
   }
   const JsonObject = ({ obj }: Props) => {
```
ollibowers: If no progress was made towards google login, should we just remove this?

frontend/src/pages/Auth/Auth.tsx
```diff
@@ -34,97 +40,65 @@ const Auth = () => {
   // };
 
   const handleCallbackResponse = (response: any) => {
```
ollibowers: Google?

frontend/src/pages/Auth/Auth.tsx
```diff
@@ -34,97 +40,65 @@ const Auth = () => {
...
-    console.log("callBackResponse with jwt: ", jwt);
+    console.log('callBackResponse with jwt: ', jwt);
     setUserObject(jwt_decode(jwt));
 
-    
     return 0;
-  }
+  };
 
   useEffect(() => {
```
ollibowers: Agan google

frontend/src/pages/Auth/Auth.tsx
```diff
@@ -9,11 +9,17 @@ import type { RootState } from 'config/store';
...
 import PageTemplate from 'components/PageTemplate';
+import Container from './Container';
+import { inDev } from 'config/constants';
+import S from './styles';
+import axios from 'axios';
+import { generateUserToken, userLogin, userLogout } from 'utils/api/userApi';
 
 const Auth = () => {
   const [userObject, setUserObject] = useState({});
+  const [dummyToken, setDummyToken] = useState('');
```
ollibowers: I don't think this state is used anymore

package-lock.json
ollibowers, line 1: i dont know if this should be here (it is not the real package-lock.json)

ollibowers (changes requested):
oops i did the changes as comments

frontend/src/pages/Auth/Auth.tsx
imagine-hussain: Googlie die

imagine-hussain (approved):
LGTM!!!

imagine-hussain:
I wonder if I have to manually merge this because of failing pipelines of if this should work

Peedee2002:
This has been merged

PR #1090, opened 2023-10-21
add (basically) working backend route

martanman:


backend/requirements.txt
```diff
@@ -14,3 +14,5 @@ requests==2.28.1
 types-paramiko==2.12.0.1
 types-requests==2.28.11.5
 uvicorn==0.20.0
+PyPDF2==3.0.1
+python-multipart==0.0.6
```
imagine-hussain, line 18: is this import needed / used?

imagine-hussain (approved):
small qn on the dependency;
otherwise, have fixed the mypy stuff, looks good to merge

imagine-hussain:
Changed target branch to`331-staging` so that we dont need to merge 331 back in.

backend/requirements.txt
martanman: do u mean are the requirements necessary? yes

imagine-hussain: okay in that case, looks good to merge

PR #1102, opened 2024-02-05
chore: update CODEOWNERS for 2024 devsoc

ollibowers:
🥺  would like platform approval before merge

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,4 @@
+* @martanman @ollibowers
+
 .github/ @devsoc-unsw/platform
 renovate.json @devsoc-unsw/platform
-
-* @martanman @ollibowers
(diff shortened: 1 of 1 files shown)


lhjt:
I think you should have the github circles team owner role rather than hardcoding you and martin to be the codeowners so we don't have to change this every year 👍

lhjt (changes requested):
Change this to reference your team handle instead of referencing you both directly

ollibowers:
> Change this to reference your team handle instead of referencing you both directly

Would you want the platform codeowners to just be platform leads as well? @lhjt

lhjt:
> > Change this to reference your team handle instead of referencing you both directly
> 
> Would you want the platform codeowners to just be platform leads as well? @lhjt

Oh I think just platform is actually fine since the team will need to be making changes as a whole 👍

ollibowers:
> > > Change this to reference your team handle instead of referencing you both directly
> > 
> > 
> > Would you want the platform codeowners to just be platform leads as well? @lhjt
> 
> Oh I think just platform is actually fine since the team will need to be making changes as a whole 👍

awesome, am good to go then

lhjt (approved):

PR #1108, opened 2024-03-25
Bring to dev

Peedee2002:
made changes to bring this branch up to date with dev. I also deleted importing and exporting the planner files, because we dont need them anymore.

Bugs fixed:
- [x] rc picker has wierd highlighting 
- [x] term planner doesnt load if no courses are picked 
- [x] remove background thing 
- [x] fix term box not growing properly
- [x] liquid didnt format the same as before (fixed as best as olli could, might just need to revert antv charts)
- [x] react query loss of focus 
- [x] bug where the toggle for colour is wierd 
- [x] grpahical selector styles not resetting and throwing many errors
- [x] Light and dark mode toggle images not being the right size (at least on my screen)
- [x] loading on first course selection tab broke
- [x] fixed noticable lag in CourseMenu
- [x] fixed progression checker course buttons styling


ollibowers (changes requested):
Looks good to me, just would like the export import json/png stuff back (and just commented out instead)

ollibowers (approved):
oki happy now

PR #1114, opened 2024-04-04
add courses run in prev years to /dump route as well

martanman:
due to popular (unilectives) demand. fixes only returning courses that are running in the current year


ollibowers (approved):
lgtm

PR #1054, opened 2023-05-19
feat(frontend) fuzzy-find-degree-wizard

imagine-hussain:
- in setup, degree is now fuzzily searched
- no longer need to type *exact* degree
- fuzz is weighted by what's effectively a normalised version of levenstein distance
- in cases of equal weighting (usualyl when both have a weight of 1), uses string len as tie breaker
- only gives user the 8 most relevant results

image

PR #1133, opened 2024-04-18
feat/fe/fuzzy to degree search 331 edition

imagine-hussain:
rehash of #1054 but to target 331-staging instead of dev


imagine-hussain:
tests are broken because 331 is broken; this should be fine

martanman (approved):
lgtm :)

PR #1134, opened 2024-04-18
Fix term planner getting course info across years

lhvy:


frontend/src/types/planner.ts
```diff
@@ -1,3 +1,6 @@
+type LiveYearType = 2024;
+export const LiveYear: LiveYearType = 2024;
```
ollibowers: Not a fan of this pascal case on the constant, can please all caps it or atleast camelCase 🥺 

Also, is there a purpose for this `LiveYearType`?
Could we just do `export const liveYear = 2024 as const;` if you want to make sure the type is a literal of `2024`?

ollibowers (changes requested):

frontend/src/types/planner.ts
lhvy: Fixed

ollibowers (approved):
looks beautiful

PR #1135, opened 2024-04-18
Feat/fe/1131/adding link course outline on course selector

### LINKED ISSUE #1131: Link to course outline on course selector
There is already a handbook link, add another.

jacksonwanggg:
adding link course outline on course selector and tracks the first term, hard-coded year to 2024,

closes #1131


frontend/src/components/CourseDescriptionPanel/CourseAttributes/CourseAttributes.tsx
```diff
@@ -83,6 +83,18 @@ const CourseAttributes = ({ course, courseCapacity }: CourseAttributesProps) =>
             </S.Link>
           ) : null
         },
+        {
+          title: 'UNSW Course Outline',
+          content: studyLevel ? (
+            <S.Link
+              href={`https://www.unsw.edu.au/course-outlines/course-outline#year=2024&term=Term%201&deliveryMode=Multimodal&deliveryFormat=Standard&teachingPeriod=${terms[0]}&deliveryLocation=Kensington&courseCode=${code}&activityGroupId=1`}
```
lhvy: > &term=Term%201

Seems like the term is still hardcoded here. Not sure if teaching period can be hardcoded as it is currently or if it needs something special?

jacksonwanggg: Thank you lucas, 
I'm working on it that t1 it would be 201 t2 it would be 202 and t3 it would be 203, then summer course it would be term=summer 

also not sure if the teaching period should be like that, because if we open up a course like comp4128 which is only offered during t3, and click the link it would show not present because the course outline would be released once the course starts, so the teaching period wouldnt be right.

lhvy: I wonder if we need to start scraping course outlines properly, and then only show actual outline URLs we have...

frontend/src/components/CourseDescriptionPanel/CourseAttributes/CourseAttributes.tsx
```diff
@@ -6,13 +6,20 @@ import { useTheme } from 'styled-components';
...
-import { TERM } from 'config/constants';
+import { CURR_YEAR, TERM } from 'config/constants';
 import S from './styles';
 
 const { Title, Text } = Typography;
 
+const termMapping: Record<number, string> = {
+  1: 'Term%201',
+  2: 'Term%202',
+  3: 'Term%203'
```
ollibowers, line 21: looks good, but we also need to account for `0: 'Term%20Summer'`

frontend/src/components/CourseDescriptionPanel/CourseAttributes/CourseAttributes.tsx
```diff
@@ -31,6 +38,16 @@ const CourseAttributes = ({ course, courseCapacity }: CourseAttributesProps) =>
 
   const { study_level: studyLevel, terms, campus, code, school, UOC } = course;
 
+  const currentTerm = getMostRecentPastTerm(CURR_YEAR).T;
+
+  const recentTermNumber = parseInt(terms[0].slice(1), 10);
```
ollibowers: Terms might be an empty array if there are no offerings of that course, which will cause a out of range error. 

Also are you meant to only be grabbing the 0th element here? For COMP2521, terms is [t0, t1, t2, t3], and for COMP1511, terms is [t1, t2, t3]. Is it just meant to be the earliest available term of this year?

ollibowers (changes requested):

frontend/src/components/CourseDescriptionPanel/CourseAttributes/CourseAttributes.tsx
jacksonwanggg: uhm i did hte check for the out of range error

but right now i'm comparing 0th element which is the most current term and comparing with the term i fetched using the function getMostRecentPastTerm and comparing it if the term i fetched it larger I would use their one but i'm not sure why its not jumping onto t2, but it is fetching the first term so it can compare

ollibowers:
#1131

frontend/src/components/CourseDescriptionPanel/CourseAttributes/CourseAttributes.tsx
```diff
@@ -6,13 +6,22 @@ import { useTheme } from 'styled-components';
 import { Course } from 'types/api';
 import { EnrolmentCapacityData } from 'types/courseCapacity';
 import { getCourseRating } from 'utils/api/unilectivesApi';
+import getMostRecentPastTerm from 'utils/getMostRecentPastTerm';
 import ProgressBar from 'components/ProgressBar';
 import TermTag from 'components/TermTag';
-import { TERM } from 'config/constants';
+import { CURR_YEAR, TERM } from 'config/constants';
 import S from './styles';
+import { c } from 'vite/dist/node/types.d-aGj9QkWt';
```
ollibowers:
```suggestion

```
I think this might have been added by accident.

frontend/src/components/CourseDescriptionPanel/CourseAttributes/CourseAttributes.tsx
```diff
@@ -31,6 +40,18 @@ const CourseAttributes = ({ course, courseCapacity }: CourseAttributesProps) =>
...
 
+  const currentTerm = getMostRecentPastTerm(CURR_YEAR).T;
+
+  const recentTermNumber = terms?.length ? parseInt(terms[0].slice(1), 10) : currentTerm;
+
+  const yearCourseAvaliable = recentTermNumber > currentTerm ? CURR_YEAR - 1 : CURR_YEAR;
+
+  const updatedTerm = recentTermNumber > currentTerm ? recentTermNumber : currentTerm;
+
+  const teachingPeriod = updatedTerm > 0 ? 'T' + updatedTerm : 'U0';
```
ollibowers:
```suggestion
  const teachingPeriod = updatedTerm > 0 ? `T${updatedTerm}` : 'U0';
```
Changing this to be a template string, as eslint complains otherwise.

ollibowers (changes requested):
looks beautiful otherwise

ollibowers (approved):
lgtm

PR #1139, opened 2024-05-09
Feat/fe/quick course add button in search

imagine-hussain:
Adds a `QuickAddCartButton` to the search bar. This allows for users to add courses to their planner quickly, without needing to load a new course tab, press add to planner and then go back to search. [Image at bottom]

The two uses of the search are the `CourseBar` in `/course-selector` and, `/graphical-selector`.
Currently both have to inject the mutations and `CoursesResponse` state as a dependency.
Consider just calling queries and mutations directly when the react-query mutations are unified.

---

We also perform a few misc. cleanups:
- Prefer `CourseBarSearchProps` over `Props` as a name
- Use `Array<T>` over `T[]`, especially for anonymous record types. This change is for the singular instance relevant to this PR and is not a treesmash.

---

image

### DIFF
--- a/frontend/src/components/CourseSearchBar/CourseSearchBar.tsx
+++ b/frontend/src/components/CourseSearchBar/CourseSearchBar.tsx
@@ -10,7 +10,7 @@ type Props = {
 
 const CourseSearchBar = ({ onSelectCallback, style }: Props) => {
   const [value, setValue] = useState<string | null>(null);
-  const [courses, setCourses] = useState<{ label: string; value: string }[]>([]);
+  const [courses, setCourses] = useState<Array<{ label: string; value: string }>>([]);
   const [isLoading, setIsLoading] = useState(false);
   const [debouncedSearchTerm] = useDebounce(value, 200);
 
--- a/frontend/src/components/CourseSearchBar/CourseSearchBar.tsx
+++ b/frontend/src/components/CourseSearchBar/CourseSearchBar.tsx
@@ -21,13 +21,13 @@ const SearchResultLabel = ({
   runMutate,
   isPlanned
 }: SearchResultLabelProps) => {
+  const codeAndTitleText = `${courseCode}: ${courseTitle}`;
+
   return (
-    <div>
+    <div title={codeAndTitleText}>
       <QuickAddCartButton courseCode={courseCode} runMutate={runMutate} planned={isPlanned} />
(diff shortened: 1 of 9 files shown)


imagine-hussain:
I think there should be more space between the quick add button and the text.
Thoughts? @ollibowers @martanman

martanman:
yeah, would probably be good if left spacing to the edge was similar to right spacing to text

imagine-hussain:
> yeah, would probably be good if left spacing to the edge was similar to right spacing to text

I've added a `1` character spacing here. Thoughts?

image

lhvy:
Looks good to me after the change

frontend/src/components/CourseSearchBar/CourseSearchBar.tsx
```diff
@@ -15,6 +15,7 @@ type SearchResultLabelProps = {
   runMutate?: (courseId: string) => void;
 };
 
+/* eslint-disable */
```
lhvy: 😭

imagine-hussain: If were up to me, this would be in root :)

(i'll remove)

ollibowers: >:(

imagine-hussain:
image

frontend/src/components/CourseSearchBar/CourseSearchBar.tsx
```diff
@@ -62,7 +115,7 @@ const CourseSearchBar = ({ onSelectCallback, style }: Props) => {
       onSearch={handleSearch}
       onSelect={handleSelect}
       notFoundContent={isLoading && value && <Spin size="small" />}
-      style={{ width: '30rem', ...style }}
+      style={{ width: '55ch', ...style }}
```
ollibowers, line 118: knitpick: mayb back to rem just for consistency and its more understandable 🥺

ollibowers (approved):

frontend/src/components/CourseSearchBar/CourseSearchBar.tsx
ollibowers: (non urgent)

imagine-hussain: Can do in future fix. Added to (mental) backlog

PR #1141, opened 2024-05-26
removing the bug feedback button on the degree wizard to make it look…

### LINKED ISSUE #1126: Make sure the bug button never covers the start browsing button
- could be that you change it's position
- maybe you decide to hide it when the "start browsing" button appears
- could look into IntersectionObserver

jacksonwanggg:
… cleaner and so it doesnt overlap with the start browsing button


lhvy:
I think this may need to be rebased/cherry-picked into `331-staging`...

ollibowers (approved):
looks good for now! 
note: in future would be nice to bring this back with some better handling

PR #1142, opened 2024-05-26
Add stronger warnings for WAM and exclusion validation

lhvy:
- Update exclusion condition to check courses in the same/current term
- Update WAM condition to give a warning instead of information message in the event that WAM goal is not met AND all courses in previous terms have associated marks


ollibowers (approved):
looks good to me :)

PR #1144, opened 2024-06-07
fix(course selector): fixed ui of disproportional loading status on first tab load

### LINKED ISSUE #1119: First tab has very weird sized loading symbol on first page load of Course Selector

pikapiku:
- Modified the spinner design for course tab load to fit into the tab component.


frontend/src/components/Spinner/Spinner.tsx
```diff
@@ -5,13 +5,25 @@ import S from './styles';
...
+  component?: string;
 };
 
-const Spinner = ({ text }: Props) => (
-  <S.Wrapper className="loading-spinner">
-    <Spin indicator={<LoadingOutlined style={{ fontSize: '2.5rem' }} spin />} />
-    <p>{text}</p>
-  </S.Wrapper>
-);
+const Spinner = ({ text, component = '' }: Props) => {
```
ollibowers: good idea, but maybe for reusability, we could instead make the `component` prop be `size`, that has a possible string type of `'small' | 'large'` (defaulting to large).

frontend/src/components/Spinner/Spinner.tsx
```diff
@@ -5,13 +5,25 @@ import S from './styles';
...
+  // modify spinner css for course tab loads
+  if (component === 'course-tab') {
+    return (
+      <S.Wrapper className="loading-spinner" style={{ gap: '1px' }}>
```
ollibowers: and then you can make two wrapper styles for small or large, and have no on-element styling.
(orrrrr rewrite styled wrapper to be an antd Flex as we woed about last meeting :sob:)

ollibowers (changes requested):

frontend/src/components/Spinner/Spinner.tsx
ollibowers: alsoooooo, make small a flex row

frontend/src/components/Spinner/Spinner.tsx
```diff
@@ -5,13 +5,30 @@ import S from './styles';
 
 type Props = {
   text: string;
+  size?: string;
```
ollibowers:
```suggestion
  size?: 'small' | 'large';
```
If we do something like this, then the intellisense will be able to tell you the options and we will get more typechecking capability.

frontend/src/components/Spinner/Spinner.tsx
```diff
@@ -5,13 +5,30 @@ import S from './styles';
...
+  if (size === 'large') {
+    return (
+      <S.LargeWrapper className="loading-spinner">
+        <Spin indicator={<LoadingOutlined spin />} size="large" />
+        <p>{text}</p>
+      </S.LargeWrapper>
+    );
+  }
+
+  return null;
```
ollibowers: I think this is fine (and should never be reachable if we pass typecheck), but might lead to bugs that become hard to find. It might be better to just make the large condition the else condition, so we never have to deal with a backup return.

ollibowers (changes requested):
some small last changes requested, otherwise this looks really good!

ollibowers (approved):
looks good to me! 
I can imagine there are a few more spots where we would need this, but we can address those as we notice them (which is much easier to do now)!

PR #1145, opened 2024-06-07
fix(frontend): fixed term planner to show empty grid box when no course selected. 

pikapiku:
- instead of a loading status on empty course select on term planner, it now prints an empty grid

### DIFF
--- a/frontend/src/pages/TermPlanner/TermPlanner.tsx
+++ b/frontend/src/pages/TermPlanner/TermPlanner.tsx
@@ -272,7 +272,7 @@ const TermPlanner = () => {
           // TODO: Fix Suspense by updating to react-query v5
           /* <Suspense fallback={<Spinner text="Loading Table..." />}> */
         }
-        {courseQueries.some((c) => c.isPending || coursesQuery.isPending || plannerQuery.isPending )  ? (
+        {courseQueries.some((c) => c.isPending) ? (
           <Spinner text="Loading Table..." />
         ) : (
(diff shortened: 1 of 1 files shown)


frontend/src/pages/TermPlanner/TermPlanner.tsx
```diff
@@ -272,7 +272,7 @@ const TermPlanner = () => {
           // TODO: Fix Suspense by updating to react-query v5
           /* <Suspense fallback={<Spinner text="Loading Table..." />}> */
         }
-        {courseQueries.some((c) => c.isPending) || !courseInfos[LIVE_YEAR] ? (
+        {courseQueries.some((c) => c.isPending || coursesQuery.isPending || plannerQuery.isPending )  ? (
```
lhvy: What motivated this change? Also, I don't think the added content should be inside the call to `some`.

pikapiku: good spotting. i'll fix this and send a pr again

ollibowers (changes requested):
looks good to me so far, if you could, please rebase off 331-staging with lucas changes, and then test out a course that has no legacy offering, (such as COMP6131 with a start year of 2021)

ollibowers:
> looks good to me so far, if you could, please rebase off 331-staging with lucas changes, and then test out a course that has no legacy offering, (such as COMP6131 with a start year of 2021)

331-staging doesnt work with this even, so blocking PR until it gets fixed in 331-staging

ollibowers:
this has been merged inn #1160

PR #1146, opened 2024-06-07
fix(frontend): Fix term planner for newly created courses

lhvy:
i.e. courses created after the user's degree started


martanman (approved):
seems fixed

PR #1148, opened 2024-06-10
fix(ci): Bump Python to 3.12 in CI

lhvy:
Bump CI to Python 3.12. Also remove the patch version from CI/Dockerfile. This will keep us at the latest patch version (no breaking changes), and speed up CI significantly as CI will now use any (usually latest) release of 3.12 pre-installed on the test runner.


ollibowers (approved):
beautiful i love you

PR #1150, opened 2024-06-14
updated scraper to use new api

martanman:
Scrapes postgrad items as well but currently still only saves undergrad courses.


backend/data/scrapers/programs_formatting.py
```diff
@@ -33,14 +33,22 @@ def format_prg_data() -> None:
     # Initialise formatted data
     programs_formatted: dict[str, dict] = {}
     for program in raw_content:
+        # if program['studyLevelValue'] != 'ugrd':
+        #     continue
```
ollibowers: comment?

backend/data/scrapers/payload.py
```diff
@@ -26,70 +86,26 @@ def create_payload(size, content_type, year = LIVE_YEAR):
...
-                                ],
-                            }
-                        },
-                        {
-                            "bool": {
-                                "minimum_should_match": "100%",
-                                "should": [
-                                    {
-                                        "query_string": {
-                                            "fields": [f"{content_type}.active"],
```
ollibowers, line 66: do we want to add back the active query field?

backend/data/scrapers/payload.py
```diff
@@ -26,70 +86,26 @@ def create_payload(size, content_type, year = LIVE_YEAR):
...
+            #     "queryField": "parentAcademicOrg",
+            #     "queryValue": "57a56ceb4f0093004aa6eb4f0310c7ae"
+            # },
+            # {
+            #     "queryField": "studyLevel",
+            #     "queryValue": "undergraduate"
+            # },{
+            #     "queryField": "studyLevelValue",
+            #     "queryValue": "ugrd"
+            # }
```
ollibowers: mayb we should uncomment these (instead of filtering) until we stop filtering, for query time. 
NOTE: not sure if this will work for specs and programs though...

backend/data/scrapers/payload.py
```diff
@@ -7,10 +7,70 @@
...
+            "siteId": "unsw-prod-pres",
+            "contentType": content_type,
+            "queryParams": [{
+                "queryField": "implementationYear",
+                "queryValue": str(year)
+            },
+            # {
+            #     "queryField": "parentAcademicOrg",
+            #     "queryValue": "57a56ceb4f0093004aa6eb4f0310c7ae"
+            # }
```
ollibowers: commnet

backend/data/scrapers/payload.py
```diff
@@ -7,10 +7,70 @@
...
+    while offset < max_items:
+        data_payload = _create_payload(offset, ITEMS_LIMIT, content_type)
+        r = requests.post(
+            URL,
+            data=json.dumps(data_payload),
+            headers=HEADERS,
+            timeout=60 * 5
+        )
+        # r.json() looks like { "data": { "data": [{}, ..., {}], "count": 123 } }
+        print(json.dumps(r.json())[:100], flush=True)
```
ollibowers: debug print (mayb good to see that its actually working and not just silently failing)

backend/data/processors/specialisations_processing.py
```diff
@@ -145,6 +147,8 @@ def get_levels(title: str) -> list[int]:
     Parses 'title' to get curriculum levels of specialisation item.
     Level can be any combination of {1, 2, 3, 4, 5, 6, 7, 8, 9}.
     """
+    if title == "any level 3 Mathematics course":
+        print("!!!!!!!")
```
ollibowers: debug print

ollibowers (changes requested):

backend/data/scrapers/programs_formatting.py
martanman: k i ll get rid of it cos i have the undergrad check when scraping

backend/data/scrapers/payload.py
martanman: i will investigate what adding it back will do :monocle_face:

backend/data/scrapers/payload.py
martanman: query time is like <1min in worst case (scraping courses) so not rly a problem

backend/data/scrapers/payload.py
martanman: ok

backend/data/scrapers/payload.py
martanman: I put it in in the right format, and it gave 0 results, so I'm just gonna leave it out and assume it doesn't work anymore

backend/data/scrapers/payload.py
martanman: i wanna keep (ill add a string at the start tho) cos it helps to see the thing's working

backend/data/processors/specialisations_processing.py
martanman: oke

ollibowers (changes requested):
would be banger to only keep undergrad programs and specialisations as well (since not super sure if this breaks things)

also run condition reparsing for pkl files

ollibowers (approved):
banger bange rbange rbagna

PR #1151, opened 2024-06-14
feat(be): Add support for new degrees

lhvy:
- Add support for Advanced Computer Science
- Prepare to add support for City Planning (Honours)

Note that City Planning is still hidden because the Degree Wizard is unable to cope with degrees that do not have any specialisations. I have made it clearer where this problem lies, but we do not have time to fix it currently.


ollibowers:
👿

ollibowers:
(i will rework degree wizard a bit before this goes in)

martanman (approved):
lgtm

PR #1152, opened 2024-06-16
Clicking progression checker now opens new tab

chanfaileui:
- added back redux state handling in Progression Checker


ollibowers (approved):
looks amazing!

PR #1155, opened 2024-06-21
fix(fe): Fix 422 errors when changing degree settings

lhvy:
- Also add a pop-up confirmation regarding updating degree length if courses will be unplanned as a result


frontend/src/pages/TermPlanner/SettingsMenu/SettingsMenu.tsx
```diff
@@ -23,35 +21,83 @@ const SettingsMenu = ({ planner }: Props) => {
...
-    try {
-      await axios.put('/user/updateDegreeLength', { numYears: value }, { params: { token } });
-    } catch {
+  async function updateStartYear(year: string) {
+    await axios.put(
+      '/user/updateStartYear',
+      { startYear: parseInt(year, 10) },
+      { params: { token } }
+    );
+  }
```
ollibowers, line 42: (in future all api requests will need to be put in helpers inside the api folder, but this is ok right now since i know its done a lot of places)

ollibowers (approved):
looks good to me

PR #1156, opened 2024-06-21
feature(global): auth?!?! USER AUTHENTICATION?!?!?!?!

ollibowers:
todo after merge (martin didnt want me to create more diff):
- **add a logout button on degree-wizard**
- react query key refactoring
- docstrings (will do this after approvals as itll cause annoying changes)
- tests? tests? tests? lol no later
- couple our session storage more closely with the csesoc auth
- remove the last few user axios requests from certain fronteng pages like SettingsMenu
- do a lil user route refactoring so they actually all directly use the database helpers i have created
- other stuff i have noted in minutes
- anything that is `TODO-OLLI(pm)` (pm stands for post auth merge)
- setup docker containers for production
  - this involves also dropping the redis stack to not include gui
- confluence docs baby
- figure out token batch deleting
- can probably do some preliminary error checking on the success login page for expected query params

# Changes (will format nicer in abit)

## backend changes
- removed nodemon from backend docker, in favour of uvicorn watch
- removed init-mongo from the main stack, in favour of just doing this step before the uvicorn start
- added new flags for runserver.py
- refactored the database code to all be in one spot
  - see backend/server/db
  - moved connections into here as well
  - created helper functions ontop the user related queries
  - dont clear users anymore on startup
  - cleaned up mongo setup and handling in general
- refactored the user database
  - now you have invariants
  - also has full mongo validation for all three user related collections
  - users now are either setup or not, no default users anymore
- added a bunch of request and validation helpers for the oidc auth code flow
- added redis and built out a session storage layer ontop of oidc
  - look in auth_utility/sessions
  - includes a set of interface functions that abstract the db helpers
  - supports guest sessions and fedauth (oidc) sessions
  - redis maps session token -> sid & uid
    - and is not persisted
  - mongo collections map;
    - refresh token -> sid
    - sid -> session info
    - uid -> user info
- created some middlewares for getting uid into routes
- converted all user routes to use the uid middleware
- underwrote the user routes to use adapters for now
  - for converting the db models into the TypedDicts
  - will do further conversion and remove the TypedDicts after merge
- summer term is false by default again
- added lifespan manager for future use (could probs remove)
- now requires a sessionsdb env file too
- changed workspace vscode settings so not auto run pytests
- changed run_app.py to use new docker containers
  - also made its environment more explicit. Before `load_dotenv` was just modifying the global environment, meaning the `get_x_env()` calls were a bit redundant
- broke tests (half way through fixing)
  - they atleast run now locally

## frontend changes
- now have token boundaries which handle redirecting based on user state
- added a identity provider which handles refreshing of tokens on page load
  - also auto refreshes just based on interval, other solutions have been explored and will continue to be
- added logout page
- reworked login page
- moved identity and token into its own redux slice
  - SHOULD NOT BE REDUX PERSISTED
- created useToken hook ontop of useSelector to ensure a defined token
  - also has a mode that allows undefined, still better than selector
- converted all request helpers to take in token as param
- removed any google auth related stuff
- converted all queries to use token
- changed where some links go (course-selector is now the first page for setup users)
- added degree wizard as a topbar page since it should be real page now
  - hamburger menu coming soon
- changed up degree wizard to work with new setup states
  - thus removed isComplete field from users
- landing page also now knows where to redirect you too based on isSetup
- error404 page dont show header anymore (since the cart required auth)
- created the required request helpers for auth
- created a withAuthorization request helper for formatting token header
- deleted some dead request helpers
- broke tests

roast me, burn me at the stake

(dm me if u need env files (you will need new env files), ill distribute soon)


backend/init-mongo.dockerfile
```diff
@@ -2,7 +2,7 @@
 # speeds up development, since we can rebuild and create new containers without
 # overwriting every time.
 
-FROM python:3.11.5
+FROM python:3.12.3-slim
```
Peedee2002: Nit: you can use alpine here to make it a bit less intensive. Sometimes that fucks up arm users, so make sure you test it on an arm machine too.

backend/server/db/helpers/models.py
```diff
@@ -0,0 +1,115 @@
...
+    courses: UserCoursesStorage
+    planner: UserPlannerStorage
+
+class NotSetupUserStorage(_BaseUserStorage):
+    setup: Literal[False] = False
+
+# https://github.com/pydantic/pydantic/issues/1223#issuecomment-1152323275
+class PartialUserStorage(BaseModel):
+    # for batch user storage update
+    # TODO-OLLI(pm): remove this and use keyword args since they cannot be None
```
Peedee2002, line 59: There is a way for uvicorn to handle this 👀
We were forced to do this because BaseModel didn't allow "NotRequired"

Peedee2002 (commented):
Very cool, will review on my PC later

imagine-hussain:
Holy hell

backend/runserver.py
```diff
@@ -4,25 +4,45 @@
...
     try:
         args = parser.parse_args()
     except argparse.ArgumentError:
         parser.print_help()
         sys.exit(0)
 
-    if args.overwrite:
+    if args.overwrite or args.dev:
+        # TODO-OLLI(pm): abstract this, and remove these local imports once we have proper connection handling
+        # TODO-OLLI(pm): also do overwrite checks for dev mode using a timestamp
```
imagine-hussain, line 39: What is the `overwrite check` consisting of?

backend/runserver.py
```diff
@@ -4,25 +4,45 @@
...
+        print("-- Finished Redis Setup")
 
-    uvicorn.run(app, host='0.0.0.0')
+    print(f"-- Starting uvicorn(reload={args.dev})")
+    uvicorn.run(
+        "server.server:app",
+        host='0.0.0.0',
+        lifespan="on",
+        reload=args.dev,
+        reload_excludes="*.json" if args.dev else None,
```
imagine-hussain, line 56: > Set glob patterns to exclude while watching
                                  for files. Includes '.*, .py[cod], .sw.*,
                                  ~*' by default; these defaults can be
                                  overridden with `--reload-include`. This
                                  option has no effect unless watchfiles is
                                  installed.

We lose the the exclusion of dot and swap files in dev, which we probably don't want to do.

backend/server/db/helpers/models.py
imagine-hussain, line 59: Is this comment
1. The fields should no longer be none?
2. The arguments passed in should not be allowed to be `None` but, the argument can be omitted?

backend/server/db/helpers/session_tokens.py
```diff
@@ -0,0 +1,75 @@
...
+from uuid import UUID
+
+from server.db.redis.conn import sdb
+from redis.commands.search.query import Query
+from .models import SessionID, SessionToken, SessionTokenInfoModel
+
+# FT.CREATE idx:uid ON HASH PREFIX 1 "stoken:" NOOFFSETS NOHL NOFIELDS NOFREQS STOPWORDS 0 SCHEMA uid TAG CASESENSITIVE
+# FT.SEARCH idx:uid "@uid:{z5362383}" NOCONTENT VERBATIM
+# FT.SEARCH idx:sid "@sid:{bf362dd3\\-046a\\-49e4\\-8d63\\-fd379f06a40f}" NOCONTENT VERBATIM
+# HMGET stoken:Nj-C2n8JYS4imxlVn7WhSX5Pa0uRZ1awa1w_YA-vnN4jMXXqqKVb7HW9rNhfGako sid uid
```
imagine-hussain, line 13: Are these meant to still be here? What do they do?

backend/server/db/helpers/sessions.py
```diff
@@ -0,0 +1,100 @@
...
+        },
+        upsert=False,
+        hint="sidIndex",
+    )
+
+    return res.matched_count == 1
+
+def delete_all_sessions(sid: SessionID) -> bool:
+    res = sessionsNewCOL.delete_many({ "sid": sid })
+    return res.deleted_count > 0
```
imagine-hussain: Makes more sense for this function to return the count. Callers that were going to do check the return value can do their own checks for `> 0` - or even keep the assert / condition the same.

backend/server/db/helpers/users.py
```diff
@@ -0,0 +1,192 @@
...
+    res = get_user(uid)
+    return res.courses if res is not None and res.setup is True else None
+
+def get_user_planner(uid: str) -> Optional[UserPlannerStorage]:
+    res = get_user(uid)
+    return res.planner if res is not None and res.setup is True else None
+
+def user_is_setup(uid: str) -> bool:
+    res = usersNewCOL.find_one({ 'uid': uid, 'setup': { "$eq": True } })
+    return res is not None
```
imagine-hussain: I think optimal API is

```
val get_user : uid -> Optional[Union[NotSetupUserStorage, UserStorage]]
val get_user_validate : uid -> Optional[UserStorage]
```

Then, for devs that want access to the `degree`, `planner` fields, etc

```
user = get_user_validated(id)
if user is None: raise ...

# The following fields should now be non-null
use user.program
use user.degree
```

imagine-hussain (changes requested):
I've had a look at some op the backend configs. Mostly looks good. Some comments about API; will review in more detail soon.

backend/runserver.py
ollibowers: this is in relation to overwriting course data on every startup, waiting 30 seconds agitates me... so mayb store timestamps on scrape, and do a quick query to check if we actually need to overwrite the archives and course data. (would only be if `dev=True`, prod should overwrite regardless)

backend/runserver.py
ollibowers: oh super interesting

backend/init-mongo.dockerfile
ollibowers: oooo ya i looked into this a while back, and read it actually kinda sucks for anything with wheels https://www.reddit.com/r/docker/comments/g5hb93/alpine_vs_pythonslim_for_deploying_python_data/ (and doesnt that include levenstein?)


(init-mongo is dead btw so this would only really be for ci-backend or whatever)

backend/server/db/helpers/models.py
ollibowers: how do u mean?

ollibowers: Im still on fence with this.
The aim was to allow just passing in Degree or Courses or a combo of any... But with Pydantic, to indicate a not required field, you must also allow that field to be `None`. Because of this, In Pydantics `model_dump` with `exclude_unset`, something like `PartialUserStorage(courses=...)` would convert to a dict with only a `courses` key, BUT, something like `PartialUserStorage(courses=..., degree=None)` converts to a dict including the `degree` field (BAD). An alternative is using the `exclude_none` param in `model_dump`, BUT an adverse side effect is that this applies over the whole dump, where we want to keep `None` in some fields... So with this Pydantic way, I have to filter out the `None` fields at the top level after deserializing...

So this comment is more just saying "I don't actually want there to be `None` fields". 

With keyword args, you can do something like `def func(*, a: int, b: int): ...` which truly allows you to have optional fields or params that are not allowed to be given as `None`, thus removing any unclear behaviour.

ollibowers: but tbh, i think the post deserialization filter is just cleanest, and gives the type validation i want, whereas keyword args would not (specifically for the courses dict), without using the pydantic validation decorators (which are more added code complexity)

backend/server/db/helpers/session_tokens.py
ollibowers: oh this is just for self reference later, since its kinda hard to figure out these queries

ollibowers: could probs remove

backend/server/db/helpers/sessions.py
ollibowers: this is actually interesting, as there should only be one session per sid (as is enforced by a unique sid index)... I think I had the delete_many for sanity.

Should I convert to delete_one and change it to return `res.deleted_count == 1`?

backend/server/db/helpers/users.py
ollibowers: good point, thats how it ends up actually working, since everything goes through `get_setup_user` in the routes (which auto throws 403), and the courses/degree/planner is just indexed off of that.

I probably wrote these before i realised that

backend/runserver.py
ollibowers: what are these files for?

backend/server/db/helpers/users.py
ollibowers: i fixed this

Peedee2002:
@ollibowers your docker compose is failing. This is because you changed the way env files worked.
https://github.com/devsoc-unsw/circles/blob/6c49c0cf609e20e6c380ef96c05012daac56f7aa/.github/workflows/ci.yml#L78
Have a look here to make a script which does create fake creds that work.

ollibowers:
> @ollibowers your docker compose is failing. This is because you changed the way env files worked.
> 
> https://github.com/devsoc-unsw/circles/blob/6c49c0cf609e20e6c380ef96c05012daac56f7aa/.github/workflows/ci.yml#L78
> 
> Have a look here to make a script which does create fake creds that work.

sory yea i didn't really understand the ci, i will fix

backend/runserver.py
imagine-hussain: I see. I would rename to `overwrite_data_on_startup`

backend/runserver.py
imagine-hussain: The `~*` files are swapfiles that are sometimes created by editors as a backup of a buffer. Often for things like preserving some state to undo to after the editor exits or, for reverting after failed file writes.

backend/server/db/helpers/models.py
```diff
@@ -0,0 +1,115 @@
...
+
+#
+# Session Models (mongo)
+#
+# TODO-OLLI(pm): figure out if we want to actually annotate id token type so no casts needed
+class SessionOIDCInfoModel(BaseModel):
+    access_token: str           # most recent access token
+    raw_id_token: str           # most recent id token string
+    refresh_token: str          # most recent refresh token
+    validated_id_token: dict    # most recent valid id token object
```
imagine-hussain, line 88: Are we able to get any stronger typing on this, than just `dict`?

backend/server/db/helpers/models.py
```diff
@@ -0,0 +1,115 @@
...
+    uid: str                    # for validation and back lookup
+    type: Literal['notsetup']   # ensure that this can get parsed correctly
+    expires_at: int             # time of expiry, will be replaced with a TTL on the cache
+
+class SessionInfoModel(BaseModel):
+    # object stored against the sid
+    uid: str                      # for validation and back lookup
+    oidc_info: SessionOIDCInfoModel
+    curr_ref_token: RefreshToken  # the most recent refresh token, only one that should be accepted
+    type: Literal['csesoc']       # ensure that this can get parsed correctly
```
imagine-hussain: quite nice :)

backend/server/db/helpers/refresh_tokens.py
```diff
@@ -0,0 +1,41 @@
...
+            "sid": info.sid,
+            "expiresAt": datetime.datetime.fromtimestamp(info.expires_at, tz=datetime.timezone.utc),
+        })
+        return True
+    except pymongo.errors.DuplicateKeyError:
+        # token already existed
+        return False
+
+def delete_all_refresh_tokens(sid: SessionID) -> None:
+    refreshTokensNewCOL.delete_many({ "sid": sid })
```
imagine-hussain: Allow this to return the number of refresh tokens deleted.

backend/server/db/helpers/session_tokens.py
```diff
@@ -0,0 +1,75 @@
...
+        # the type bindings are wrong, the result has the shape
+        #  {'attributes': [], 'total_results': 1, 'format': 'STRING', 'results': [{'id': '...', 'values': [...]}], 'warning': []}
+        assert isinstance(all_matches, dict)
+        if all_matches['total_results'] > 0:
+            # delete them all
+            # NOTE: potentially this wont work after 1GB worth of keys...
+            sdb.delete(*(m["id"] for m in all_matches["results"]))
+
+        if all_matches['total_results'] <= DELETE_BATCH_SIZE:
+            return  # no more to delete
```
imagine-hussain, line 76: I don't think I understand the need for paging here.
Is there a way to delete everything in a single query? Is there anything that ends up breaking if we do that?

backend/server/db/helpers/models.py
imagine-hussain: I do like storing the type like this and letting that be verified but, I don't like that there is no connection between the typeof `SessionInfoModel::type` and `NotSetupSessionModel::type`

Is there a way to create a relation as follows:
```
type NotSetupMarker = Literal['notsetup']
type CSESOCMarker = Literal['csesoc']
...

class SessionInfoModel(...):
  type: Literal[SessionInfoType[CSESOCMarker]]
``

File: backend/server/db/helpers/refresh_tokens.py [hunk 12]
```diff
@@ -0,0 +1,41 @@
...
+        return None
+
+    exp = int(info["expiresAt"].timestamp())
+    if exp <= int(time()):
+        return None
+
+    return RefreshTokenInfoModel(
+        sid=SessionID(info["sid"]),
+        expires_at=exp,
+    )
```
imagine-hussain: How do expired tokens get cleaned up?

backend/server/routers/auth_utility/oidc/errors.py
```diff
@@ -0,0 +1,127 @@
...
+from typing import Any, Dict, Optional
+
+class OIDCError(Exception):
+    # base class
+    def __init__(self, error_description: Optional[str] = None) -> None:
+        self.error_description: Optional[str] = error_description
+
+    def __str__(self) -> str:
+        desc = self.error_description or "No description."
+        return f"{super().__str__()}\n  Description: {desc}"
```
imagine-hussain: Super nitpicky but, I woudl remove the single space after `\n` and replace it with a `\t` or, just remove the ` ` entirely

backend/server/routers/courses.py
```diff
@@ -25,6 +25,8 @@
     tags=["courses"],
 )
 
+require_uid = HTTPBearerToUserID()
+
 # TODO: would prefer to initialise ALL_COURSES here but that fails on CI for some reason
 ALL_COURSES: Optional[Dict[str, str]] = None
```
imagine-hussain, line 31: Since we are finally getting a proper startup with redis, we should look into getting this fixed

backend/server/db/helpers/sessions.py
imagine-hussain: If we are sure that there will be only one - which I think we are able to enforce using the database itself right?
In that case, `delete_one` should be fine / preferred

imagine-hussain (commented):
Reviewed much more of the backend - still some distance to go.

The wrappers around the db / redis code really make me wish that we were able to enforce something like C++'s `[[nodiscard]]` or rust's `#[must_use]`.

However, this doesn't look like something that `mypy` currently supports :c
We can just be extra-careful when it comes to review I suppose.

https://github.com/python/mypy/issues/17069

backend/runserver.py
ollibowers: oh wait if you mean the args (i thought you meant the comments), then yea `overwrite` was for always overwriting data and has been there for ages, `dev` is used for when your in dev, so has the hot reload and overwrite (or as mentioned above, hopefully smart overwrite)...

but yea, will rename

backend/server/db/helpers/sessions.py
ollibowers: agreed now

backend/server/db/helpers/models.py
ollibowers: I do actually have strong types for this in the oidc packages...

My reasoning at the time for doing this was firstly that the expected fields can change afaik, so I didn't want to put strict validation on it in the mongo validations, thus neither here since these models are meant to be representation of the database contents. I also felt like due to this, it could couple the session abstraction to tightly with our expected oidc config. (like ideally the database handling just knows its a dict, then its the oidc layer is responsible for knowing its shape)

BUT that being said, doing this raw dict typing leads to casts being dropped around the place, so I am not sure I agree with my reasoning anymore. 
AND there are minimum set of fields that should always exist, so maybe I could figure out some base level of typing atleast and let there be extra fields. 
ALSO maybe since this session shape is specifically for the csesoc fedauth, that coupling is actually not a bad thing. Worst case, the fields change, we see some validation errors, change our validation shape, and drop all the old sessions which is completely fine.

not too sure tho, thoughts?

backend/server/db/helpers/refresh_tokens.py
ollibowers: real af

backend/server/db/helpers/refresh_tokens.py
ollibowers: I have put a ttl index on the documents based on the expiresAt field, which will handle it for us which is crazy cool (see the setup functions)

backend/server/routers/auth_utility/oidc/errors.py
ollibowers: true

backend/server/routers/courses.py
ollibowers: real

backend/server/db/helpers/session_tokens.py
ollibowers: Not really unfortunately, by default its paged to 10 results hence why i give a paged amount, and the max you can do is 10k (without changing configs), so I figured I'd just deal with it. (see LIMIT of FT.SEARCH)

Can probs up the constant though.

backend/server/db/helpers/models.py
imagine-hussain: If we couple this to the CSE fedauth, then we can get validation failures in testing

backend/server/routers/auth.py
```diff
@@ -1,14 +1,276 @@
...
+from server.db.helpers.users import delete_user, insert_new_user
+
+from .auth_utility.sessions.errors import ExpiredRefreshTokenError, ExpiredSessionTokenError, OldRefreshTokenError
+from .auth_utility.sessions.interface import create_new_guest_token_pair, get_session_info_from_refresh_token, get_session_info_from_session_token, logout_session, setup_new_csesoc_session, create_new_csesoc_token_pair, setup_new_guest_session
+
+from .auth_utility.middleware import HTTPBearer401, set_secure_cookie
+from .auth_utility.oidc.requests import DecodedIDToken, exchange_and_validate, generate_oidc_auth_url, get_userinfo_and_validate, refresh_and_validate, revoke_token, validate_authorization_response
+from .auth_utility.oidc.errors import OIDCInvalidGrant, OIDCInvalidToken, OIDCTokenError, OIDCValidationError
+
+STATE_TTL = 10 * 60
```
imagine-hussain: We should be more specific about what this it a TTL for

backend/server/routers/auth.py
```diff
@@ -1,14 +1,276 @@
...
+            # invalid grant could happen if the oidc token is expired, thats fine
+            # NOTE: i dont think we want to throw error if it is invalid? but its fine i guess
+            raise HTTPException(
+                status_code=HTTP_401_UNAUTHORIZED,
+                detail="Your csesoc session has expired, please re-login.",
+                headers={ "WWW-Authenticate": "Bearer", "set-cookie": res.headers["set-cookie"] },
+            ) from e
+        except OIDCTokenError as e:
+            # cant imagine what error this could be, but if it happens, its bad
+            print(e)
```
imagine-hussain, line 271: Do we currently have a way to track any internal server errors that take place in prod?

backend/server/routers/auth_utility/ideas.md
imagine-hussain, line 1: Just double-checking, this is committed intentionaly?

backend/server/db/helpers/users.py
```diff
@@ -0,0 +1,160 @@
...
+    if len(payload) == 0:
+        # most semantically correct
+        return user_is_setup(uid)
+
+    if "courses" in payload and "degree" in payload and "planner" in payload:
+        # enough to declare user as setup
+        payload["setup"] = True
+
+    res = usersNewCOL.update_one(
+        { "uid": uid, "setup": True } if "setup" not in payload else { "uid": uid },
```
imagine-hussain: Is that ternary the right way around?
In the first if-statement, having the key means that the user is fully set up.
However, in the update, we insert the `setup: True` if we don't have the setup key in the payload

imagine-hussain (changes requested):
Finished review of the backend and added comments.

.vscode/settings.json
```diff
@@ -4,7 +4,6 @@
     ],
     "VitestRunner.executionArg": "npx --prefix frontend",
     "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true,
```
Peedee2002, line 7: 🤔 i thought this worked correctly and made sure you can run your pytest using vscode

backend/dev.dockerfile
```diff
@@ -25,5 +15,11 @@ COPY . .
...
 EXPOSE 8000
 
+# At time of writing this, uvicorn reload wouldn't pass through stdout
+# https://testdriven.io/blog/fastapi-docker-traefik/
+# ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
 # Run the server
-ENTRYPOINT nodemon --exec python3 -u runserver.py
+# https://forums.docker.com/t/docker-run-cannot-be-killed-with-ctrl-c/13108/2
```
Peedee2002, line 24: HOLY SHIT that is why? I always thought it was random if it cancelled or not. Wild.

.vscode/settings.json
ollibowers: it wassss, but it was also causing variables at top level to get evaluated in the test files, (and there are two variables which required route calls, so it was making random route calls everytime i saved which was bugging me)

backend/server/db/helpers/users.py
ollibowers: It's kinda confusing, but this is the filter arg for the update (what is used to find the documents we want to update).

You are right, we add the "setup" key into the payload above if they have given us all the other keys required to make a user setup. 
SO, if they haven't given us all the properties we need (and the key is not present), we are only patching certain fields, thus we require our search to look for a document that already setup (or we will be missing fields). 
OTHERWISE, if they have given us all the keys we need, then we can apply the update to the user regardless if they are setup or not. 

So we are doing:
> If we do not have all keys, look for a match that is already setup, otherwise just look for a user

Maybe I can flip the ternary to make this a bit clearer?

backend/server/routers/auth_utility/ideas.md
ollibowers: true i will remove

backend/server/routers/auth.py
ollibowers: i dont think so, a proper logger instance and usage is definitely needed (and something I have on the list of post merge todos)

backend/server/routers/auth.py
ollibowers: really true, i will rename it to `AUTH_STATE_COOKIE_TTL` to match the other constants,

backend/server/db/helpers/models.py
ollibowers: good point, i will change

Peedee2002:
There should be some docs https://github.com/devsoc-unsw/circles/blob/dev/backend/README.md that require update

ollibowers:
> There should be some docs https://github.com/devsoc-unsw/circles/blob/dev/backend/README.md that require update

good point, i will get this done!

backend/requirements.txt
```diff
@@ -1,18 +1,55 @@
...
 pytest==7.4.4
+python-dateutil==2.9.0.post0
 python-dotenv==1.0.1
 python-Levenshtein==0.25.1
 python-multipart==0.0.9
+pytz==2024.1
+rapidfuzz==3.9.0
+redis==5.0.3
 requests==2.31.0
+six==1.16.0
```
Peedee2002, line 45: this is a valid decision if you are **really finnicky** about your transitive dependencies. It will just make a lot of noise because you will have to manage these dependencies yourself and will get dependabot on your ass.

backend/server/config.py
```diff
@@ -2,10 +2,10 @@
...
 URI = "mongodb://mongodb:27017/?readPreference=primary&appname=MongoDB%20Compass&directConnection=true&ssl=false"
 
-CLIENT_ID = "1017197944285-i4ov50aak72667j31tuieffd8o2vd5md.apps.googleusercontent.com"
-
 FINAL_DATA_PATH = "./data/final_data/"
 
 ARCHIVED_DATA_PATH = "./data/final_data/archive/processed/"
 
-DUMMY_TOKEN = "token"
+DUMMY_TOKEN = "token"  # TODO: get rid of this
```
Peedee2002, line 9: yea, this was a mistake 💀 . There is a similar issue in the frontend with "bad_planner" or something. will be good to refactor that at some point.

backend/server/db/helpers/session_tokens.py
Peedee2002: if these are helpful, probably make them real functions so you can delete them from the comments

backend/server/routers/user.py
```diff
@@ -1,59 +1,115 @@
...
+from server.routers.model import CourseMark, CourseStorage, DegreeWizardInfo, CourseStorageWithExtra, DegreeLocalStorage, LocalStorage, PlannerLocalStorage, Storage, SpecType
 from server.routers.programs import get_programs
 from server.routers.specialisations import get_specialisation_types, get_specialisations
 
+import server.db.helpers.users as udb
+from server.db.helpers.models import PartialUserStorage, UserStorage as NEWUserStorage, UserDegreeStorage as NEWUserDegreeStorage, UserPlannerStorage as NEWUserPlannerStorage, UserCoursesStorage as NEWUserCoursesStorage, UserCourseStorage as NEWUserCourseStorage
+
+
 from pydantic import BaseModel
+# TODO-OLLI(pm): i think we can get rid of this now since we do not use ObjectId anywhere
```
Peedee2002: yea, sounds legit

docker-compose.yaml
```diff
@@ -73,19 +93,8 @@ services:
...
-      dockerfile: init-mongo.dockerfile
-    container_name: init-mongo
-    env_file:
-      - ./env/backend.env
-    depends_on:
-      - mongodb
-    networks:
-      - circles-net
+
+  # TODO-OLLI(pm): idek if this works or how it should work post init-mongo death
```
Peedee2002: I doubt it

Peedee2002 (commented):
a few small comments, damn its been a min since i have seen circles code

docker-compose.yaml
ollibowers: Oh i actually was thinking how to replace this, since i think ci mainly uses it to get the backend containers to run and cleaned...

I think if we can assume that the containers start empty, then we can just shift the ci to start backend directly (as itll run in dev, which at the moment only drops users). 
But, if we cannot, then i can make a simple new dockerfile that is just the same as the production backend docker (overwrites courses), but uses a new clarg to drop users too...

thoughts?

backend/server/routers/user.py
ollibowers: oki 🫡

backend/server/config.py
ollibowers: yaaa dummy data like bad_planner is used a lot in frontend because we never made a standard way of dealing with loading and error query states. 
in my react query followon PR, i will likely unify it all.

for dummy_token, its only used now in testing, so when we fix tests, we can just use guest login route instead.

backend/requirements.txt
ollibowers: yeaaaa ://// 

I was thinking of investigating something like poetry, or some more advanced (but simple) package manager that has a lock file.

backend/dev.dockerfile
ollibowers: bahahahhahahhaha, it is reallllyyy whack, before finding that everywhere i ever read also just said array and string were equivalent

docker-compose.yaml
Peedee2002: I believe the purpose was something to do with handling the production instances of databases. Not 100% sure tho.

ollibowers: interesting

backend/server/db/helpers/session_tokens.py
ollibowers: they already are (the functions in the file are implementations of them), was moreso keeping them around coz i use them in debugging... I will remove after merge when we revisit the indexing (i want to get rid of redis search its not appropriate)

backend/server/routers/courses.py
ollibowers: (will do in future)

backend/runserver.py
ollibowers: (looks like it combines them, will revisit if we find an issue)

backend/requirements.txt
ollibowers: i like it

frontend/src/components/Auth/RequireToken.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { token } = useAppSelector(selectIdentity) ?? {};
+
+  const {
+    isPending,
+    data: isSetup,
+    error
+  } = useQuery({
+    queryKey: ['degree', 'isSetup'], // TODO-OLLI(pm): fix this key, including userId
+    queryFn: () => getUserIsSetup(token!),
+    enabled: token !== undefined,
```
zax-xyz, line 25: thoughts on making a generic `useQueryWithToken` hook or similar that grabs the token and injects it into the `enabled` etc to reduce boilerplate?

frontend/src/components/PlannerButton/PlannerButton.tsx
```diff
@@ -12,7 +13,11 @@ interface PlannerButtonProps {
 }
 
 const PlannerButton = ({ course, isAddedInPlanner }: PlannerButtonProps) => {
-  const handleMutation = isAddedInPlanner ? removeCourse : addToUnplanned;
+  const token = useToken();
+
+  const handleMutation = isAddedInPlanner
+    ? (code: string) => removeCourse(token, code)
+    : (code: string) => addToUnplanned(token, code);
```
zax-xyz, line 20: thoughts on
```suggestion
  const handleMutation = (code: string) => (isAddedInPlanner ? removeCourse : addToUnplanned)(token, code);
```

zax-xyz: or perhaps
```suggestion
  const handleMutation = (code: string) => isAddedInPlanner ? removeCourse(token, code) : addToUnplanned(token, code);
```

frontend/src/components/QuickAddCartButton/QuickAddCartButton.tsx
```diff
@@ -14,7 +15,11 @@ type Props = {
 };
 
 const QuickAddCartButton = ({ courseCode, runMutate, planned }: Props) => {
-  const handleMutation = planned ? removeCourse : addToUnplanned;
+  const token = useToken();
+
+  const handleMutation = planned
+    ? (code: string) => removeCourse(token, code)
+    : (code: string) => addToUnplanned(token, code);
```
zax-xyz, line 22: same here

frontend/src/components/ResetModal/ResetModal.test.tsx
```diff
@@ -20,23 +19,25 @@ describe('ResetModal', () => {
...
+  //   useDispatchMock.mockReturnValue(dummyDispatch);
+
+  //   await renderWithProviders(<ResetModal open />);
+  //   // TODO: idk what to do with token here
+  //   await setIsComplete('', true);
+  //   await userEvent.click(screen.getByText('Reset'));
+  //   expect(dummyDispatch.mock.calls).toEqual([
+  //     [{ payload: undefined, type: 'courseTabs/resetTabs' }]
+  //   ]);
+  // });
```
zax-xyz, line 40: 💀

frontend/src/config/store.ts
```diff
@@ -2,12 +2,14 @@ import { configureStore } from '@reduxjs/toolkit';
...
 import courseTabsReducer from 'reducers/courseTabsSlice';
+import identityReducer from 'reducers/identitySlice';
 import settingsReducer from 'reducers/settingsSlice';
 import persistMigrate, { persistVersion } from './migrations';
 
 export const rootReducer = combineReducers({
   courseTabs: courseTabsReducer,
-  settings: settingsReducer
+  settings: settingsReducer,
+  identity: identityReducer
```
zax-xyz, line 12: surely add trailing commas to prettier config

frontend/src/hooks/useToken.ts
```diff
@@ -0,0 +1,52 @@
...
+function useToken(): string;
+function useToken(options?: Options): string | undefined;
+// need this general one to allow for dynamically-created params to still match a type
+// since something like { allowUnset: 100 === 200 } will be type boolean, and thus no return value can be inferred
+
+function useToken(options?: Options): string | undefined {
+  const token = useAppSelector(selectToken);
+
+  if (token === undefined && options?.allowUnset !== true) {
+    // This shouldn't occur of useToken is used inside a RequireToken
```
zax-xyz: :pwned:
```suggestion
    // This shouldn't occur if useToken is used inside a RequireToken
```

frontend/src/hooks/useToken.ts
```diff
@@ -0,0 +1,52 @@
...
+    throw TypeError('useToken: allowUnset was false when token was undefined.', {
+      cause: { token, allowUnset: !!options?.allowUnset }
+    });
```
zax-xyz: thoughts on just putting the cause as part of the error message string?

frontend/src/pages/CourseSelector/CourseMenu/CourseMenu.tsx
```diff
@@ -76,10 +79,8 @@ const CourseMenu = ({ planner, courses, degree }: CourseMenuProps) => {
 
   const queryClient = useQueryClient();
   const courseMutation = useMutation({
-    mutationFn: async (courseId: string) => {
-      const handleMutation = inPlanner(courseId) ? removeCourse : addToUnplanned;
-      await handleMutation(courseId);
-    },
+    mutationFn: async (courseId: string) =>
+      inPlanner(courseId) ? removeCourse(token, courseId) : addToUnplanned(token, courseId),
```
zax-xyz, line 83: does this still need to be async?

frontend/src/pages/DegreeWizard/DegreeWizard.tsx
```diff
@@ -31,12 +33,30 @@ const DegreeWizard = () => {
...
+    onSuccess: () => {
+      queryClient.invalidateQueries({
+        queryKey: ['degree']
+      });
+      queryClient.invalidateQueries({
+        queryKey: ['courses']
+      });
+      queryClient.invalidateQueries({
+        queryKey: ['planner']
+      });
```
zax-xyz, line 55: thoughts on 
```suggestion
      ['degree', 'courses', 'planner'].forEach(key =>
        queryClient.invalidateQueries({ queryKey: [key] });
      );
```

frontend/src/pages/LoginSuccess/LoginSuccess.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const navigate = useNavigate();
+  const [query] = useSearchParams();
+
+  useEffect(() => {
+    const exchangeCode = async () => {
+      try {
+        const identity = await CSELogin(Object.fromEntries(query.entries()));
+
+        dispatch(updateIdentityWithAPIRes(identity));
+        // TODO-OLLI(pm): could use a conditional useQuery too
```
zax-xyz: so true

frontend/src/pages/TermPlanner/ImportPlannerMenu/ImportPlannerMenu.tsx
```diff
@@ -1,31 +1,33 @@
...
   const planner = plannerQuery.data || badPlanner;
   const inputRef = useRef<HTMLInputElement>(null);
   const [loading, setLoading] = useState(false);
-  const { token } = useSelector((state: RootState) => state.settings);
 
   const handleSetUnplannedCourseToTerm = async (data: UnPlannedToTerm) => {
     try {
-      await axios.post('planner/unPlannedToTerm', data, { params: { token } });
+      await axios.post('planner/unPlannedToTerm', data, {
+        headers: { ...withAuthorization(token) }
```
zax-xyz: thoughts on 
```suggestion
        headers: withAuthorization(token)
```

frontend/src/reducers/identitySlice.ts
```diff
@@ -0,0 +1,45 @@
...
+        userId: action.payload.userId
+      };
+    },
+    updateIdentityWithAPIRes: (state, action: PayloadAction<IdentityResponse>) => {
+      return {
+        expiresAt: action.payload.exp,
+        token: action.payload.session_token,
+        userId: action.payload.uid
+      };
+    },
```
zax-xyz: thoughts on 
```suggestion
    updateIdentity: (state, action: PayloadAction<CompleteIdentity>) => ({
      expiresAt: action.payload.expiresAt,
      token: action.payload.token,
      userId: action.payload.userId
    )},
    updateIdentityWithAPIRes: (state, action: PayloadAction<IdentityResponse>) => ({
      expiresAt: action.payload.exp,
      token: action.payload.session_token,
      userId: action.payload.uid
    )},
```

zax-xyz (changes requested):
lgtm

frontend/src/components/Auth/RequireToken.tsx
ollibowers: this is definitely something we will add after merge!

frontend/src/components/ResetModal/ResetModal.test.tsx
ollibowers: 💀

frontend/src/components/PlannerButton/PlannerButton.tsx
ollibowers: Really good point. 
After looking more, I have noticed this is inconsistent in many places, so i might address in the query followup PR.
(These funky mutations were not added in this branch)

frontend/src/config/store.ts
ollibowers: real

frontend/src/components/QuickAddCartButton/QuickAddCartButton.tsx
ollibowers: ditto

frontend/src/pages/CourseSelector/CourseMenu/CourseMenu.tsx
ollibowers: ditto

frontend/src/hooks/useToken.ts
ollibowers: yes, i will do this!!!!!!

.github/workflows/ci.yml
lhjt, line 1: Very crazy have you been able to test if this is functional?

frontend/src/pages/DegreeWizard/DegreeWizard.tsx
ollibowers: oooo this is also good idea! will do in follow up!

frontend/src/pages/LoginSuccess/LoginSuccess.tsx
ollibowers: i actually disagree now 😡

frontend/src/reducers/identitySlice.ts
ollibowers: love it

frontend/src/pages/TermPlanner/ImportPlannerMenu/ImportPlannerMenu.tsx
ollibowers: good pick up, have fixed!

.github/workflows/ci.yml
ollibowers: not yet, have not fully fixed ci yet :c

ollibowers: @lhjt as far as I can tell, the docker container `ci-backend` is only used in our `ci.yml`... 
I hear that it might also be used elsewhere behind the scenes, is this true? Am I good to have editted it?

ollibowers: (i have also now fixed the ci to be in its final state for this PR, and the job now runs, so if you need to check anything 🥺)

martanman (approved):
lgtm

zax-xyz (approved):
trash

PR #1163, opened 2024-07-12
style(all): fixing pipelines

ollibowers:
- created 7 new jobs out of the previous 2.
  - seperate lint, typecheck and test jobs for backend and frontend
  - one docker build and start prod job that runs if all others pass
  - (currently test jobs are disabled as these fixes will come separately)
- fixed pylint errors
  - for some, i added explicit ignores as i felt they were appropriately broken
  - for some, i added explicit ignores with todos, as i do not want to refactor completely until we get tests
  - removed some redundant or poorly placed pylint disables
  - intentionally left docstring disables in there as we will fix these in a future PR
  - similarly, intentionally left circular import ignores as i will fix this next in a utility refactor + user dict removal PR
  - **reworked the opening to run_app.py**, so its less annoying
- fixed eslint disables
  - removed a bunch of bad eslint disables, only leaving the necessary ones in
  - disabled no-console and allowed non-default exports for the api directory
  - **refactored term planner** decently, as the eslint-disable removal required it. this was a much needed refactor and is one step further to getting TermPlanner.tsx readable. 
  - moved some axios requests into the api folder, so i could remove some other eslint-disables
    - all will be moved in a upcoming PR (currently assigned to a subbie)
  - **reworked graphical selector** slightly to rely more on the unlocked courses route, as previously it was trying to use a `.unlocked` property on the user courses response (which didn't exist). now we have colors again!!!
- verified that there are no mypy `type: ignore`s that we do not want (or that wont get culled by something else very soon)
- verified that there are no tsc `@ts-nocheck`s that we do not want
- update team.md


.github/workflows/ci.yml
```diff
@@ -14,78 +14,112 @@ jobs:
...
+        uses: actions/checkout@v4
+      - name: Install deps
+        uses: ./.github/actions/setup-ci-backend
+      - name: Setup environment
+        uses: ./.github/actions/setup-ci-env
       - name: Run the backend container
         run: docker compose up -d ci-backend
-      - name: Run mypy checks
-        run: cd backend && python -m mypy . && cd ..
+      - name: Wait for backend to initialise  # cringe as hell, remove when we get health checks to wait on
```
Peedee2002: Yea sorry oops

I didn't know about health checks when I wrote that cringe

ollibowers: nonononnooononononon, not calling previous writers cringe, you are not cringe 🥺. calling my sleep timeout a cringe workaround

sachk (approved):
not super thorough but lgtm, if you want more feedback u can ask jared since he's more experienced with gh actions.

martanman (approved):
lgtm

PR #1164, opened 2024-07-17
fix(be): Fix backend tests

lhvy:
- Fixes tests for the backend post auth changes

Huge thanks @Peedee2002 for finishing the remaining tests


backend/server/routers/courses.py
```diff
@@ -301,9 +301,13 @@ def get_all_unlocked(userData: UserData) -> Dict[str, Dict]:
...
 
     coursesState = {}
-    user = User(fix_user_data(userData.dict()))
+    user = User(fix_user_data(userData.model_dump()))
     for course, condition in CONDITIONS.items():
+        if course == "MATH1231":
+            print("MATH1231")
+            print("AAAAAAAAAAAAAAA")
         result, warnings = condition.validate(user) if condition is not None else (True, [])
+        print("ooop", result, warnings)
```
ollibowers:
```suggestion
        result, warnings = condition.validate(user) if condition is not None else (True, [])
```
was there something wrong with MATH1231 that should be further tested?

backend/server/tests/user/utility.py
```diff
@@ -14,3 +15,9 @@ def clear():
     """drop users in database. Used before every test is run."""
     setup_user_related_collections(drop=True)
     setup_redis_sessionsdb()
+
+def get_token():
+    return requests.post('http://127.0.0.1:8000/auth/guest_login').json()["session_token"]
+
+def get_token_headers():
+    return {"Authorization": f"Bearer {get_token()}"}
```
ollibowers: highest key, this should take the token and not generate a new one itself. (and thus everywhere should first get_token and then get_token_headers with the previous token value)

ollibowers: separation of responsibility

backend/server/tests/courses/test_search_course.py
```diff
@@ -4,30 +4,55 @@
...
 
 
 def test_search_course():
-    x = requests.post('http://127.0.0.1:8000/courses/searchCourse/COMP1', json=USER)
+    user_token = get_token()
+    requests.post('http://127.0.0.1:8000/user/setupDegreeWizard', headers={
+        "Authorization": f"Bearer {user_token}"
+    }, json=USER)
+    x = requests.post('http://127.0.0.1:8000/courses/searchCourse/COMP1', headers={
+        "Authorization": f"Bearer {user_token}"
```
ollibowers: after we change `get_token_headers` to take in a token for its arg, then this should not be manually built anywhere

backend/server/tests/courses/test_search_course.py
```diff
@@ -4,30 +4,55 @@
...
-    USER = json.load(f)["user6"]
+    USER = json.load(f)["user_degree_wizard"]
 f.close()
 
 
 def test_search_course():
-    x = requests.post('http://127.0.0.1:8000/courses/searchCourse/COMP1', json=USER)
+    user_token = get_token()
+    requests.post('http://127.0.0.1:8000/user/setupDegreeWizard', headers={
+        "Authorization": f"Bearer {user_token}"
```
ollibowers: ditto

backend/algorithms/objects/user.py
```diff
@@ -153,7 +153,7 @@ def wam(self, category: Category = AnyCategory()) -> Tuple[Optional[float], bool
                     counted_uoc += uoc
                     total_mark += uoc * grade
 
-        return (None if counted_uoc == 0 else total_mark / counted_uoc, total_uoc == counted_uoc)
+        return (None if counted_uoc == 0 else total_mark / counted_uoc, total_uoc == counted_uoc and counted_uoc != 0)
```
ollibowers, line 157: technically, if counted_uoc == 0, then all uoc do have an associated mark?

ollibowers (changes requested):

backend/server/routers/courses.py
Peedee2002: It was a confusing error in autoplanning, but had nothing to do with this course code. Sorry, should have removed this.

backend/algorithms/objects/user.py
Peedee2002: I did it like this because from a user perspective, we can't tell if a user just forgot to put in their marks or they actually did not have a mark. Same assumption as UNSW. they assume you have an error wam till the end of your first term

backend/server/routers/courses.py
ollibowers: oh all good

backend/algorithms/objects/user.py
ollibowers: Wait so (writing out so i fully understand), currently we use this boolean to tell whether we say a course is unlocked or not (ie the warnings are "soft" or "hard" warnings). 

So the idea behind it was that, if we are certain you do not meet the wam requirement (since you have entered all marks for every course you have), then we will say the course is locked (and the warnings will be hard warnings). Otherwise, if you have not entered all your marks, then we are not certain a WAM requirement is accurately met, thus we still let the course be unlocked. 

Currently, if
- no marks and no courses, then we are certain you do not make WAM requirement (since you have no WAM)
- 0 to some marks and some courses, we cannot be certain
- all marks for all courses, we can be certain on warnings

So by making this change, you are suggesting that if they have no courses and no marks, then we cannot be certain? (case 1)

ollibowers: i actually agree

Peedee2002: Yea

ollibowers (approved):
lgtm

PR #1182, opened 2024-08-17
bug(tp+pc): fix multiterm to function correctly when dragged and when reporting plannedFor

Peedee2002:
fixed this thingo
image
to make sure it can handle multiterm
and also fixed this

https://github.com/user-attachments/assets/fd1e597d-4ec2-4201-9a6f-cba296de7f34

to make sure we correctly handle when we move a multiterm course to unplanned


Peedee2002:
POG tiny review - might be controversial tho

PR #1191, opened 2024-08-28
adding the three buttons needed and reworked the texts

### LINKED ISSUE #1170: rework the error boundary to be more appropriate post 331
Historically when user data was stored on the frontend, it would often get into illegal states. This is why the error boundary featured two buttons; those being a reset user data, and a ignore and return. 

Nowadays, if user data gets into an illegal state, that is REALLY BAD, so we no longer want to just reset the user straight away. Instead, create three buttons:
- back to circles (instead send them back to landing page)
- logout (which logs them out and sends them back to
- reset (resets redux state and sends a reset user POST request)

Also, rework the error boundary to first tell them to try refresh the current page, and if it persists, send a screenshot to us. 
Would be also good to get some more helpful information on the screen (currently just displays the error as string):
- display the full user data (if it exists)
- display the full redux state (if possible)

Feel free to wait for salinas JSON ticket to be finished for the user data display part. 
- (can use the export buttons on the error boundary)

jacksonwanggg:
Waiting on the other PR so i can get the user information

### DIFF
--- a/frontend/src/components/ErrorBoundary/ErrorBoundary.tsx
+++ b/frontend/src/components/ErrorBoundary/ErrorBoundary.tsx
@@ -40,8 +40,12 @@ class ErrorBoundary extends React.Component<Props, State> {
       window.location.href = '/degree-wizard';
     };
 
-    const gotoCourses = () => {
-      window.location.href = '/course-selector';
+    const gotoLanding = () => {
+      window.location.href = '/';
+    };
+
+    const gotologout = () => {
+      window.location.href = '/logout';
     };
 
     const openModal = () => {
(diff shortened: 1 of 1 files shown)


frontend/src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -40,8 +40,12 @@ class ErrorBoundary extends React.Component<Props, State> {
       window.location.href = '/degree-wizard';
     };
 
-    const gotoCourses = () => {
-      window.location.href = '/course-selector';
+    const gotoLanding = () => {
+      window.location.href = '/';
+    };
+
+    const gotologout = () => {
```
lhvy, line 47: `gotoLogout`

PR #1194, opened 2024-09-01
feat(all): Add JSON import and export support

### LINKED ISSUE #1167: bring back json export buttons
On dev, there currently exists import and export as JSON buttons (also currently have export as PNG). 
In 331, we got rid of them, but kept the code (for example, ImprotPlannerMenu.tsx still exists). 

Bring back the json import and export functionality for 331. This can be just one button if you want, or two buttons like before. Feel free to use the existing code, HOWEVER, you will need to refactor it to utilise the user API routes (there is `/user/data` and `/user/saveLocalStorage` I believe). You may create new routes to achieve this too. 

You do not need to use react query to do this, you can just use the API helpers directly.

lhvy:
Resolves #1167


martanman (approved):
lgtm

lhvy:
Tested locally, seems fine

PR #1195, opened 2024-09-08
feat(api): convert last few routes to use token instead of old models

### LINKED ISSUE #1174: tokenify the last few payload routes
There are 2 (or rather 3 if u include ctf) routes that still use a userdata payload (`prepareUserPayload`). These routes are `/courses/getAllUnlocked/` and `/courses/coursesUnlockedWhenTaken/`. 

Fix these to use token, as it is annoying me. 

These routes require a `User` type object to work with (what is known as the "algo user"), as they do condition checking. However, you only get a `Storage` type object from the `get_setup_user(userid)` function. Thus, you will also need to create a function to convert `Storage -> User`. Give this a go yourself first, otherwise I do have some old code for this. Feel free to put this function as a static method on the `User` class, would go hard. 

Once done, also remove the prepareUserPayload function, and trim down any user data fetches that were being called just for this.

ollibowers:
tldr
- no more routes require user data in body
- no more 1000000 user representations, shrunk down to 2-3
  - `Storage`: the current in-use database representation (to be soon removed since we have the new database models)
  - `User`: the algorithm representation
- no more circular dependencies

major
- converted last routes away from user payloads
  - `/courses/getAllUnlocked`
  - `/courses/unselectCourse/{course}`
  - `/courses/coursesUnlockedWhenTaken/{course}`
  - `/ctf/validateCtf` (indirectly)
  - `/planner/autoplanning`
- converted frontend api requests and queries accordingly
- updated all tests accordingly
  - added new example users to support this (derived from 'exampleUser.json')
- created new functions to support this culling
  - **created `user_storage_to_algo_user(Storage) -> User` as the fundamental bridge between the main two user representations**
  - derived `get_all_course_states_for_user(User) -> CourseStates` from `/courses/getAllUnlocked` body
  - created `parse_mark_to_int(Mark) -> Optional[int]`
  - created `user_storage_to_raw_plan(Storage) -> RawUserPlan` to attach UOC and Marks to user planner (for easier term validation)
- removed some no-longer used models and bad functions
  - removed `fix_user_data` (previously used to add UOC to old planner dicts)
  - removed `convert_to_planner_data` and `ValidPlannerData` (previously used for `validate_terms`)
  - removed `UserData` (previously used by payload routes)
  - removed `ValidCourseState` and `ValidCoursesState`
  - removed `PlannerData` (previously used by autoplanning)
  - removed `prepareUserPayload` frontend helper
- convert `validate_terms` function to use new `RawUserPlan` type
  - this is a shape that is derived from combing user planner and course marks
- removed `markMap` from backend, and unified all letter grade conversions

minor
- add new type annotation for `User::load_json`
- removed env file reliance in 'oidc/constants.py'
- changed no-unused-vars and import order eslint rules to be warning, so compilation is still possible
- changed eslint script to treat warnings as a fail


martanman (approved):
lgtm

PR #1197, opened 2024-10-11
update most dependencies to latest

Peedee2002:
This is a precursor to trying to fix the tests - I would rather have a version bump.
Most of the code change is just package-lock.json


lhvy:
Most of this work was done in #1218, ESLint 9.x upgrade should be done properly in a later PR

PR #1199, opened 2024-10-20
feat: Add support for user data migration from local storage

### LINKED ISSUE #1172: setup migration for redux persist (localstorage -> backend)
Firstly, we need to migrate localstorage userdata to backend. Redux will immediately try destroy the old user data shape when 331 releases (i believe). So first step is creating a migrator that puts it in localStorage or similar, just so we can wait for the user to log in. 

Then, we need a way to get this data to the backend... This should probably happen when they first login to a zID, it prompts them with "you have stuff in localstorage, do you want to upload?". We could do it for guest login too, but thats probably dumb. Keep it if they say no, maybe give them 3 chances to say yes?

Secondly, we will only be hosting full auth on https://circles.devsoc.app/ (csesoc.app will become a redirect). This means you must create a way for the redux data to be taken, and sent with the redirect, to maybe a parking page `/upload?data=...` or something, and then on this devsoc page, it will put it into localStorage for the above step to deal with. 
Feel free to build this fully out, but it doesn't need to work within a circles deployment, since we will likely just put a standalone redirect deployment on the csesoc url. (directors can figure this out, just setup the redirect logic to be used). 

Feel free to discuss further...

lhvy:
Resolves #1172 

- Add Redux migrator that moves local storage user data from `persist:root` into `oldUser`
- Add functionality in Degree Wizard to import from, download, or clear `oldUser`
- Add support for the `migration` query parameter on the landing page, which given a valid `UserJson` type as a JSON string, will write it to `oldUser` on page load, allowing for `circles.csesoc.app` to redirect to `circles.devsoc.app` with the converted data in the URL, and then the existing migration flow via Degree Wizard to be followed

All tested locally but please please please test it more! Thanks. Also I bumped `antd` for new link style. Seems fine and changelog didn't indicate any breaking changes we needed to worry about.

### DIFF
--- a/frontend/src/utils/export.ts
+++ b/frontend/src/utils/export.ts
@@ -19,7 +19,7 @@ export type UserJson = {
 
 const settingsSchema = z.strictObject({
   showMarks: z.boolean(),
-  hiddenYears: z.array(z.number().int().positive())
+  hiddenYears: z.array(z.number().int().nonnegative())
 });
 
 const degreeSchema = z.strictObject({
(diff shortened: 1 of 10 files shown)


frontend/src/utils/export.ts
```diff
@@ -19,17 +19,17 @@ export type UserJson = {
...
 const degreeSchema = z.strictObject({
-  programCode: z.string(),
+  programCode: z.string().length(4),
   specs: z.array(z.string())
 });
 
 const plannerSchema = z.strictObject({
   unplanned: z.array(z.string()),
-  startYear: z.number(),
+  startYear: z.number().int().gte(2019),
```
imagine-hussain, line 32: banger

martanman (approved):
lgtm!!!!

PR #1205, opened 2024-11-03
frontend(queries): refactor to unified query interface to help maintainability

ollibowers:
Created an abstraction layer around useQuery and useMutation, to help standardise the settings and key usage:
- created query/mutation hook factory functions
- reworked ALL query keys (previously none matched)
- added a staleTime for both static and user queries, as well as gcTime for static queries
  - previously staleTime was defaulted to 0, meaning queries always got refetched
  - this meant mutations (and similarly keys) theoretically did nothing as the cache never got considered
  - IMPORTANT: this means some bugs may appear where data isnt being invalidated at the right time... but triv fix if so

Next up (when i have more time again):
- make it so that queryOptions doesnt have to be given in queries (change variadic type params)
- make it so we dont have to do funky defaults for dependent queries
  - currently if we try use `!`, we hit non-null assertion typeerrors
  - might make a queryOption `allowUnsetParams` which auto does the `enable` key and allows undefined params

**IMPORTANT**: Also snuck in a graph refactor to finally fix the bubble highlighting on mutation...


martanman (approved):
lgtm

PR #1212, opened 2025-04-17
fix(frontend): Fix links to sponsor sites

Keatsm:
Previously, the sponser logos would not properly link to their corresponding sites. This was due to a reliance on the names of logo files which were altered in production.

This change accounts for such.


matthew-lim-matthew-lim:
LGTM

matthew-lim-matthew-lim (approved):
Looks good W

PR #1213, opened 2025-04-18
fix(lp): changed the color of text from white to black on lightmode

jcongc:
- Previously could not see text on the landing page and header if on light mode
- Changed the text color to be white if dark mode, and otherwise it should be black

image
- Currently cannot see white text on lightmode


matthew-lim-matthew-lim (approved):
LGTM Nice work Jason. PS. the 2 checks for CI fail just because of an outdated CI setup on 331 that I fixed on dev branch. I can fix that

PR #1222, opened 2025-07-14
Migrate react-beautiful-dnd to @hello-pangea/dnd

lhvy:
See https://github.com/devsoc-unsw/notangles/pull/1027 for more detail. I was making the same change for Notangles, and realised this could be beneficial over on Circles.

This should push you in the right direction for enabling strict mode, and upgrading to React 19.

**Warning: I have not tested this PR for you.**


jcongc:
for some reason when i drag a course, after dropping it, it appears for a split second where it was dragged from

lhvy:
~~Hm. Lemme check if that was a pre-existing bug or not...~~

Yeah ok definitely some edge case with the new library. Unsure what is causing that unfortunately. If you check the console, there is a warning about an unsupported usecase with the drag and drop library, but that was there even with the old one. Maybe fixing it will make the bug go away if the new library has been updated further assuming we wouldn't use it the way we have? Otherwise I'm not sure... One not insane idea is to do optimistic updates in React Query (this was going to be a long term goal already), it is possible that the flash on the screen is the moment before the backend responds with the updated data from moving a course.

Given that I have a separate project to maintain, I'll probably leave this here for now in case someone else wants to take on fixing this small visual bug (or y'all are happy with it as is).

matthew-lim-matthew-lim:
Converting into a draft PR for now, will offer to subcom to pick up ticket, otherwise will add this back to the board as a ticket.

PR #1223, opened 2025-07-21
fix (tp): added defensive checking when accessing course in tp

jcongc:
after removing a course which was already in the planner or a course that was in the unplanned column, it would still be passed in the props, and would result in a crash


matthew-lim-matthew-lim (approved):
Great work Jason, keep it up!

PR #1224, opened 2025-07-23
Fix removing multiterm courses from planner

lhvy:


matthew-lim-matthew-lim (approved):
LGTM. Additionally, package-lock.json got changed, but Lucas said it should be like this.

PR #1227, opened 2025-07-24
Fixed notif text color in dark theme

jcongc:
probably not the best way to do this but should fix the issue


matthew-lim-matthew-lim (approved):
LGTM

PR #1240, opened 2025-09-22
Chore/mobile compatibility

ChronoComplexity:
Added mobile compatibility to the main page, improved sidebar design and functionality, repositioned rating system so it is visible.


ChronoComplexity:
Greatly improved mobile compatibility (Note that sidebar is now default closed on mobile, as otherwise it takes up an unreasonable amount of screen space, but can be opened at any time):
With a course open:
Improved course mobile compatibility demo
Default view:
Improved course default homepage mobile compatibility demo
Sidebar opened:
Sidebar opened on updated mobile view

Improved medium width compatibility: 
homepage medium width compatibility demo

Comparison with old mobile view which was replaced by the update:
Bad old mobile view that was replaced

ChronoComplexity:
Other noteable changes:
Changed sidebar to be able to be fully closed instead of having a min width and to default to being closed so users are encouraged to use the search bar selection (although the sidebar can be opened and closed at will on mobile), fixed sizing of elements to scale with a combination of rem and the screen viewport width of the user rather than being fixed size in px.

PR #1248, opened 2025-10-10
Feat/Difficulty popup for a term

le-eug:
Adds a popup that indicates "Easy", "Medium" or "Hard" in  each TermBox based on manageability ratings for the courses from unilectives. 

Feedback and suggestions for UI are welcome!

NOTE: For useCourseRatings, there is room to include other ratings from unilectives, eg enjoyability.


ChronoComplexity:
image

Is unacceptably inaccurate stating one of the most well known hardest CSE courses, a high workload course, and a known difficult math course together in a 3-course term qualify as easy.

PR #1250, opened 2025-11-09
Automatically adds core courses to the unplanned section after the completion of degree wizard

le-eug:
Note: I did honestly copy-paste some of the logic from the function add_to_unplanned so if you want me to refactor this part to avoid repetition just lmk


matthew-lim-matthew-lim:
LGTM, great work

PR #1251, opened 2025-11-29
(devx) Update docker container names so that they don't collide with generic docker container names

matthew-lim-matthew-lim:
Devx/Chore: A dev might have multiple things using the name "frontend" or "backend" (me), this requires the dev to remove the container each time they want to launch the other project, losing a lot of time.

Simple fix just renames the docker container names.


matthew-lim-matthew-lim:
Taku said on discord this will work fine

PR #1253, opened 2025-11-29
(Feat) Enable FE unit tests

matthew-lim-matthew-lim:
They were disabled after 331 merge, don't know why. This PR is threefold: 1. Active FE unittests, 2. See why they fail, 3. Fix them

Edit: Found out why; they don't pass


ChinoGoblino:
Sounds good to me. I think this area of the pipeline is more within your expertise

matthew-lim-matthew-lim:
> Sounds good to me. I think this area of the pipeline is more within your expertise

Cheers, assigned to @jeremyle56 for the time being

PR #1266, opened 2026-04-05
Remove timetable-scraper usage from Circles frontend

risu729:
## Summary
- remove the timetable-scraper API hook and related helper/types from the frontend
- remove the `Course Capacity` UI from the `/course-selector` course sidebar
- clean up the now-unused constants and cache typing that only existed for timetable data

## Why this is being removed
The frontend path was only being used to render course capacity in the `CourseDescriptionPanel` sidebar on `/course-selector`. The `GraphicalSelector` reused the same panel component, but that route did not enable the timetable query.

## Analysis: why the current deployment was not visibly breaking
I traced the usage in `circles`, checked `../timetable-scraper`, and checked `../deployment/projects/timetable-scraper/` plus the live deployment.

Frontend usage:
- `frontend/src/config/constants.ts` built `TIMETABLE_API_URL`
- `frontend/src/utils/api/timetableApi.ts` called that URL
- `frontend/src/utils/apiHooks/static.ts` exposed `useCourseTimetableQuery`
- `frontend/src/components/CourseDescriptionPanel/CourseDescriptionPanel.tsx` used that hook only when `pathname === '/course-selector'`
- `frontend/src/components/CourseDescriptionPanel/CourseAttributes/CourseAttributes.tsx` rendered the `Course Capacity` block from that data

Backend behavior:
- in `timetable-scraper/api/src/notangles/index.ts`, the API does `req.params.termId.substring(5)`
- that means requests like `2024-T3`, `2025-T3`, and `2026-T3` all collapse to the same in-memory bucket: `T3`
- the API stores a single shared dataset in `/data/data.json`; it does not keep separate year-partitioned files
- the deployed API is a `StatefulSet` with a persistent `/data` volume, so it keeps serving whatever dataset is currently loaded there
- the deployed scraper cronjob is explicitly pinned to `YEAR=2024`

Live checks:
- `https://timetable.devsoc.app/api/currentterm/` returned `T3`
- `https://timetable.devsoc.app/api/currentstartdate/` returned `09/09/2024`
- `https://timetable.devsoc.app/internal/dump` showed `lastUpdated = 1775379830907` (`2026-04-05 19:03:50 AEST`)
- `https://timetable.devsoc.app/api/terms/2026-T3/courses/COMP1511` returned `200`, but the payload contained `classes[0].termDates.start = 09/09/2024`
- `2024-T3`, `2025-T3`, and `2026-T3` for `COMP1511` returned identical payload hashes

So the app was not staying alive because the timetable API was healthy in a year-aware way. It was staying alive because the backend ignores the year portion of `termId`, and the frontend only used that data in one narrow UI surface.

## Verification
- `npm run types`
- `npm run lint`
- `npx vitest run src/pages/CourseSelector/CourseSelector.test.tsx` *(fails in this workspace because the test harness expects a backend on `127.0.0.1:8000` and errors with `ECONNREFUSED` before the assertions run)*


jcongc (approved):
lgtm taku 😙

PR #1309, opened 2026-07-23
chore(backend): add type casts to guest login rate limit test

sleepyjen4:
# Brief

Cleaned up type checker complaints in the guest login rate limit test as Redis client returned loose types.

# Changes

- Added the from typing import cast import.
- Wrapped limiter_db.keys("*") in cast(list, ...) so it knows we're dealing with a list when we assert and loop over the keys.
- Wrapped limiter_db.pttl(key) in cast(int, ...) so the > 0 check stops throwing a warning, as pttl comes back as a vague union type otherwise.

### DIFF
--- a/backend/server/tests/user/test_guest_login_rate_limit.py
+++ b/backend/server/tests/user/test_guest_login_rate_limit.py
@@ -1,5 +1,6 @@
 import os
 import time
+from typing import cast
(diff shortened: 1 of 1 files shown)


jcongc (approved):
lgtm

PR #1324, opened 2026-08-04
feat: tag popular electives per program and specialisation

KevinLin1803:
Marks the electives that students on your program and specialisation most commonly have in their plan.

### Backend

New endpoint `GET /courses/popularElectives/{programCode}[/{spec}]`:

```json
{
  "programCode": "3778",
  "sampleSize": 30,
  "popular": [{ "courseCode": "COMP2041", "count": 28, "percent": 0.93 }]
}
```

- Counts how many setup users have each course in their `courses` map, via a Mongo aggregation.
- Scoped to the caller's specialisation, so a COMPA1 student sees COMPA1 students' choices rather than the whole program.
- Core and general-education courses are excluded; only electives can be popular.
- Returns the top 5, and nothing at all below 20 students in the cohort (small samples are too noisy to mean anything).

### Frontend

- Course sidebar: a small flame next to popular electives, sized to match the row text. Hovering shows the exact student count.
- Course description: a rounded "Popular course" pill next to the title, with the same hover count.

### Testing

- 12 integration tests covering the aggregation helper and the endpoint: program scoping, spec scoping, core/gen-ed exclusion, the sample floor, exact counts and ordering, and the top-5 cap.
- Verified in the browser on a seeded 30-student cohort: exactly the expected 5 flames appear, the 6th-ranked elective is correctly cut, and non-elective and unseeded courses show nothing.
- Checked at laptop and narrow widths, in light and dark mode. The pill does not overflow down to a 320px container.

### Note for reviewers

There is no real popularity data locally, so all screenshots and manual checks used a synthetic cohort. The thresholds (top 5, minimum 20 students) are a judgement call and worth sanity-checking against real usage once this is on staging.


lhvy:
Where does this data come from? Do users have the ability to opt-out to their course choices being used for anonymous statistics?

KevinLin1803:
Data comes from MongoDB. 

Yeah good catch. Users don't have a way to opt out. Will add one at signup and somewhere else for existing users

